PR #19, opened 2025-12-18
chore: add PAIMON_THIRDPARTY_MIRROR_URL env

lucasfang:
### Purpose

set PAIMON_THIRDPARTY_MIRROR_URL env is a more user friendly way to set THIRDPARTY_MIRROR_URL in cmake


lxy-9602 (approved):
+1

Issue #5, opened 2025-12-15
[Feature] Add Global Index for Bitmap and ANN search


lxy-9602:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/paimon-cpp/issues) and found nothing similar.

### Motivation

In modern analytical and AI-driven workloads, efficient data retrieval—especially for high-cardinality filters or similarity-based queries—is critical. Traditional file-level or partition-level metadata (e.g., min/max statistics) often fall short when queries involve selective predicates on non-partition columns or require nearest-neighbor lookups in vector spaces. This leads to excessive I/O, scanning irrelevant files, and poor end-to-end latency.

To address this, we introduce global indexing in Paimon—a unified, table-wide index structure that spans all data files across partitions and snapshots. 

### Solution

Proposed Solution
We propose building global indexes in Paimon using Row Tracking (which assigns each record a stable, globally unique Row ID) and Data Evolution (ensuring consecutive row ID without gaps).

We will support two index types:

- Bitmap-based inverted indexes for fast scalar filtering (e.g.,type = X),
- Vector indexes via our in-house vector engine Lumina for efficient DiskANN search.

Index construction and lookup are distributed, enabling analytical engines like Spark and StarRocks to skip irrelevant data and fetch only matching records—dramatically improving query performance.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #23, opened 2025-12-19
feat(scan): support built-in global index search during scan process

lszskye:
### Purpose
feat(scan): support built-in global index lookup during scan

This PR enhances the scan operation to natively support global index lookups in two ways:

1. **Index-driven scan**: When a distributed query provides precomputed index results, the scan can directly utilize these results to fetch only the relevant data, significantly reducing I/O overhead.

2. **Built-in scan**: During the scan process, if global index search is enabled and applicable global indexes exist, the system automatically evaluates the scan predicates, uses them to query the relevant global index first, and then retrieves only the corresponding data. 

Linked issue: #5

### Tests
GlobalIndexTest

### DIFF
--- a/src/paimon/core/global_index/indexed_split_test.cpp
+++ b/src/paimon/core/global_index/indexed_split_test.cpp
@@ -169,5 +169,24 @@ TEST(IndexedSplitTest, TestValidate) {
         ASSERT_NOK_WITH_MSG(split.Validate(),
                             "Scores length does not match row ranges in indexed split.");
     }
+    {
+        std::vector<Range> row_ranges = {};
+        IndexedSplitImpl split(data_split, row_ranges);
+        ASSERT_NOK_WITH_MSG(split.Validate(),
+                            "Invalid IndexedSplit: row ranges mismatch data files.");
+    }
+    {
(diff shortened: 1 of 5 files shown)


src/paimon/core/global_index/indexed_split_test.cpp
```diff
@@ -137,25 +137,35 @@ TEST(IndexedSplitTest, TestIndexedSplitWithScore) {
...
-        IndexedSplitImpl split(/*data_split=*/nullptr, /*row_ranges=*/std::vector<Range>());
-        ASSERT_NOK_WITH_MSG(split.Validate(), "IndexedSplit must have non-empty row ranges");
-    }
     {
         std::vector<Range> row_ranges = {Range(10, 12), Range(30, 31)};
         std::vector<float> scores = {10.01f, 10.11f, 10.21f, -30.01f};
-        IndexedSplitImpl split(/*data_split=*/nullptr, row_ranges, scores);
+        IndexedSplitImpl split(data_split, row_ranges, scores);
         ASSERT_NOK_WITH_MSG(split.Validate(),
                             "Scores length does not match row ranges in indexed split.");
```
lxy-9602: add invalid test for: row ranges mismatch data files

PR #30, opened 2025-12-23
refactor(global_index): remove global range awareness from plugin

lxy-9602:
### Purpose
The globalindex plugin no longer needs to handle global row ranges.
Range management are now fully delegated to the framework layer, simplifying the plugin interface and implementations.

Linked issue: #5


lucasfang:
+1

Issue #42, opened 2026-01-04
[Bug] fix orc read timestamp under debian


lszskye:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/paimon-cpp/issues) and found nothing similar.

### Paimon-cpp version

main

### Minimal reproduce step

[https://github.com/alibaba/paimon-cpp/pull/37](url)

### What doesn't meet your expectations?

We recently received a user report that, on Debian systems, reading timestamp data from the same ORC file yields different results compared to Ubuntu systems. 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

lszskye:
thanks to @SGZW 
for your extremely thorough investigation and proactive contributions!

PR #43, opened 2026-01-04
fix: fix orc read timestamp under debian

lszskye:
### Purpose

Linked issue: #42

We recently received a user report that, on Debian systems, reading timestamp data from the same ORC file yields different results compared to Ubuntu systems. Through debugging, we found that if the timezone recorded in the ORC file's footer is not GMT and the timestamp is earlier than 1900-12-31 15:54:17, Debian and Ubuntu systems produce different results. To address this, we have updated our dataset to ensure all ORC files use GMT as their timezone. Additionally, we have explicitly set the timezone to GMT in both the ORC reader and writer.

### Tests

ScanInteTest/ReadInteTest

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -552,6 +552,9 @@ macro(build_avro)
     get_target_property(AVRO_ZLIB_INCLUDE_DIR zlib INTERFACE_INCLUDE_DIRECTORIES)
     get_filename_component(AVRO_ZLIB_ROOT "${AVRO_ZLIB_INCLUDE_DIR}" DIRECTORY)
 
+    get_target_property(AVRO_FMT_INCLUDE_DIR fmt INTERFACE_INCLUDE_DIRECTORIES)
+    get_filename_component(AVRO_FMT_ROOT "${AVRO_FMT_INCLUDE_DIR}" DIRECTORY)
+
     set(AVRO_CMAKE_CXX_FLAGS "${EP_CXX_FLAGS} -Wno-error")
     set(AVRO_CMAKE_C_FLAGS "${EP_C_FLAGS} -Wno-error")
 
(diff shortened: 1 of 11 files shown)


lxy-9602:
@SGZW PTAL, Thanks!

src/paimon/format/orc/orc_format_writer.cpp
```diff
@@ -239,6 +239,7 @@ Result<::orc::WriterOptions> OrcFormatWriter::PrepareWriterOptions(
                            OptionsUtils::GetValueFromMap<size_t>(options, ORC_ROW_INDEX_STRIDE,
                                                                  DEFAULT_ROW_INDEX_STRIDE));
     writer_options.setRowIndexStride(row_index_stride);
+    writer_options.setTimezoneName("GMT");
```
SGZW: Great work! You can refer to the changes from the Arrow community or our discussion in the issue.

src/paimon/format/orc/orc_file_batch_reader.cpp
```diff
@@ -241,6 +241,7 @@ Result<::orc::RowReaderOptions> OrcFileBatchReader::CreateRowReaderOptions(
         include_fields.push_back(field_name);
     }
     row_reader_options.include(include_fields);
+    row_reader_options.setTimezoneName("GMT");
```
SGZW: Great work! You can refer to the changes from the Arrow community or our discussion in the issue.

src/paimon/format/orc/orc_format_writer.cpp
lszskye: ok~

src/paimon/format/orc/orc_file_batch_reader.cpp
lszskye: ok~

lxy-9602 (approved):
+1

SGZW:
thanks for fix, pls merge it

PR #55, opened 2026-01-15
fix(build): Improve build system reliability and flexibility

Eyizoha:
### Purpose

- feat(third_party): Add checksum verification for downloaded dependencies
    - Avoid redundant downloads of dependency packages
- fix(build): Set Protobuf_ROOT for ORC
    - Prevent ORC from potentially using system protoc during compilation, which could lead to mismatch between generated code and library
- fix(build): Set the expected output path
    - Replace undefined path variables and unify artifact placement paths
- feat(build): Add option to disable C++11 ABI
    - Facilitate integration for projects not using C++11 ABI
- feat(build): Add build and package script
    - Provide a simple automated build and packaging script

### Tests

None

### API and Format

No impact on API or format

### Documentation

No new features introduced

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 *.jar filter=lfs diff=lfs merge=lfs -text
 *.tar filter=lfs diff=lfs merge=lfs -text
 *.gz filter=lfs diff=lfs merge=lfs -text
+79d01717-8380-4504-86e1-387e6c058d0a filter=lfs diff=lfs merge=lfs -text
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,7 @@ repos:
         exclude: (^test/test_data/.*|^third_party/.*)
       - id: check-yaml
       - id: check-added-large-files
+        args: ["--maxkb=5120", "--enforce-all"]
 
   - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v20.1.8
--- a/include/paimon/memory/memory_pool.h
+++ b/include/paimon/memory/memory_pool.h
@@ -74,10 +74,25 @@ class PAIMON_EXPORT MemoryPool {
     ///
     /// Releases a previously allocated memory block back to the pool.
     /// The size must match the size used during allocation.
+    /// The alignment used during allocation is not provided, subclass should store it
+    /// by themselves if needed.
     ///
     /// @param p Pointer to the memory block to deallocate.
     /// @param size Size of the memory block (must match allocation size).
     virtual void Free(void* p, uint64_t size) = 0;
+
+    /// Deallocate memory back to the pool with specified alignment.
+    ///
+    /// Releases a previously allocated memory block back to the pool.
+    /// The size and alignment must match the values used during allocation.
+    /// Subclass can override this method to optimize deallocation based on alignment.
+    ///
+    /// @param p Pointer to the memory block to deallocate.
+    /// @param size Size of the memory block (must match allocation size).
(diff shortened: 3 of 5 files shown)


cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -649,6 +714,7 @@ macro(build_orc)
         "-DZSTD_HOME=${ORC_ZSTD_ROOT}"
         "-DZLIB_HOME=${ORC_ZLIB_ROOT}"
         "-DPROTOBUF_HOME=${ORC_PROTOBUF_ROOT}"
+        "-DProtobuf_ROOT=${ORC_PROTOBUF_ROOT}"
```
zjw1111: Maybe it's already right with `-DPROTOBUF_HOME`. See orc's code, all `PROTOBUF_HOME`、`PROTOBUF_ROOT`、`Protobuf_ROOT` are linked to `PROTOBUF_HOME`.

Is there any situation in your development environment, which set `-DPROTOBUF_HOME` is not ok, but set `-DProtobuf_ROOT` is ok? 

Image

Eyizoha: Yes, in my development environment, the system has a higher incompatible version of protobuf installed (3.9.1). During the build process, ORC unexpectedly uses the system's protoc to generate code, but at the linking stage, it uses the project-specified version of libprotobuf, which leads to undefined reference errors.

Based on my analysis, the issue is that find_package(Protobuf CONFIG) in ORC's FindProtobuf.cmake will prioritize finding the system's protobuf and use the system's protoc. PROTOBUF_HOME is not among CMake's standard search variables, so if only PROTOBUF_HOME is set, find_package() will ignore it and search in system paths instead.

After I set Protobuf_ROOT, the problem was resolved. This issue should be reproducible if you install a higher version of protobuf on the system (e.g., 3.9.1).

build_and_package.sh
```diff
@@ -0,0 +1,109 @@
...
+if [ "$PACKAGE" = true ]; then
+    echo "Step 3: Packaging..."
+    mkdir -p "$OUTPUT_DIR"
+    PACKAGE_DIR="$OUTPUT_DIR/$BUILD_NAME"
+    rm -rf "$PACKAGE_DIR"
+    mkdir -p "$PACKAGE_DIR/include" "$PACKAGE_DIR/lib"
+
+    cp -r include/* "$PACKAGE_DIR/include/"
+    cp -r "$BUILD_DIR"/$BUILD_TYPE/*.a "$PACKAGE_DIR/lib/"
+    cp -r "$BUILD_DIR"/$BUILD_TYPE/*.so "$PACKAGE_DIR/lib/"
```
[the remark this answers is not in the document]
Eyizoha: Done

third_party/versions.txt
```diff
@@ -30,51 +30,65 @@ THIRDPARTY_MIRROR_URL=
 
 PAIMON_ZLIB_BUILD_VERSION=1.3.1
 PAIMON_ZLIB_BUILD_SHA256_CHECKSUM=9a93b2b7dfdac77ceba5a558a580e74667dd6fede4585b91eefb60f03b72df23
+PAIMON_ZLIB_PKG_NAME=zlib-1.3.1.tar.gz
```
lucasfang: To reduce redundancy, consider deriving *_PKG_NAME from *_BUILD_VERSION (e.g., zlib- $ {PAIMON_ZLIB_BUILD_VERSION}.tar.gz) instead of hardcoding the version twice？

Eyizoha: Done

PR #56, opened 2026-01-15
feat(catalog, schema): Add existence check and schema improvements

Eyizoha:
### Purpose

- feat(catalog): Support db/tbl exists check
    - Add methods to Catalog interface for checking database and table existence
- refactor(schema): Use TableSchema directly
    - Refactor Schema interface implementation to directly inherit the interface using TableSchema, similar to FileSystemCatalog and other classes, avoiding one layer of indirection overhead
- feat(schema): Add GetJsonSchema for Schema interface
    - Add method to Schema interface for retrieving schema in JSON format, which can be conveniently used in places like ReadContextBuilder::SetTableSchema

### Tests

None

### API and Format

Three new interface methods added to API, no impact on existing interfaces or format

### Documentation

No new features introduced


lxy-9602 (approved):
+1

PR #58, opened 2026-01-16
fix: lazy create merge function in merge file split read

zjw1111:
### Purpose

Avoid false "unsupported aggregation method" error reports in scenarios such as PK+DV where reading does not actually require MOR (merge-on-read).

Original version: When reading the PK table, if an unsupported aggregation method is encountered, an error is directly reported during the creation of the reader, regardless of whether the MOR is actually needed during the data reading process.

Updated to: The validation of the aggregation method support will only occur when the DataSplit data requires MOR, avoiding false error reports in scenarios such as PK+DV where reading does not actually require MOR.

### Tests

TEST_P(ScanAndReadInteTest, TestWithPKWithDvWithInvalidAggregateBatchScanSnapshot3)
TEST_P(ScanAndReadInteTest, TestWithPKWithMorWithInvalidAggregateBatchScanSnapshot3)


lxy-9602 (approved):
+1

Issue #34, opened 2025-12-24
[Feature] support projection using field id.


jichen20210919:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/paimon-cpp/issues) and found nothing similar.

### Motivation

paimon-cpp currently only support select columns using column name, propose to support select columns using field id as well.

### Anything else?

plan to extend ReadContext to support column selection based on field id.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #59, opened 2026-01-17
support specify field ids for table read context

jichen20210919:
### Purpose
paimon-cpp currently only support select columns using column name, the patch is used to support select columns using field id as well.

Linked issue:  #34

### Tests

add UT internal_read_context_test.cpp#TestReadWithSpecifiedFieldId
update UT read_context_test.cpp
### API and Format

add SetReadFieldIds(const std::vector<int32_t>& read_field_ids) in class ReadContextBuilder

### DIFF
--- a/include/paimon/read_context.h
+++ b/include/paimon/read_context.h
@@ -25,6 +25,7 @@
 
 #include "paimon/predicate/predicate.h"
 #include "paimon/result.h"
+#include "paimon/special_field_id.h"
 #include "paimon/type_fwd.h"
 #include "paimon/visibility.h"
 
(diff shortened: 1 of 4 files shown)


include/paimon/read_context.h
```diff
@@ -142,6 +148,17 @@ class PAIMON_EXPORT ReadContextBuilder {
...
+    ///
+    /// If not set, all fields from the table schema will be read. This is useful for
+    /// projection pushdown to reduce I/O and improve performance by reading only
+    /// the required columns.
+    ///
+    /// @param read_field_ids Vector of field ids to read from the table.
+    /// @return Reference to this builder for method chaining.
+    /// @note Currently supports top-level field selection. Future versions may support
+    ///       nested field selection using ArrowSchema for more granular projection
+    ReadContextBuilder& SetReadFieldIds(const std::vector<int32_t>& read_field_ids);
```
lxy-9602: Consider add comments to clarify the behavior when both SetReadFieldIds and SetReadSchema are called. E.g., SetReadFieldIds and SetReadSchema are mutually exclusive. Calling both may result in undefined behavior. Only one should be used.

jichen20210919: thanks, will update according to your suggestion.

PR #61, opened 2026-01-19
fix: fix typo in catalog

zjw1111:
### Purpose

ListTables(const std::string& database_names) -> ListTables(const std::string& db_name)
DataBaseExists -> DatabaseExists

### API and Format

DataBaseExists -> DatabaseExists


lucasfang (approved):

lucasfang:
+1

PR #87, opened 2026-01-27
fix(build): Check target existence before add_dependencies

xylaaaaa:
When building with PAIMON_BUILD_SHARED=OFF, the shared library targets do not exist, causing add_dependencies to fail. This fix filters out non-existent targets from ARG_DEPENDENCIES before calling add_dependencies.

### Purpose

Linked issue: close #xxx


zjw1111 (approved):
+1

PR #94, opened 2026-01-29
feat: divide range larger than cache range limit for performance

lucasfang:
### Purpose

The changes ensure that individual ranges larger than the configured range_size_limit are automatically divided into smaller chunks before caching, preventing oversized cache entries that could negatively impact performance.


lxy-9602 (approved):
+1

PR #98, opened 2026-01-30
fix: bugfix for predicate batch reader not check predicate nullptr

lucasfang:
### Purpose

Predicate nullptr is not checked. This may cause coredump when EnablePredicateFilter is on and no predicate is set.

### Tests

PredicateBatchReaderTest, TestInvalidInput


lxy-9602 (approved):
+1

PR #119, opened 2026-02-05
refactor(memory): Refactor memory pool adaptor

### LINKED ISSUE #126: [Feature] Decoupling BatchReader from returned batch lifecycles
### Search before asking

### Motivation

Currently, due to the Memory Pool Adaptor implementation, the lifecycle of BatchReader must outlive the ReadBatch it returns. Otherwise, batch destruction will cause dangling pointer access. Specifically, to adapt the Paimon memory pool to Arrow or ORC memory pool interfaces, the BatchReader internally wraps the incoming Paimon memory pool using an adaptor and passes it to Arrow or ORC APIs. As a result, the memory pool object referenced by PoolBuffer inside the batch is not the original memory pool object directly passed to the Paimon API, but rather the adaptor object held internally by the BatchReader.

This requires users to keep the BatchReader alive until all returned batches (specifically, the internal PoolBuffers) are released. This constraint couples Reader and Batch in a way that is unintuitive and error-prone for users, and may also limit certain resource optimization techniques.

### Solution

If we could decouple the reader from the batches it returns, it would be much more convenient and elegant for users. However, such a refactoring would inevitably affect all places where memory pools are used, including not only the core components but also existing and future plugins. Therefore, a minimally invasive solution is preferred. An ideal approach would be to transfer adaptor management from the reader to the memory pool itself, allowing adaptors to share the same lifecycle as the pool. This would resolve the coupling issue while maintaining compatibility with existing code.

### Anything else?

_No response_

Eyizoha:
### Purpose
Linked issue: close #126 

Background and Motivation:

Currently, due to the Memory Pool Adaptor implementation, the lifecycle of BatchReader must outlive the ReadBatch it returns. Otherwise, batch destruction will cause dangling pointer access. Specifically, to adapt the Paimon memory pool to Arrow or ORC memory pool interfaces, the BatchReader internally wraps the incoming Paimon memory pool using an adaptor and passes it to Arrow or ORC APIs. As a result, the memory pool object referenced by PoolBuffer inside the batch is not the original memory pool object directly passed to the Paimon API, but rather the adaptor object held internally by the BatchReader.

This requires users to keep the BatchReader alive until all returned batches (specifically, the internal PoolBuffers) are released. This constraint couples Reader and Batch in a way that is unintuitive and error-prone for users, and may also limit certain resource optimization techniques.

Solution:

To address this issue, this PR refactors the memory pool adaptor pattern by placing the adaptor directly into the Paimon memory pool, ensuring they share consistent lifecycles. This allows readers to use the adaptors directly without creating and holding them independently. Adaptor creation is lazy/one-time.

This refactoring doesn't prevent existing plugins from continuing to use the original pattern—customizing their own adapters and having readers hold them directly. The downside is that this reintroduces coupling between the plugin's reader and batch components. So if you don't mind the coupling, existing plugins should work as-is without any changes.

### Tests

Existing tests cover this change. Additionally, manual testing has verified the decoupling of reader and batch lifecycles.
### API and Format

This change introduce a new public methods to the memory pool:  `AsSpecifiedMemoryPool`, returns a pointer to the specified adaptor. 
Additionally, introduces `MemoryPoolAdaptor<T>` as a CRTP base class for implementing custom memory pool adaptors. Subclasses must:
- Inherit from `MemoryPoolAdaptor<Subclass>`
- Implement a static `Identifier()` method
- Provide a constructor accepting `MemoryPool&`

The change maintains forward compatibility.

### Documentation

No new documentation required.

### DIFF
--- a/src/paimon/format/parquet/file_reader_wrapper_test.cpp
+++ b/src/paimon/format/parquet/file_reader_wrapper_test.cpp
@@ -28,6 +28,7 @@
 #include "arrow/io/caching.h"
 #include "arrow/memory_pool.h"
 #include "gtest/gtest.h"
+#include "paimon/common/utils/arrow/arrow_memory_pool_adaptor.h"
 #include "paimon/common/utils/arrow/status_utils.h"
 #include "paimon/common/utils/path_util.h"
 #include "paimon/format/parquet/parquet_field_id_converter.h"
--- a/src/paimon/format/parquet/parquet_format_writer_test.cpp
+++ b/src/paimon/format/parquet/parquet_format_writer_test.cpp
@@ -33,6 +33,7 @@
 #include "arrow/ipc/api.h"
 #include "arrow/memory_pool.h"
 #include "gtest/gtest.h"
+#include "paimon/common/utils/arrow/arrow_memory_pool_adaptor.h"
 #include "paimon/common/utils/date_time_utils.h"
 #include "paimon/common/utils/path_util.h"
 #include "paimon/format/parquet/parquet_field_id_converter.h"


include/paimon/memory/memory_pool.h
```diff
@@ -110,6 +120,30 @@ class PAIMON_EXPORT MemoryPool {
...
+    /// internally by this memory pool instance. This method is thread-safe.
+    ///
+    /// @return Pointer to the ORC memory pool adapter, or nullptr if ORC support is unavailable.
+    /// @note IMPORTANT: This method should only be used in ORC format-related code.
+    ///       The default implementation requires the ORC format library (paimon_orc_file_format)
+    ///       to be linked, as it depends on the ORC adaptor factory registered by that library.
+    ///       If the ORC format library is not linked, this method returns nullptr.
+    /// @note Subclasses can override this method to provide custom ORC integration.
+    virtual ::orc::MemoryPool* AsOrcMemoryPool();
+
```
lxy-9602: The memory pool should not depend on plugin-specific types. Currently, it requires knowledge of concrete reader/format classes, which breaks modularity. A user's custom plugin shouldn't force changes in core memory headers. May need to refactor for better decoupling between memory management and data readers.

include/paimon/memory/memory_pool.h
```diff
@@ -110,6 +117,37 @@ class PAIMON_EXPORT MemoryPool {
...
+    /// Adapts this memory pool to the Arrow memory pool interface.
+    ///
+    /// Returns a pointer to an Arrow memory pool adapter that wraps this memory pool.
+    /// The returned pointer is lazily created on first call and its lifetime is managed
+    /// internally by this memory pool instance.
+    /// This method is thread-safe.
+    ///
+    /// @return Pointer to the Arrow memory pool adapter.
+    /// @note Subclasses can override this method to provide custom Arrow integration.
+    virtual arrow::MemoryPool* AsArrowMemoryPool();
```
lucasfang: paimon interface only has c arrow api, arrow::MemoryPool is C++ symbol which should not be in here

Eyizoha: Done

PR #178, opened 2026-03-12
fix(cmake): fix factory registry in example

lucasfang:
### Purpose

Added -Wl,--no-as-needed to CMAKE_EXE_LINKER_FLAGS to prevent linker from dropping shared library dependencies that are specified via target_link_libraries but not directly referenced in the code. This ensures expected shared objects show up in the executable’s ldd output.


zjw1111 (approved):
+1

PR #212, opened 2026-04-08
feat(btree): Intro b-tree global index and add tests for java compatibility.

### LINKED ISSUE #38: [Feature] Support btree global index.
### Search before asking

### Motivation

Ref https://github.com/apache/paimon/pull/6869

### Solution

_No response_

### Anything else?

_No response_

ChaomingZhangCN:
### Purpose

Linked issue: close #38 

### Tests

Note: Google Test filter = BTree*
[==========] Running 47 tests from 6 test suites.
[----------] Global test environment set-up.
[----------] 7 tests from BTreeIndexMetaTest
[----------] 7 tests from BTreeFileFooterTest
[----------] 16 tests from BTreeGlobalIndexerTest
[----------] 7 tests from BTreeGlobalIndexWriterTest
[----------] 5 tests from BTreeGlobalIndexIntegrationTest
[----------] 5 tests from BTreeCompatibilityTest
[----------] Global test environment tear-down
[==========] 47 tests from 6 test suites ran. (26 ms total)
[  PASSED  ] 47 tests.

### Generative AI tooling

Generated-by: Claude Code


lxy-9602:
Impressive work on this complex feature. Thanks for the contribution — review coming up.

src/paimon/common/file_index/file_index_reader.cpp
```diff
@@ -98,4 +98,56 @@ Result<std::shared_ptr<FileIndexResult>> FileIndexReader::VisitNotIn(
...
 }
+
+Result<std::shared_ptr<FileIndexResult>> FileIndexReader::VisitAnd(
+    const std::vector<Result<std::shared_ptr<FileIndexResult>>>& children) {
+    if (children.empty()) {
+        return Status::Invalid("VisitAnd called with no children");
+    }
+
+    // Start with the first child
+    auto result = children[0];
```
lxy-9602: Let's standardize on `PAIMON_ASSIGN_OR_RAISE` and `PAIMON_RETURN_NOT_OK` throughout these changes.

include/paimon/predicate/function_visitor.h
```diff
@@ -74,5 +74,41 @@ class PAIMON_EXPORT FunctionVisitor {
 
     /// Evaluates whether string values like the given string.
     virtual Result<T> VisitLike(const Literal& literal) = 0;
+
+    /// Evaluates the BETWEEN predicate with the given lower and upper bounds.
+    virtual Result<T> VisitBetween(const Literal& from, const Literal& to) {
+        // Default implementation: BETWEEN is equivalent to >= AND <=
```
lxy-9602, line 80: I'm wondering if `VisitBetween` is strictly necessary, since `PredicateBuilder::Between` is achieved by `AND` of `GreaterOrEqual` and `LessOrEqual`. And `VisitAnd` and `VisitOr` for FileIndexReader is only used for `Between`?

src/paimon/common/file_index/empty/empty_file_index_reader.h
```diff
@@ -65,6 +65,37 @@ class EmptyFileIndexReader : public FileIndexReader {
         const std::vector<Literal>& literals) override {
         return FileIndexResult::Skip();
     }
+
+    Result<std::shared_ptr<FileIndexResult>> VisitNotEqual(const Literal& literal) override {
+        // Empty file has no data, so all records are not equal to any value
```
lxy-9602: If empty means the file contains no data, then intuitively it should be safe to skip regardless of the predicate — since there's nothing to read.

Could you help me understand why additional implementation is needed here?

src/paimon/common/file_index/file_index_reader.cpp
```diff
@@ -98,4 +98,56 @@ Result<std::shared_ptr<FileIndexResult>> FileIndexReader::VisitNotIn(
...
+        PAIMON_ASSIGN_OR_RAISE(current, current->And(child));
+    }
+
+    return current;
+}
+
+Result<std::shared_ptr<FileIndexResult>> FileIndexReader::VisitOr(
+    const std::vector<Result<std::shared_ptr<FileIndexResult>>>& children) {
+    if (children.empty()) {
+        return Status::Invalid("VisitOr called with no children");
```
lxy-9602, line 125: Using `Result<T>` as a function parameter is generally not recommended — `Result` should be used for return types, not inputs. Also, before adding more visitor methods like `VisitOr`, could we first confirm whether `FileIndexReader` truly needs this interface?

src/paimon/common/utils/roaring_bitmap64.cpp
```diff
@@ -227,6 +228,10 @@ bool RoaringBitmap64::operator==(const RoaringBitmap64& other) const noexcept {
 PAIMON_UNIQUE_PTR<Bytes> RoaringBitmap64::Serialize(MemoryPool* pool) const {
     GetRoaringBitmap(roaring_bitmap_).runOptimize();
     auto& bitmap = GetRoaringBitmap(roaring_bitmap_);
+    // Use default pool if no pool is provided
```
lxy-9602, line 231: Could you please also fix this issue in `RoaringBitmap32::Serialize`?

src/paimon/common/utils/roaring_navigable_map64.h
```diff
@@ -0,0 +1,238 @@
...
+namespace paimon {
+
+/**
+ * A compressed bitmap for 64-bit integer aggregated by tree.
+ * This is a wrapper around RoaringBitmap64 that provides additional functionality
+ * and a more convenient interface.
+ */
+class PAIMON_EXPORT RoaringNavigableMap64 {
+ public:
+    /// Default constructor creates an empty bitmap
```
lxy-9602: Note that Java's `RoaringNavigableMap64` is functionally equivalent to C++'s `RoaringBitmap64`.

Since we already have a well-tested implementation in C++, there’s no need to re-implement the same semantics here.

include/paimon/defs.h
```diff
@@ -414,6 +414,27 @@ struct PAIMON_EXPORT Options {
...
+    static const char BTREE_INDEX_CACHE_SIZE[];
+    /// "btree-index.high-priority-pool-ratio" - The high priority pool ratio to use for BTreeIndex.
+    /// Default value is 0.1.
+    static const char BTREE_INDEX_HIGH_PRIORITY_POOL_RATIO[];
+    /// "btree-index.records-per-range" - The expected number of records per BTree Index File.
+    /// Default value is 1000000.
+    static const char BTREE_INDEX_RECORDS_PER_RANGE[];
+    /// "btree-index.build.max-parallelism" - The max parallelism of Flink/Spark for building
+    /// BTreeIndex. Default value is 4096.
+    static const char BTREE_INDEX_BUILD_MAX_PARALLELISM[];
```
lxy-9602: Please make these options in `globalindex/btree/btree_index_options.h`, same as java.

src/paimon/common/sst/sst_file_reader.h
```diff
@@ -38,15 +37,20 @@ class SstFileIterator;
...
 /// queries. Note that this class is NOT thread-safe.
-class SstFileReader {
+class PAIMON_EXPORT SstFileReader {
  public:
     static Result<std::shared_ptr<SstFileReader>> Create(
-        const std::shared_ptr<InputStream>& input, MemorySlice::SliceComparator comparator,
-        const std::shared_ptr<CacheManager>& cache_manager,
-        const std::shared_ptr<MemoryPool>& pool);
+        const std::shared_ptr<MemoryPool>& pool, const std::shared_ptr<InputStream>& input,
+        const BlockHandle& index_block_handle,
```
lxy-9602: Could you please put `pool` to the last param?

src/paimon/common/sst/sst_file_writer.cpp
```diff
@@ -83,10 +87,7 @@ Result<std::shared_ptr<BloomFilterHandle>> SstFileWriter::WriteBloomFilter() {
     return handle;
 }
 
-Status SstFileWriter::WriteFooter(const BlockHandle& index_block_handle,
-                                  const std::shared_ptr<BloomFilterHandle>& bloom_filter_handle) {
-    BlockFooter footer(index_block_handle, bloom_filter_handle);
-    auto slice = footer.WriteBlockFooter(pool_.get());
+Status SstFileWriter::WriteSlice(const MemorySlice& slice) {
     auto data = slice.ReadStringView();
```
lxy-9602, line 86: What's the difference between `WriteSlice(slice)` and `Write(slice)`? Seems only one func is enough.

src/paimon/common/memory/memory_segment_utils.cpp
```diff
@@ -63,7 +63,9 @@ PAIMON_UNIQUE_PTR<Bytes> MemorySegmentUtils::CopyToBytes(const std::vector<Memor
                                                          MemoryPool* pool) {
     assert(pool);
     auto bytes = Bytes::AllocateBytes(num_bytes, pool);
-    CopyToBytes(segments, offset, bytes.get(), 0, num_bytes);
+    if (num_bytes > 0) {
+        CopyToBytes(segments, offset, bytes.get(), 0, num_bytes);
+    }
```
lxy-9602: I'm not sure why this check is being added here. If we decide that input validation is necessary for this method, then all similar functions in MemorySegmentUtils — such as other Copy overloads — should have the same protection.

src/paimon/common/lookup/sort/sort_lookup_store_footer.cpp
lxy-9602, line 30: In java `if (offset && size && expected_entries)`. Please see `SortLookupStoreFooter.java` as ref.

lxy-9602: Also `// skip padding` (`sliceInput.setPosition(ENCODED_LENGTH - 4);`) in java is omitted?

src/paimon/common/sst/sst_file_io_test.cpp
```diff
@@ -141,8 +144,19 @@ TEST_P(SstFileIOTest, TestSimple) {
...
+    ASSERT_OK(in->Read(footer_bytes->data(), footer_bytes->size()));
+    auto footer_segment = MemorySegment::Wrap(std::move(footer_bytes));
+    auto footer_slice = MemorySlice::Wrap(footer_segment);
+    auto footer_input = footer_slice.ToInput();
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<SortLookupStoreFooter> read_footer,
+                         SortLookupStoreFooter::ReadSortLookupStoreFooter(&footer_input));
+    ASSERT_OK_AND_ASSIGN(
+        auto reader,
+        SstFileReader::Create(pool_, in, read_footer->GetIndexBlockHandle(),
+                              read_footer->GetBloomFilterHandle(), comparator_, cache_manager_));
```
lxy-9602: Consider reusing `SortLookupStoreFactory::CreateReader()` and extracting the inner `SstFileReader` creation logic to avoid code duplication.

src/paimon/common/io/cache/cache.h
```diff
@@ -49,9 +49,22 @@ class Cache {
...
+    void Invalidate(const std::shared_ptr<CacheKey>& key) override;
+    void InvalidateAll() override;
+    size_t Size() const override;
+};
+
 class CacheValue {
  public:
-    CacheValue(const MemorySegment& segment, CacheCallback callback)
+    explicit CacheValue(const MemorySegment& segment, CacheCallback callback = nullptr)
         : segment_(segment), callback_(std::move(callback)) {}
```
lxy-9602, line 55: Please avoid default parameter values in public function signatures.

src/paimon/common/io/cache/cache.h
```diff
@@ -49,9 +49,22 @@ class Cache {
     virtual size_t Size() const = 0;
 };
 
+class PAIMON_EXPORT NoCache : public Cache {
+ public:
+    Result<std::shared_ptr<CacheValue>> Get(
+        const std::shared_ptr<CacheKey>& key,
+        std::function<Result<std::shared_ptr<CacheValue>>(const std::shared_ptr<CacheKey>&)>
```
lxy-9602: Now that we have LRUCache support, I'd like to confirm: is NoCache still necessary?

src/paimon/common/global_index/btree/btree_index_meta.h
```diff
@@ -0,0 +1,67 @@
...
+                                                       paimon::MemoryPool* pool);
+    std::shared_ptr<Bytes> Serialize(paimon::MemoryPool* pool) const;
+
+ public:
+    BTreeIndexMeta(const std::shared_ptr<Bytes>& first_key, const std::shared_ptr<Bytes>& last_key,
+                   bool has_nulls)
+        : first_key_(first_key), last_key_(last_key), has_nulls_(has_nulls) {}
+
+    std::shared_ptr<Bytes> FirstKey() const {
+        return first_key_;
```
lxy-9602, line 40: Consider return `const std::shared_ptr<Bytes>&` for reduce shared_ptr copy cost.

src/paimon/common/global_index/btree/btree_index_meta.cpp
```diff
@@ -0,0 +1,69 @@
...
+    if (last_key_) {
+        output.WriteBytes(last_key_);
+    }
+
+    // Write has_nulls
+    output.WriteValue(static_cast<int8_t>(has_nulls_ ? 1 : 0));
+
+    auto slice = output.ToSlice();
+    return slice.CopyBytes(pool);
+}
```
lxy-9602, line 66: Just return `getHeapMemory()` is ok as you have just compute the total_size.

src/paimon/common/global_index/btree/btree_index_meta.cpp
```diff
@@ -0,0 +1,69 @@
...
+    return std::make_shared<BTreeIndexMeta>(first_key, last_key, has_nulls);
+}
+
+std::shared_ptr<Bytes> BTreeIndexMeta::Serialize(paimon::MemoryPool* pool) const {
+    // Calculate total size: first_key_len(4) + first_key + last_key_len(4) + last_key +
+    // has_nulls(1)
+    int32_t first_key_size = first_key_ ? first_key_->size() : 0;
+    int32_t last_key_size = last_key_ ? last_key_->size() : 0;
+    int32_t total_size = 4 + first_key_size + 4 + last_key_size + 1;
+
```
lxy-9602, line 47: int32_t total_size = Size();

src/paimon/common/global_index/btree/btree_index_meta.cpp
```diff
@@ -0,0 +1,69 @@
...
+    std::shared_ptr<Bytes> first_key;
+    if (first_key_len) {
+        first_key = input.ReadSlice(first_key_len).CopyBytes(pool);
+    }
+    auto last_key_len = input.ReadInt();
+    std::shared_ptr<Bytes> last_key;
+    if (last_key_len) {
+        last_key = input.ReadSlice(last_key_len).CopyBytes(pool);
+    }
+    auto has_nulls = input.ReadByte() == 1;
```
lxy-9602: Please cast to int8_t.

src/paimon/common/global_index/btree/btree_index_meta_test.cpp
```diff
@@ -0,0 +1,204 @@
...
+    // Serialize
+    auto serialized = meta->Serialize(pool_.get());
+    ASSERT_NE(serialized, nullptr);
+
+    // Deserialize
+    auto deserialized = BTreeIndexMeta::Deserialize(serialized, pool_.get());
+    ASSERT_NE(deserialized, nullptr);
+
+    // Verify keys are null
+    EXPECT_EQ(deserialized->FirstKey(), nullptr);
```
lxy-9602: Prefer `ASSERT_*` over `EXPECT_*` when the failure of a condition would make the rest of the test meaningless or lead to undefined behavior (e.g., null pointers, invalid setup).

src/paimon/common/global_index/btree/btree_index_meta_test.cpp
```diff
@@ -0,0 +1,204 @@
...
+    auto deserialized = BTreeIndexMeta::Deserialize(serialized, pool_.get());
+    ASSERT_NE(deserialized, nullptr);
+
+    // Verify has_nulls is false
+    EXPECT_FALSE(deserialized->HasNulls());
+}
+
+TEST_F(BTreeIndexMetaTest, SerializeDeserializeWithOnlyFirstKey) {
+    // Create a BTreeIndexMeta with only first_key (edge case)
+    auto first_key = std::make_shared<Bytes>("first", pool_.get());
```
lxy-9602, line 135: Is there any scenario in which this case could occur?

lxy-9602: Also SerializeDeserializeWithOnlyLastKey().

src/paimon/common/global_index/btree/btree_file_footer.h
```diff
@@ -0,0 +1,80 @@
...
+        return index_block_handle_;
+    }
+
+    std::shared_ptr<BlockHandle> GetNullBitmapHandle() const {
+        return null_bitmap_handle_;
+    }
+
+ public:
+    static constexpr int32_t MAGIC_NUMBER = 0x50425449;
+    static constexpr int32_t CURRENT_VERSION = 1;
```
lxy-9602, line 70: For new code, please use `kMagicNumber`.

src/paimon/common/global_index/btree/btree_file_footer.h
```diff
@@ -0,0 +1,80 @@
...
+#include "paimon/common/memory/memory_slice_output.h"
+#include "paimon/common/sst/block_handle.h"
+#include "paimon/common/sst/bloom_filter_handle.h"
+
+namespace paimon {
+/// The Footer for BTree file.
+class BTreeFileFooter {
+ public:
+    static Result<std::shared_ptr<BTreeFileFooter>> Read(MemorySliceInput& input);
+    static MemorySlice Write(const std::shared_ptr<BTreeFileFooter>& footer, MemoryPool* pool);
```
lxy-9602, line 31: For out param, please use `*` rather than `&`.

src/paimon/common/global_index/btree/btree_file_footer.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+Result<std::shared_ptr<BTreeFileFooter>> BTreeFileFooter::Read(MemorySliceInput& input) {
+    // read version and verify magic number
+    PAIMON_RETURN_NOT_OK(input.SetPosition(ENCODED_LENGTH - 8));
+
+    int32_t version = input.ReadInt();
+    int32_t magic_number = input.ReadInt();
+    if (magic_number != MAGIC_NUMBER) {
+        return Status::IOError("File is not a btree index file (bad magic number)");
+    }
```
lxy-9602, line 33: Please return `Status::Invalid` with a formatted message (using `fmt`) to show the expected vs. actual magic number — helps with debugging.

src/paimon/common/global_index/btree/btree_file_footer_test.cpp
```diff
@@ -0,0 +1,197 @@
...
+                                                    null_bitmap_handle);
+
+    auto serialized = BTreeFileFooter::Write(footer, pool_.get());
+    EXPECT_EQ(serialized.Length(), BTreeFileFooter::ENCODED_LENGTH);
+
+    auto input = serialized.ToInput();
+    auto deserialized = BTreeFileFooter::Read(input);
+    ASSERT_OK(deserialized.status());
+    auto deserialized_footer = deserialized.value();
+
```
lxy-9602, line 50: ASSERT_OK_AND_ASSIGN

src/paimon/common/global_index/btree/btree_file_footer_test.cpp
```diff
@@ -0,0 +1,197 @@
...
+    ASSERT_NE(ib_handle, nullptr);
+    EXPECT_EQ(ib_handle->Offset(), 200);
+    EXPECT_EQ(ib_handle->Size(), 80);
+
+    EXPECT_EQ(deserialized_footer->GetNullBitmapHandle(), nullptr);
+}
+
+TEST_F(BTreeFileFooterTest, MagicNumberVerification) {
+    auto index_block_handle = std::make_shared<BlockHandle>(200, 80);
+    auto footer = std::make_shared<BTreeFileFooter>(nullptr, index_block_handle, nullptr);
```
lxy-9602: Seems no need as `ReadWriteWithAllNullHandles` touch this case.

src/paimon/common/global_index/btree/btree_file_footer_test.cpp
```diff
@@ -0,0 +1,197 @@
...
+    output.WriteValue(static_cast<int32_t>(0));
+
+    output.WriteValue(static_cast<int32_t>(12345));  // Invalid magic number
+
+    auto serialized = output.ToSlice();
+    auto input = serialized.ToInput();
+
+    auto deserialized = BTreeFileFooter::Read(input);
+    EXPECT_FALSE(deserialized.ok());
+    EXPECT_TRUE(deserialized.status().IsIOError());
```
lxy-9602: ASSERT_NOK_WITH_MSG

src/paimon/common/global_index/btree/btree_file_footer_test.cpp
```diff
@@ -0,0 +1,197 @@
...
+}
+
+TEST_F(BTreeFileFooterTest, EncodedLength) {
+    EXPECT_EQ(BTreeFileFooter::ENCODED_LENGTH, 52);
+
+    auto bloom_filter_handle = std::make_shared<BloomFilterHandle>(100, 50, 1000);
+    auto index_block_handle = std::make_shared<BlockHandle>(200, 80);
```
lxy-9602: `ReadWriteRoundTrip` also cover this case.

src/paimon/common/global_index/btree/btree_global_index_factory.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include "paimon/common/global_index/btree/btree_global_index_factory.h"
+
+#include <utility>
+
+#include "paimon/common/global_index/btree/btree_global_indexer.h"
+namespace paimon {
+
+const char BTreeGlobalIndexerFactory::IDENTIFIER[] = "btree";
+
```
lxy-9602, line 25: Please add `-global` suffix for `GlobalIndexerFactory` (see bitmap for ref).

lxy-9602: The main reason for this separation is to distinguish it from the registration mechanism used by file-index.

src/paimon/common/global_index/btree/btree_global_indexer.h
```diff
@@ -0,0 +1,140 @@
...
+
+    static Result<std::shared_ptr<RoaringNavigableMap64>> ReadNullBitmap(
+        const std::shared_ptr<BlockCache>& cache, const std::shared_ptr<BlockHandle>& block_handle);
+
+ private:
+    std::map<std::string, std::string> options_;
+};
+
+class BTreeGlobalIndexReader : public GlobalIndexReader {
+ public:
```
lxy-9602: Consider extracting a dedicated `BTreeGlobalIndexReader` class to encapsulate the read-path logic.

src/paimon/common/global_index/btree/btree_global_index_writer.h
```diff
@@ -0,0 +1,90 @@
...
+namespace paimon {
+
+/// Writer for BTree Global Index files.
+/// This writer builds an SST file where each key maps to a list of row IDs.
+class BTreeGlobalIndexWriter : public GlobalIndexWriter {
+ public:
+    BTreeGlobalIndexWriter(const std::string& field_name, ::ArrowSchema* arrow_schema,
+                           const std::shared_ptr<GlobalIndexFileWriter>& file_writer,
+                           const std::shared_ptr<MemoryPool>& pool, int32_t block_size = 4096,
+                           int64_t expected_entries = 100000);
```
lxy-9602: Could you please avoid default parameter values in public function signatures.

src/paimon/common/global_index/btree/btree_global_index_writer.h
```diff
@@ -0,0 +1,90 @@
...
+    int32_t CompareBinaryKeys(const std::shared_ptr<Bytes>& a,
+                              const std::shared_ptr<Bytes>& b) const;
+
+ private:
+    std::string field_name_;
+    std::shared_ptr<arrow::DataType> arrow_type_;
+    std::shared_ptr<GlobalIndexFileWriter> file_writer_;
+    std::shared_ptr<MemoryPool> pool_;
+    int32_t block_size_;
+
```
lxy-9602, line 77: Since `file_writer_` may depend on the `pool_`, please declare the pool as a prior member to ensure correct destruction order (pool outlives writer).

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+      current_row_id_(0),
+      bloom_filter_(BloomFilter::Create(expected_entries, 0.01)) {
+    // Allocate memory for bloom filter and set memory segment
+    if (bloom_filter_) {
+        int64_t bloom_filter_size = bloom_filter_->ByteLength();
+        auto bloom_filter_segment =
+            MemorySegment::AllocateHeapMemory(bloom_filter_size, pool.get());
+        auto status = bloom_filter_->SetMemorySegment(bloom_filter_segment);
+        if (!status.ok()) {
+            // Failed to set memory segment for bloom filter
```
lxy-9602: If construction can fail, consider using a static `Result<T> Create(...)` factory method instead of public constructors. Also mark the actual constructor as private.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+        }
+    }
+
+    // Import schema to get the field type
+    if (arrow_schema) {
+        auto schema_result = arrow::ImportSchema(arrow_schema);
+        if (schema_result.ok()) {
+            auto schema = schema_result.ValueOrDie();
+            if (schema->num_fields() > 0) {
```
lxy-9602: Don't use `ValueOrDie()` — prefer `PAIMON_ASSIGN_OR_RAISE_FROM_ARROW` for safe error propagation. Please validate input parameters in `Create`. See other global index modules for reference.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+      file_writer_(file_writer),
+      pool_(pool),
+      block_size_(block_size),
+      null_bitmap_(std::make_shared<RoaringNavigableMap64>()),
+      has_nulls_(false),
+      current_row_id_(0),
+      bloom_filter_(BloomFilter::Create(expected_entries, 0.01)) {
+    // Allocate memory for bloom filter and set memory segment
+    if (bloom_filter_) {
+        int64_t bloom_filter_size = bloom_filter_->ByteLength();
```
lxy-9602: Does this break compatibility with Java? As far as I know, Java doesn’t support storing Bloom filters in SST files for now.

src/paimon/common/global_index/btree/btree_global_index_writer.h
lxy-9602: Similarly, the lifetimes of sst_writer_ and output_stream_ should be carefully managed.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+    auto import_result = arrow::ImportArray(arrow_array, arrow_type_);
+    if (!import_result.ok()) {
+        return Status::Invalid("Failed to import array: " + import_result.status().ToString());
+    }
+    auto array = import_result.ValueOrDie();
+
+    // Initialize SST writer on first batch
+    if (!sst_writer_) {
+        auto file_name_result = file_writer_->NewFileName(field_name_);
+        if (!file_name_result.ok()) {
```
lxy-9602: Can we create the writer in the constructor like Java does? That way, we can properly pass `BlockCompressionFactory` and other dependencies during setup.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    // Group row IDs by key value
+    // Use std::map with custom comparator for binary keys
+    // Keys are stored in binary format to match Java's serialization
+    std::map<std::shared_ptr<Bytes>, std::vector<int64_t>,
+             std::function<bool(const std::shared_ptr<Bytes>&, const std::shared_ptr<Bytes>&)>>
+        key_to_row_ids([this](const std::shared_ptr<Bytes>& a, const std::shared_ptr<Bytes>& b) {
+            return CompareBinaryKeys(a, b) < 0;
+        });
+
```
lxy-9602, line 99: Data shuffling/sorting should be done by the upstream engine, as in Java. Doing it here adds cost and breaks consistency.

Please remove local sorting logic to avoid performance risks and maintain cross-language alignment.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // Import schema to get the field type
+    if (arrow_schema) {
+        auto schema_result = arrow::ImportSchema(arrow_schema);
+        if (schema_result.ok()) {
+            auto schema = schema_result.ValueOrDie();
+            if (schema->num_fields() > 0) {
+                arrow_type_ = schema->field(0)->type();
+            }
+        }
```
lxy-9602, line 45: Important: As documented in `global_index_writer.h`, the `arrow_array` passed to `BTreeGlobalIndexWriter::AddBatch` must be a `StructArray` — even when there's only one index key column (i.e., single-field case wrapped in a struct)..

This ensures consistency across all global index implementations and prepares us for future support of composite (multi-column) keys. Accepting non-struct arrays today would create special cases that break this model.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
lxy-9602: You can add a new method like `AddBatch(::ArrowArray* arrow_array, const std::vector<iny64_t>& row_ids)`

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    return sst_writer_->Write(std::move(key), std::move(value));
+}
+
+std::shared_ptr<Bytes> BTreeGlobalIndexWriter::SerializeRowIds(
+    const std::vector<int64_t>& row_ids) {
+    // Format: [num_row_ids (VarLenLong)][row_id1 (VarLenLong)][row_id2]...
+    // Use VarLenLong for row IDs to match Java's DataOutputStream.writeVarLong
+    int32_t estimated_size = 10 + row_ids.size() * 10;  // Conservative estimate
+    auto output = std::make_shared<MemorySliceOutput>(estimated_size, pool_.get());
```
lxy-9602, line 234: java use `MemorySliceOutput` rather than `DataOutputStream`? Consider adjust comments.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    // Write index block
+    PAIMON_ASSIGN_OR_RAISE(auto index_block_handle, sst_writer_->WriteIndexBlock());
+
+    // Write bloom filter
+    PAIMON_ASSIGN_OR_RAISE(auto bloom_filter_handle, sst_writer_->WriteBloomFilter());
+
+    // Write null bitmap
+    PAIMON_ASSIGN_OR_RAISE(auto null_bitmap_handle, WriteNullBitmap(output_stream_));
+
```
lxy-9602, line 296: The write order here doesn’t match Java (null bitmap → bloom filter → index block).

Should we align with Java’s sequence to maintain format compatibility?

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+        // No data was written, return empty metadata
+        return std::vector<GlobalIndexIOMeta>();
+    }
+
+    // Flush any remaining data in the data block writer
+    PAIMON_RETURN_NOT_OK(sst_writer_->Flush());
+
+    // Write index block
+    PAIMON_ASSIGN_OR_RAISE(auto index_block_handle, sst_writer_->WriteIndexBlock());
+
```
lxy-9602, line 293: Please explicitly specify the variable type in `PAIMON_ASSIGN_OR_RAISE` instead of using `auto`.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // Write bloom filter
+    PAIMON_ASSIGN_OR_RAISE(auto bloom_filter_handle, sst_writer_->WriteBloomFilter());
+
+    // Write null bitmap
+    PAIMON_ASSIGN_OR_RAISE(auto null_bitmap_handle, WriteNullBitmap(output_stream_));
+
+    // Write BTree file footer
+    auto footer = std::make_shared<BTreeFileFooter>(
+        bloom_filter_handle, std::make_shared<BlockHandle>(index_block_handle), null_bitmap_handle);
+    auto footer_slice = BTreeFileFooter::Write(footer, pool_.get());
```
lxy-9602, line 302: `BlockHandle` only has int64_t and int32_t, please rm shard_ptr for its cost.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+                // Store as 8-byte little-endian to match Java's DataOutputStream.writeLong
+                key_bytes = std::make_shared<Bytes>(sizeof(int64_t), pool_.get());
+                memcpy(key_bytes->data(), &value, sizeof(int64_t));
+                break;
+            }
+            case arrow::Type::FLOAT: {
+                auto float_array = std::static_pointer_cast<arrow::FloatArray>(array);
+                float value = float_array->Value(i);
+                // Store as 4-byte IEEE 754 to match Java's DataOutputStream.writeFloat
+                key_bytes = std::make_shared<Bytes>(sizeof(float), pool_.get());
```
lxy-9602, line 147: Consider java use `Float.floatToIntBits`—just double-check compatibility. Also, maybe we could extract the serialization logic from `AddBatch` since it's getting a bit long?

src/paimon/common/global_index/btree/btree_global_index_writer_test.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+    // Add batch
+    auto status = writer->AddBatch(&c_array);
+    ASSERT_OK(status);
+
+    // Finish writing
+    auto result = writer->Finish();
+    ASSERT_OK(result.status());
+    auto metas = result.value();
+    ASSERT_EQ(metas.size(), 1);
```
lxy-9602, line 197: ASSERT_OK_AND_ASSIGN

src/paimon/common/global_index/btree/btree_global_index_writer_test.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+    // Verify metadata
+    const auto& meta = metas[0];
+    EXPECT_FALSE(meta.file_path.empty());
+    EXPECT_GT(meta.file_size, 0);
+
+    // Verify that metadata contains null bitmap info (has_nulls should be true)
+    EXPECT_NE(meta.metadata, nullptr);
+
```
lxy-9602, line 204: Does this comment match the code? Would a single case be sufficient here?

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+#include "paimon/file_index/bitmap_index_result.h"
+#include "paimon/global_index/bitmap_global_index_result.h"
+#include "paimon/memory/bytes.h"
+#include "paimon/predicate/literal.h"
+
+namespace paimon {
+
+// Helper function to get cache size from options with default value
+static int64_t GetBTreeIndexCacheSize(const std::map<std::string, std::string>& options) {
+    auto it = options.find(Options::BTREE_INDEX_CACHE_SIZE);
```
lxy-9602: Good use of anonymous namespace to encapsulate the helper function—limits visibility appropriately.

lxy-9602: Might want to return an error if options are misconfigured, rather than proceeding silently.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+        }
+    }
+    // Default: 128 MB
+    return 128 * 1024 * 1024;
+}
+
+// Helper function to get high priority pool ratio from options with default value
+static double GetBTreeIndexHighPriorityPoolRatio(
+    const std::map<std::string, std::string>& options) {
+    auto it = options.find(Options::BTREE_INDEX_HIGH_PRIORITY_POOL_RATIO);
```
lxy-9602: Consider use `OptionsUtils::GetValueFromMap()`.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+            return [](const MemorySlice& a, const MemorySlice& b) -> int32_t {
+                if (a.Length() < static_cast<int32_t>(sizeof(int32_t)) ||
+                    b.Length() < static_cast<int32_t>(sizeof(int32_t))) {
+                    size_t min_len =
+                        std::min(static_cast<size_t>(a.Length()), static_cast<size_t>(b.Length()));
+                    int cmp = memcmp(a.ReadStringView().data(), b.ReadStringView().data(), min_len);
+                    if (cmp != 0) return cmp < 0 ? -1 : 1;
+                    if (a.Length() < b.Length()) return -1;
+                    if (a.Length() > b.Length()) return 1;
+                    return 0;
```
lxy-9602: Could you help clarify the purpose of `a.Length() < static_cast<int32_t>(sizeof(int32_t))`? It seems unusual. The corresponding code in `KeySerializer.java` seems not have this logic?

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+                }
+                int32_t a_val = a.ReadInt(0);
+                int32_t b_val = b.ReadInt(0);
+                if (a_val < b_val) return -1;
+                if (a_val > b_val) return 1;
+                return 0;
+            };
+        case FieldType::BIGINT:
+        case FieldType::DATE:
+        case FieldType::TIMESTAMP:
```
lxy-9602: Note that DATE is semantically similar to INT and should be represented as int, not long.

lxy-9602: Also timestamp should consider precision.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+                // Store as 4-byte little-endian
+                key_bytes = std::make_shared<Bytes>(sizeof(int32_t), pool_.get());
+                memcpy(key_bytes->data(), &value, sizeof(int32_t));
+                break;
+            }
+            case arrow::Type::TIMESTAMP: {
+                auto ts_array = std::static_pointer_cast<arrow::TimestampArray>(array);
+                int64_t value = ts_array->Value(i);
+                // Store as 8-byte little-endian
+                key_bytes = std::make_shared<Bytes>(sizeof(int64_t), pool_.get());
```
lxy-9602: TIMESTAMP write should consider precision. See `KeySerializer.java` for ref.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
```diff
@@ -0,0 +1,321 @@
...
+                float value = float_array->Value(i);
+                // Store as 4-byte IEEE 754 to match Java's DataOutputStream.writeFloat
+                key_bytes = std::make_shared<Bytes>(sizeof(float), pool_.get());
+                memcpy(key_bytes->data(), &value, sizeof(float));
+                break;
+            }
+            case arrow::Type::DOUBLE: {
+                auto double_array = std::static_pointer_cast<arrow::DoubleArray>(array);
+                double value = double_array->Value(i);
+                // Store as 8-byte IEEE 754 to match Java's DataOutputStream.writeDouble
```
lxy-9602, line 154: Java use writeLong and readLong.

include/paimon/defs.h
lxy-9602: BTREE_INDEX_BUILD_MAX_PARALLELISM seems not needed by paimon-core.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+    auto null_bitmap_view = null_bitmap_slice.ReadStringView();
+
+    // Calculate CRC32C checksum
+    uint32_t crc_value = CRC32C::calculate(null_bitmap_view.data(), null_bitmap_view.size());
+
+    // Read expected CRC value (stored as uint32_t in little-endian)
+    uint32_t expected_crc_value = 0;
+    for (int i = 0; i < 4; ++i) {
+        expected_crc_value |= static_cast<uint32_t>(static_cast<uint8_t>(slice_input.ReadByte()))
+                              << (i * 8);
```
lxy-9602: Consider use `EndianSwapValue`.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+    const int64_t MAX_ROWS_FOR_SUBTRACTION = 10000000;
+
+    bool use_subtraction = (total_rows <= MAX_ROWS_FOR_SUBTRACTION) &&
+                           (null_count < static_cast<uint64_t>(total_rows * NULL_RATIO_THRESHOLD));
+
+    if (use_subtraction) {
+        RoaringNavigableMap64 result;
+        result.AddRange(Range(0, total_rows - 1));
+        result.AndNot(*null_bitmap_);
+        return result;
```
lxy-9602: If row ids are not in order (keys are sorted), we can't rely on short-circuiting logic as all range is unknown.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+                                   std::string(e.what()));
+        }
+    }
+
+    if (type == FieldType::TINYINT) {
+        try {
+            auto value = literal.GetValue<int8_t>();
+            auto bytes = std::make_shared<Bytes>(1, pool);
+            bytes->data()[0] = static_cast<char>(value);
+            return MemorySlice::Wrap(bytes);
```
lxy-9602: Why there is try and catch? Consider remove it.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+        }
+    }
+
+    if (type == FieldType::INT) {
+        try {
+            auto value = literal.GetValue<int32_t>();
+            auto bytes = std::make_shared<Bytes>(4, pool);
+            bytes->data()[0] = static_cast<char>(value & 0xFF);
+            bytes->data()[1] = static_cast<char>((value >> 8) & 0xFF);
+            bytes->data()[2] = static_cast<char>((value >> 16) & 0xFF);
```
lxy-9602: Why go through a slice conversion instead of comparing int values directly? Could we use `Literal.GetInt()` vs `MemorySlice.GetInt()` to avoid byte order issues?

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+    }
+
+    if (type == FieldType::TIMESTAMP) {
+        try {
+            auto value = literal.GetValue<int64_t>();
+            // Convert to string to match the format used in BTreeGlobalIndexWriter
+            std::string str_value = std::to_string(value);
+            auto bytes = std::make_shared<Bytes>(str_value, pool);
+            return MemorySlice::Wrap(bytes);
+        } catch (const std::exception& e) {
```
lxy-9602: For timestamp，please use `literal.GetValue<Timestamp>`.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
lxy-9602: See `RowCompactedSerializer::CompareField` as ref.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -0,0 +1,967 @@
...
+            int cmp_lower = comparator ? comparator(entry->key, lower_bound) : 0;
+
+            // Check lower bound
+            if (!lower_inclusive && cmp_lower == 0) {
+                continue;
+            }
+
+            // Check upper bound
+            int cmp_upper = comparator ? comparator(entry->key, upper_bound) : 0;
+
```
lxy-9602: Consider assume comparator always valid, otherwise return error status.

src/paimon/common/global_index/CMakeLists.txt
```diff
@@ -25,11 +31,13 @@ add_paimon_lib(paimon_global_index
...
+               glog
                fmt
                dl
                Threads::Threads
                SHARED_LINK_LIBS
                paimon_shared
                paimon_file_index_shared
+               paimon_static
                SHARED_LINK_FLAGS
                ${PAIMON_VERSION_SCRIPT_FLAGS})
```
lxy-9602, line 42: Now that we've added `PAIMON_EXPORT`, do we still need `glog` and `paimon_static`? They might be redundant now.

src/paimon/common/sst/sst_file_writer.cpp
ChaomingZhangCN: There's no function named `Write(slice)`.

src/paimon/common/global_index/btree/btree_global_index_writer.cpp
ChaomingZhangCN: [SstFileWriter.java:169
](https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/sst/SstFileWriter.java#L169C5-L180C6)

```java
    @Nullable
    public BloomFilterHandle writeBloomFilter() throws IOException {
        if (bloomFilter == null) {
            return null;
        }
        MemorySegment buffer = bloomFilter.getBuffer();
        BloomFilterHandle bloomFilterHandle =
                new BloomFilterHandle(out.getPos(), buffer.size(), bloomFilter.expectedEntries());
        writeSlice(MemorySlice.wrap(buffer));
        LOG.info("Bloom filter size: {} bytes", bloomFilter.getBuffer().size());
        return bloomFilterHandle;
    }
```

lxy-9602 (approved):
+1

Issue #93, opened 2026-01-29
[Feature] Support Efficient Compaction for Storage Optimization and Query Performance


lxy-9602:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/paimon-cpp/issues) and found nothing similar.

### Motivation

Compaction is essential for maintaining high performance and storage efficiency in modern data systems. Key benefits include:
- **For Append Tables**: Reduces small files by merging existing data files, improving scan performance and metadata scalability.
- **For Primary Key (PK) Tables**: Minimizes the number of segments that need to be merged during read-time (`merge-on-read`), significantly speeding up queries.
- **For PK+DV Tables**: Enables writing DV (`Delete Vector`) files to mark outdated rows, allowing efficient read performance.

Currently, the lack of a dedicated compaction mechanism limits our ability to optimize storage layout and query latency.

### Solution

The compaction framework should support the following capabilities:
1. **Support for both append tables and primary key (PK) tables**, with appropriate strategies for each;
2. **Execution via background tasks or manual triggers**, allowing flexibility in operation;
3. **Built-in basic compaction policies** aligned with Java Paimon;
4. **Generation of Delete Vector (DV) files** during/after compaction to track stale rows;
5. **Design support for data-evolution scenarios**, including both vertical compaction (merging small files) and horizontal compaction (consolidating partial-column files);
6. **Ensure output data format is fully compatible with Java Paimon**.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #216, opened 2026-04-09
refactor(lookup): Decouple RemoteLookupFileManager from LookupLevels and refactor Levels callback lifecycle

lxy-9602:
### Purpose

Linked issue: #93 

Previously, `RemoteLookupFileManager` was a template class (`RemoteLookupFileManager<T>`) that held a raw pointer to `LookupLevels<T>`, creating a tight coupling and circular dependency: `LookupMergeTreeCompactRewriter` owned both `LookupLevels` (via `unique_ptr`) and `RemoteLookupFileManager` (via `unique_ptr`), while `RemoteLookupFileManager` held a raw back-pointer to `LookupLevels`. This made the construction order fragile and introduced use-after-free risks if the destruction order was not carefully managed.

Additionally, `LookupLevels` registered itself as a `DropFileCallback` on `Levels`, but never unregistered — meaning if `LookupLevels` was destroyed before `Levels`, the dangling callback pointer could be invoked, leading to undefined behavior.

1. Remove template parameter and raw pointer from `RemoteLookupFileManager` construction.
2. Add deregistration mechanism for `Levels::DropFileCallback`
3. Inject `RemoteLookupFileManager` into `LookupLevels` at construction
4. Reorder `pool` parameter to last position in compact rewriter hierarchy

### Generative AI tooling
Partially Generated-by: Claude-4.6-Opus

### DIFF
--- a/src/paimon/core/mergetree/compact/merge_tree_compact_manager_factory.cpp
+++ b/src/paimon/core/mergetree/compact/merge_tree_compact_manager_factory.cpp
@@ -165,14 +165,14 @@ Result<std::shared_ptr<CompactRewriter>> MergeTreeCompactManagerFactory::CreateL
     const LookupStrategy& lookup_strategy,
     const std::shared_ptr<FileStorePathFactoryCache>& path_factory_cache,
     const std::shared_ptr<CancellationController>& cancellation_controller) const {
+    PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<RemoteLookupFileManager> remote_lookup_file_manager,
+                           CreateRemoteLookupFileManager(partition, bucket));
     if (lookup_strategy.is_first_row) {
         if (options_.DeletionVectorsEnabled()) {
             return Status::Invalid(
                 "First row merge engine does not need deletion vectors because there is no "
                 "deletion of old data in this merge engine.");
         }
-        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<RemoteLookupFileManager> remote_lookup_file_manager,
-                               CreateRemoteLookupFileManager(partition, bucket));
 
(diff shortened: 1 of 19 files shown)


src/paimon/core/mergetree/levels.cpp
```diff
@@ -96,6 +96,12 @@ void Levels::AddDropFileCallback(DropFileCallback* callback) {
     drop_file_callbacks_.push_back(callback);
 }
 
+void Levels::RemoveDropFileCallback(DropFileCallback* callback) {
+    drop_file_callbacks_.erase(
+        std::remove(drop_file_callbacks_.begin(), drop_file_callbacks_.end(), callback),
+        drop_file_callbacks_.end());
```
lxy-9602, line 102: Add tests.

src/paimon/core/mergetree/compact/merge_tree_compact_manager_factory.cpp
```diff
@@ -302,12 +304,16 @@ MergeTreeCompactManagerFactory::CreateLookupRewriterWithoutDeletionVector(
     const std::shared_ptr<CancellationController>& cancellation_controller) const {
     PAIMON_ASSIGN_OR_RAISE(std::vector<std::string> trimmed_primary_keys,
                            table_schema_->TrimmedPrimaryKeys());
+    PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<RemoteLookupFileManager> remote_lookup_file_manager,
+                           CreateRemoteLookupFileManager(partition, bucket));
+
```
lszskye: only call once in CreateLookupRewriter function

lszskye (approved):

PR #221, opened 2026-04-10
fix(compaction): fix TestHash function in BloomFilter

lszskye:
### Purpose
fix TestHash function in BloomFilter to keep compatibility with Java Paimon


lxy-9602 (dismissed):
+1

lxy-9602 (approved):
+1

PR #223, opened 2026-04-11
feat: support load table by table location directly

### LINKED ISSUE #215: Can we export FileSystemCatalog/SchemaManager
I am integrated Paimon-CPP with my own database, hope to use SchemaManager directly to get table schema.

Smith-Cruise:
### Purpose

Linked issue: close #215

Support load table by path directly.

### DIFF
--- a/include/paimon/catalog/identifier.h
+++ b/include/paimon/catalog/identifier.h
@@ -17,7 +17,6 @@
 #pragma once
 
 #include <string>
-#include <string_view>
 
 #include "paimon/type_fwd.h"
 #include "paimon/visibility.h"
@@ -27,7 +26,7 @@ namespace paimon {
 /// An identifier for a table containing database and table name.
 class PAIMON_EXPORT Identifier {
  public:
-    static constexpr std::string_view kUnknownDatabase = "unknown";
+    static const char kUnknownDatabase[];
 
     explicit Identifier(const std::string& table);
     Identifier(const std::string& database, const std::string& table);
--- a/include/paimon/catalog/identifier.h
+++ b/include/paimon/catalog/identifier.h
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <string>
+#include <string_view>
 
 #include "paimon/type_fwd.h"
 #include "paimon/visibility.h"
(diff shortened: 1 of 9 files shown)


include/paimon/catalog/table.h
```diff
@@ -32,6 +33,10 @@ namespace paimon {
 /// A table provides basic abstraction for table type.
 class PAIMON_EXPORT Table {
  public:
+    static Result<std::shared_ptr<Table>> Create(const std::shared_ptr<FileSystem>& file_system,
+                                                 const std::string& table_path,
+                                                 const Identifier& identifier);
```
lxy-9602, line 44: Thanks for the quick response! 
Could you please add a static constant `kUnknownDatabase = "unknown"` to the `Identifier` class, similar to the Java implementation (when the database is not specified or doesn't exist). Also please add comments to the`Create` method explaining what each parameter represents.

Please also update the `Table::Name()` and `Table::FullName()` methods to align with Java:
 - `Name()` → returns only the table name
 - `FullName()` → returns `"<database>.<table>"`, but if the database is `"unknown"`, it should return just the table name

You can refer to `Identifier.java` and `AbstractFileStoreTable.java` for the expected behavior.

Finally, please add a unit test for `Table::Create` that covers this scenario — especially verifying correct behavior of `Name()`, `FullName()`, and `LatestSchema()` when database is unknown (test data in `test/test_data/` if needed).

Smith-Cruise: done

include/paimon/catalog/identifier.h
```diff
@@ -26,6 +27,9 @@ namespace paimon {
 /// An identifier for a table containing database and table name.
 class PAIMON_EXPORT Identifier {
  public:
+    static constexpr std::string_view kUnknownDatabase = "unknown";
+
```
lxy-9602, line 30: For consistency, could you please use `static const char kUnknownDatabase[]` instead.

include/paimon/catalog/table.h
```diff
@@ -40,12 +51,13 @@ class PAIMON_EXPORT Table {
...
     std::string Name() const {
-        return database_ + "." + table_name_;
+        return table_name_;
     }
 
     /// Full name of the table, default is database.tableName.
     std::string FullName() const {
-        return Name();
+        return database_ == Identifier::kUnknownDatabase ? table_name_
+                                                         : database_ + "." + table_name_;
```
lxy-9602: One final request: could you please move the implementation to `src/table.cpp` to hide details from the header.

Smith-Cruise: done

include/paimon/catalog/identifier.h
Smith-Cruise: done

lucasfang (approved):
+1

lxy-9602 (approved):
+1

PR #234, opened 2026-04-17
feat(scan): add BucketSelectConverter for predicate-based bucket pruning

liangjie3138:
## Purpose

When reading bucketed primary key tables, if the query predicate contains equality conditions on all bucket key fields (for example, a point lookup such as `WHERE pk = 'value'`), the current implementation still scans data files from all buckets.

This PR introduces `BucketSelectConverter`, which automatically extracts literal values for bucket keys from `EQUAL` predicates and computes the target bucket ID using the configured bucket function (`DEFAULT`, `MOD`, or `HIVE`). This allows irrelevant buckets to be pruned during the scan phase, significantly reducing I/O and computation overhead for point lookup scenarios.

### Main Changes

- Add a new `BucketSelectConverter` class that extracts equality constraints on bucket key fields from `AND` predicates, builds a `BinaryRow`, and computes the target bucket using the corresponding `BucketFunction`.
- Integrate the converter into `KeyValueFileStoreScan::SplitAndSetKeyValueFilter()`, so that when a target bucket can be derived, a bucket filter is automatically applied.
- Add a protected method `SetBucketFilterIfAbsent()` in `FileStoreScan` to ensure that an explicitly configured bucket filter is not overridden.

## Tests

- add `BucketSelectConverterTest` 
## API and Format

None. This change does not affect public APIs under the `include` directory and does not introduce any storage format or protocol changes.

## Documentation

None.

## Generative AI Tooling

Generated-by: Claude Code (Claude Opus 4.6)


src/paimon/core/bucket/bucket_select_converter.h
```diff
@@ -0,0 +1,83 @@
...
+    /// @param bucket_function_type The bucket function type (DEFAULT, MOD, HIVE).
+    /// @param num_buckets The total number of buckets.
+    /// @param pool Memory pool for BinaryRow construction.
+    /// @return The target bucket ID, or nullopt if predicates don't fully constrain all bucket
+    /// keys.
+    static Result<std::optional<int32_t>> Convert(
+        const std::shared_ptr<Predicate>& predicate,
+        const std::vector<std::string>& bucket_key_names,
+        const std::vector<FieldType>& bucket_key_types,
+        const std::vector<std::shared_ptr<arrow::DataType>>& bucket_key_arrow_types,
```
lxy-9602, line 60: It looks like `bucket_key_arrow_types` and `bucket_key_types` have some redundancy. 

Consider removing the parameter passing of `bucket_key_types` and instead deriving it from `bucket_key_arrow_types`. This could simplify the interface and reduce maintenance overhead.

src/paimon/core/bucket/bucket_select_converter.h
```diff
@@ -0,0 +1,83 @@
...
+    /// Extract single literal per bucket key field from EQUAL predicates.
+    /// Splits the predicate by AND and looks for EQUAL leaf predicates on bucket key fields.
+    /// @return A map from field name to literal, or nullopt if not all bucket keys are constrained.
+    static std::optional<std::map<std::string, Literal>> ExtractEqualLiterals(
+        const std::shared_ptr<Predicate>& predicate,
+        const std::vector<std::string>& bucket_key_names);
+
+    /// Write a Literal value to a BinaryRowWriter at the given position.
+    static Status WriteLiteralToRow(BinaryRowWriter* writer, int32_t pos, const Literal& literal,
+                                    FieldType field_type,
```
lxy-9602: Could you please place output parameters (e.g., `writer`) at the end of the parameter list?

src/paimon/core/bucket/bucket_select_converter.cpp
```diff
@@ -0,0 +1,205 @@
...
+namespace paimon {
+
+Result<std::optional<int32_t>> BucketSelectConverter::Convert(
+    const std::shared_ptr<Predicate>& predicate, const std::vector<std::string>& bucket_key_names,
+    const std::vector<FieldType>& bucket_key_types,
+    const std::vector<std::shared_ptr<arrow::DataType>>& bucket_key_arrow_types,
+    BucketFunctionType bucket_function_type, int32_t num_buckets, MemoryPool* pool) {
+    if (!predicate || bucket_key_names.empty() || num_buckets <= 0 || !pool) {
+        return std::optional<int32_t>(std::nullopt);
+    }
```
lxy-9602, line 50: Since `BucketSelectConverter::Convert` is an internal implementation detail and not exposed in the public headers, can we assume that `pool` is always valid when called?

If so, would it be acceptable to just assert on it (e.g. `assert(pool)`) rather than handling null explicitly? This would simplify the code under the assumption that callers uphold the precondition.

src/paimon/core/bucket/bucket_select_converter.cpp
```diff
@@ -0,0 +1,205 @@
...
+            writer->WriteFloat(pos, literal.GetValue<float>());
+            break;
+        case FieldType::DOUBLE:
+            writer->WriteDouble(pos, literal.GetValue<double>());
+            break;
+        case FieldType::STRING:
+        case FieldType::BINARY: {
+            std::string value = literal.GetValue<std::string>();
+            writer->WriteStringView(pos, std::string_view(value));
+            break;
```
lxy-9602, line 163: To avoid ambiguity in overload resolution or unintended cast interpretation by the compiler, prefer using `std::string_view{value}` instead of implicit conversions.

src/paimon/core/bucket/bucket_select_converter.cpp
```diff
@@ -0,0 +1,205 @@
...
+Result<std::unique_ptr<BucketFunction>> BucketSelectConverter::CreateBucketFunction(
+    BucketFunctionType type, const std::vector<FieldType>& bucket_key_types) {
+    switch (type) {
+        case BucketFunctionType::DEFAULT:
+            return std::unique_ptr<BucketFunction>(std::make_unique<DefaultBucketFunction>());
+        case BucketFunctionType::MOD: {
+            if (bucket_key_types.size() != 1) {
+                return Status::Invalid("MOD bucket function requires exactly one bucket key field");
+            }
+            PAIMON_ASSIGN_OR_RAISE(auto func, ModBucketFunction::Create(bucket_key_types[0]));
```
lxy-9602: In the `PAIMON_ASSIGN_OR_RAISE` usage, could we use an explicit type instead of `auto`?  

While `auto` works, using an explicit type improves code clarity and makes it easier to understand the expected value type at a glance — especially in macros where type deduction isn't always obvious.

lxy-9602: Also, it is ok to just return `ModBucketFunction::Create(bucket_key_types[0])`.

src/paimon/core/bucket/bucket_select_converter.cpp
```diff
@@ -0,0 +1,205 @@
...
+            return std::unique_ptr<BucketFunction>(std::move(func));
+        }
+        case BucketFunctionType::HIVE: {
+            std::vector<HiveFieldInfo> field_infos;
+            field_infos.reserve(bucket_key_types.size());
+            for (const auto& ft : bucket_key_types) {
+                field_infos.emplace_back(ft);
+            }
```
lxy-9602, line 214: `HiveFieldInfo` needs special handling for `decimal` types, as they require precision (and scale) information to be correctly represented.

Instead of passing basic type info, it's recommended to pass the full `arrow_type` so that `HiveFieldInfo` can extract necessary metadata like precision and scale.

Please refer to the `HiveFieldInfo` class comment for details.

src/paimon/core/bucket/bucket_select_converter_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+    auto predicate = PredicateBuilder::Equal(0, "id", FieldType::INT, lit);
+
+    auto result =
+        BucketSelectConverter::Convert(predicate, {"id"}, {FieldType::INT}, {arrow::int32()},
+                                       BucketFunctionType::DEFAULT, num_buckets, pool_.get());
+    ASSERT_TRUE(result.ok());
+    ASSERT_TRUE(result.value().has_value());
+
+    // Verify by computing the expected bucket manually
+    BinaryRow row(1);
```
lxy-9602: For test cases, consider using `BinaryRowGenerator::GenerateRow` — it provides a simpler and more concise way to generate test rows.

src/paimon/core/bucket/bucket_select_converter_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+TEST_F(BucketSelectConverterTest, SingleIntEqualDefault) {
+    int32_t num_buckets = 10;
+    Literal lit(static_cast<int32_t>(42));
+    auto predicate = PredicateBuilder::Equal(0, "id", FieldType::INT, lit);
+
+    auto result =
+        BucketSelectConverter::Convert(predicate, {"id"}, {FieldType::INT}, {arrow::int32()},
+                                       BucketFunctionType::DEFAULT, num_buckets, pool_.get());
+    ASSERT_TRUE(result.ok());
```
lxy-9602: Consider using `ASSERT_OK_AND_ASSIGN` here instead of manual `.ok` checking. It simplifies the code and makes test logic cleaner.

Also, there are several similar patterns elsewhere in the tests — please refer to `testharness.h` for other available helpers like `ASSERT_OK`, `ASSERT_NOK_WITH_MSG`, etc.

src/paimon/core/bucket/bucket_select_converter_test.cpp
```diff
@@ -0,0 +1,333 @@
...
+    writer_compact.WriteTimestamp(0, ts, /*precision=*/3);
+    writer_compact.Complete();
+
+    BinaryRow row_non_compact(1);
+    BinaryRowWriter writer_non_compact(&row_non_compact, 0, pool_.get());
+    writer_non_compact.WriteTimestamp(0, ts, /*precision=*/6);
+    writer_non_compact.Complete();
+
+    // The two rows have different memory layouts, so hash codes should differ
+    ASSERT_NE(row_compact.HashCode(), row_non_compact.HashCode());
```
lxy-9602: This test seems to only cover `BinaryRowWriter` — is `bucketSelectConverter` actually being tested here? Consider refining or removing if not.

src/paimon/CMakeLists.txt
```diff
@@ -524,6 +525,7 @@ if(PAIMON_BUILD_TESTS)
                     SOURCES
                     core/append/append_only_writer_test.cpp
```
lxy-9602, line 530: Could you please add an end-to-end test in `scan_and_read_inte_test.cpp` to cover this strategy. Test data is available in `test/test_data/`.

lxy-9602:
Great work — thanks for the high-quality submission!  
Looking forward to your updates. Feel free to reach out if anything is unclear.

src/paimon/CMakeLists.txt
lxy-9602: BTW, Java also supports similar append-table logic — is extending support here part of your future plan?  

Would love to see consistency between the two implementations over time.

lxy-9602 (dismissed):
+1

src/paimon/core/bucket/bucket_select_converter.cpp
```diff
@@ -0,0 +1,222 @@
...
+    PAIMON_ASSIGN_OR_RAISE(
+        std::unique_ptr<BucketFunction> bucket_function,
+        CreateBucketFunction(bucket_function_type, bucket_key_types, bucket_key_arrow_types));
+    int32_t bucket = bucket_function->Bucket(row, num_buckets);
+    return std::optional<int32_t>(bucket);
+}
+
+std::optional<std::map<std::string, Literal>> BucketSelectConverter::ExtractEqualLiterals(
+    const std::shared_ptr<Predicate>& predicate, const std::vector<std::string>& bucket_key_names) {
+    std::set<std::string> key_set(bucket_key_names.begin(), bucket_key_names.end());
```
lszskye, line 96: It looks like ExtractEqualLiterals does not yet support **OR** and **IN** predicates. A TODO can be added for this.

lszskye (dismissed):

src/paimon/core/bucket/bucket_select_converter.cpp
lxy-9602: Could you please add a `// TODO` comment based on the review feedback. After that, we can merge this.

lxy-9602 (approved):
+1

PR #236, opened 2026-04-17
chore(log): add logging for scanner and compact manager

ChaomingZhangCN:
- StaticFromSnapshotStartingScanner: log info when no snapshot exists
- DeltaFollowUpScanner: log warning when ignoring non-append snapshot
- BucketedAppendCompactManager: log warning on file overlap detection

### Purpose

Linked issue: close #xxx

### DIFF
--- a/src/paimon/core/append/bucketed_append_compact_manager.h
+++ b/src/paimon/core/append/bucketed_append_compact_manager.h
@@ -61,7 +61,13 @@ class BucketedAppendCompactManager : public CompactFutureManager {
                 return false;
             }
             if (!ignore_overlap && IsOverlap(lhs, rhs)) {
-                // TODO(yonghao.fyh): add log
+                static auto logger = Logger::GetLogger("BucketedAppendCompactManager");
+                PAIMON_LOG_WARN(logger,
+                                "Overlap detected between files %s and %s, "
+                                "sequence range [%ld, %ld] and [%ld, %ld].",
+                                lhs->file_name.c_str(), rhs->file_name.c_str(),
+                                lhs->min_sequence_number, lhs->max_sequence_number,
+                                rhs->min_sequence_number, rhs->max_sequence_number);
             }
             return lhs->min_sequence_number < rhs->min_sequence_number;
         };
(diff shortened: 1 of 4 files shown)


src/paimon/core/table/source/snapshot/delta_follow_up_scanner.h
```diff
@@ -27,7 +28,10 @@ class DeltaFollowUpScanner : public FollowUpScanner {
         if (snapshot.GetCommitKind() == Snapshot::CommitKind::Append()) {
             return true;
         }
-        // TODO(liancheng.lsz): log
+        static auto logger = Logger::GetLogger("DeltaFollowUpScanner");
+        PAIMON_LOG_WARN(
+            logger, "Ignore snapshot #%ld with commit kind %s in delta follow-up scanner.",
+            snapshot.Id(), Snapshot::CommitKind::ToString(snapshot.GetCommitKind()).c_str());
         return false;
```
lxy-9602, line 36: Java logs this as `DEBUG` — is `WARN` necessary here? If it's not an actual issue, `DEBUG/INFO` might be more accurate.

ChaomingZhangCN: make sense.

src/paimon/core/append/bucketed_append_compact_manager.h
```diff
@@ -61,7 +61,13 @@ class BucketedAppendCompactManager : public CompactFutureManager {
                 return false;
             }
             if (!ignore_overlap && IsOverlap(lhs, rhs)) {
-                // TODO(yonghao.fyh): add log
+                static auto logger = Logger::GetLogger("BucketedAppendCompactManager");
```
lucasfang: Consider making the logger a private member variable of the class rather than a static local variable？

ChaomingZhangCN: Since current method is static, we can not make the logger a private member variable of the class.

lucasfang: Okay, I missed that—thanks for pointing it out. For the other two methods that are not static, update the logger to be a private member variable?

ChaomingZhangCN: Already done.

lxy-9602 (approved):
+1

PR #242, opened 2026-04-21
chore(ut): make btree index ut suitable for PAIMON_TEST_DATA

SGZW:
### Purpose
make btree index ut suitable for PAIMON_TEST_DATA env variables

Linked issue: close #xxx


lxy-9602:
Thank you for the patch! I'm already working on a B-tree refactor and will fix this accordingly using `GetDataDir` instead of `getcwd`. Will let you know when it's pushed.

PR #246, opened 2026-04-24
feat(scan): support timestamp-based snapshot lookup

mrdrivingduck:
## Summary

- Add `FROM_TIMESTAMP` startup mode support for both batch and streaming scan, allowing users to specify a target timestamp to locate the corresponding snapshot
- Two options: `scan.timestamp-millis` (unix milliseconds) and `scan.timestamp` (human-readable datetime string with local-timezone conversion), aligning with Java Paimon's time-travel capability
- Snapshot lookup uses binary search over snapshot IDs, with less-equal semantics for batch and strictly-less semantics for streaming, matching Java Paimon's behavior

## Test plan

- [ ] Verify batch scan resolves to the correct snapshot at or before the given timestamp
- [ ] Verify streaming scan starts consuming deltas strictly after the timestamp boundary
- [ ] Verify error is returned when no snapshot exists at or before the given timestamp (batch)
- [ ] Verify `scan.timestamp` string parsing for formats: `yyyy-MM-dd`, `yyyy-MM-dd HH:mm:ss`, `yyyy-MM-dd HH:mm:ss.SSS`
- [ ] Verify trailing garbage in timestamp string is rejected


src/paimon/core/table/source/abstract_table_scan.h
```diff
@@ -83,6 +83,33 @@ class AbstractTableScan : public TableScan {
...
+                    "FROM_TIMESTAMP");
+            }
+            if (is_streaming) {
+                PAIMON_ASSIGN_OR_RAISE(
+                    std::optional<Snapshot> earlier_snapshot,
+                    snapshot_manager->EarlierThanTimeMillis(timestamp_millis.value()));
+                int64_t start_id =
+                    earlier_snapshot ? earlier_snapshot->Id() + 1 : Snapshot::FIRST_SNAPSHOT_ID;
+                return std::make_shared<ContinuousFromSnapshotStartingScanner>(snapshot_manager,
+                                                                               start_id);
```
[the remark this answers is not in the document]
mrdrivingduck: This is actually the intended behavior, consistent with Java Paimon.

For streaming FROM_TIMESTAMP, the goal is to start reading **from the first snapshot at or after the timestamp**. The logic is:

1. `EarlierThanTimeMillis(ts)` finds the last snapshot with `timeMillis < ts` (strictly less than).
2. Starting from `earlier_snapshot->Id() + 1` means we begin at the first snapshot whose `timeMillis >= ts`.

When `ts` exactly equals a snapshot's timeMillis:
- `EarlierThanTimeMillis` returns the snapshot **before** it (or nullopt if it's the first).
- `id + 1` then points to exactly the snapshot at the timestamp — which **should** be included for streaming.

This matches [Java Paimon's `AbstractDataTableScan`](https://github.com/apache/paimon/blob/master/paimon-core/src/main/java/org/apache/paimon/table/source/AbstractDataTableScan.java) where streaming FROM_TIMESTAMP includes the snapshot at the boundary timestamp.

src/paimon/core/utils/snapshot_manager.cpp
```diff
@@ -248,4 +248,49 @@ Result<std::vector<Snapshot>> SnapshotManager::GetAllSnapshots() const {
...
+        PAIMON_ASSIGN_OR_RAISE(Snapshot snapshot, LoadSnapshot(mid));
+        if (compare(snapshot.TimeMillis(), timestamp_millis)) {
+            lo = mid + 1;
+            result = std::move(snapshot);
+        } else {
+            hi = mid - 1;
+        }
+    }
+
+    return result;
```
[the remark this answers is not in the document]
mrdrivingduck: I think it is by design of Paimon: snapshot expiration always removes from the earliest end, keeping the range contiguous.

src/paimon/core/utils/snapshot_manager.cpp
```diff
@@ -248,4 +248,49 @@ Result<std::vector<Snapshot>> SnapshotManager::GetAllSnapshots() const {
...
+    return FindSnapshotBeforeTimestamp(timestamp_millis, std::less_equal<int64_t>{});
+}
+
+Result<std::optional<Snapshot>> SnapshotManager::EarlierThanTimeMillis(
+    int64_t timestamp_millis) const {
+    return FindSnapshotBeforeTimestamp(timestamp_millis, std::less<int64_t>{});
+}
+
+Result<std::optional<Snapshot>> SnapshotManager::FindSnapshotBeforeTimestamp(
+    int64_t timestamp_millis, const std::function<bool(int64_t, int64_t)>& compare) const {
```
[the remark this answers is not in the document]
mrdrivingduck: Not worth the tradeoff, compared with the I/O cost.

src/paimon/common/utils/string_utils.cpp
```diff
@@ -148,4 +148,70 @@ Result<int32_t> StringUtils::StringToDate(const std::string& str) {
...
+        return Status::Invalid(
+            fmt::format("failed to convert string '{}' to timestamp, "
+                        "unexpected trailing characters",
+                        str));
+    }
+
+    std::time_t time = mktime(&timeinfo);
+    if (time == -1) {
+        return Status::Invalid(fmt::format("failed to convert string '{}' to timestamp", str));
+    }
```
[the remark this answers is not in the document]
mrdrivingduck: Keep consist with Java impl, not involving extra strictness.

mrdrivingduck:
@lxy-9602 @lucasfang hi, this PR is ready for review. Support time-based snapshot selection which will be used for time travel. The implementation is almost the same as Java Paimon.

src/paimon/common/utils/string_utils_test.cpp
```diff
@@ -401,6 +401,54 @@ TEST_F(StringUtilsTest, TestStringToDate) {
...
 
+TEST_F(StringUtilsTest, TestStringToTimestampMillis) {
+    // "yyyy-MM-dd HH:mm:ss" format
+    {
+        ASSERT_OK_AND_ASSIGN(auto millis,
+                             StringUtils::StringToTimestampMillis("1970-01-01 00:00:00"));
+        // Local timezone: mktime converts local time. In UTC+8 this would be negative.
+        // Just verify it returns a valid result without checking exact value (timezone-dependent).
+        (void)millis;
+    }
```
lxy-9602, line 412: Consider using `paimon::test::TimezoneGuard` to stabilize the test timezone. Also, prefer asserting on the `millis` value directly for more reliable results.

src/paimon/core/core_options_test.cpp
```diff
@@ -598,4 +598,50 @@ TEST(CoreOptionsTest, TestNormalizeValueInCoreOption) {
...
+                         CoreOptions::FromMap({{Options::SCAN_TIMESTAMP, "2023-06-01 00:00:00"}}));
+    ASSERT_EQ(opts1.GetScanTimestampMillis().value(), opts2.GetScanTimestampMillis().value());
+}
+
+TEST(CoreOptionsTest, TestScanTimestampMillisAndStringMutuallyExclusive) {
+    ASSERT_NOK_WITH_MSG(CoreOptions::FromMap({{Options::SCAN_TIMESTAMP_MILLIS, "1721614515032"},
+                                              {Options::SCAN_TIMESTAMP, "2023-06-01 00:00:00"}}),
+                        "scan.timestamp-millis and scan.timestamp cannot be set at the same time");
+}
+
```
lxy-9602, line 645: Thanks for the careful timezone handling! Could you add integration tests in `scan_and_read_inte_test.cpp` to verify the full config path end-to-end (for batch and stream scan)?

src/paimon/common/utils/string_utils_test.cpp
mrdrivingduck: Done.

src/paimon/core/core_options_test.cpp
mrdrivingduck: Done, add test cases which passes time information into scanner, and compare the snapshot ID of the plan with the expected snapshot ID.

test/inte/scan_and_read_inte_test.cpp
```diff
@@ -468,6 +471,111 @@ TEST_P(ScanAndReadInteTest, TestWithAppendSnapshotWithStreamWithDefaultMode) {
...
+        ASSERT_EQ(plan->SnapshotId().value(), 5);
+    }
+    // EarlierOrEqual(0) → no snapshot found, expect error
+    {
+        ScanContextBuilder builder(table_path);
+        builder.AddOption(Options::SCAN_TIMESTAMP_MILLIS, "0");
+        ASSERT_OK_AND_ASSIGN(std::unique_ptr<ScanContext> ctx, builder.Finish());
+        ASSERT_OK_AND_ASSIGN(std::unique_ptr<TableScan> scan, TableScan::Create(std::move(ctx)));
+        ASSERT_NOK(scan->CreatePlan());
+    }
```
lxy-9602: Great and comprehensive test case — thanks for adding this coverage!

It looks like the main focus is on validating the snapshot IDs generated by the scan. Given that, do you think this test might be a better fit for `scan_inte_test.cpp`?

Also, for clarity, could we use `ASSERT_NOK_WITH_MSG` instead of `ASSERT_NOK`? Adding a custom error message would make failures easier to debug.

mrdrivingduck: Yeah, will do.

mrdrivingduck: > "It looks like the main focus is on validating the snapshot IDs generated by the scan."

I originally planned to test the full scan-and-read path end-to-end, but realized that once the snapshot ID is resolved from the timestamp, the downstream read path is entirely reused from existing code that's already covered by tests. So these tests are really just about snapshot selection, and `scan_inte_test.cpp` is a better home.

mrdrivingduck: done

lxy-9602 (approved):
+1

PR #248, opened 2026-04-24
fix: Fix missing StatusCode switch cases and correct Literal hash qualifier

Smith-Cruise:
### Purpose

- Add missing StatusCode::NotExist and StatusCode::Exist cases in Status::CodeAsString(StatusCode), so these error codes produce proper string representations instead of falling through to "Unknown".
- Fix `std::hash<paimon::Literal>` specialization to use `::paimon::Literal` (global namespace qualifier) instead of `paimon::Literal`, avoiding potential ambiguity in nested namespace contexts.

Linked issue: close #xxx

### DIFF
--- a/include/paimon/predicate/literal.h
+++ b/include/paimon/predicate/literal.h
@@ -106,7 +106,7 @@ class PAIMON_EXPORT Literal {
 
 namespace std {
 template <>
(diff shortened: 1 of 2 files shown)


lxy-9602 (approved):
+1

PR #247, opened 2026-04-24
fix(global_index): compatibility for legacy lumina index type

lszskye:
### Purpose
This PR adds a compatibility mapping in GlobalIndexerFactory::Get: when the identifier is "lumina-vector-ann", it is transparently rewritten to "lumina" before looking up the factory.

### Generative AI tooling
No

PR #249, opened 2026-04-24
fix(global_index): compatibility for legacy lumina index type (#247)

lszskye:
### Purpose
This PR:
(1) Adds a compatibility mapping in GlobalIndexerFactory::Get: when the identifier is "lumina-vector-ann", it is transparently rewritten to "lumina" before looking up the factory.
(2) update lumina lib to v0.2.1


lxy-9602 (dismissed):
+1

lxy-9602 (approved):
+1

PR #252, opened 2026-04-25
refactor(btree): eliminate hot-path overhead from allocs and deserialization

lxy-9602:
### Purpose

Linked issue: #38 

Optimize BTree global index read performance by reducing memory allocations and eliminating unnecessary deserialization in the hot path.

### Changes

- **MemorySegment**: add non-owning `WrapView` mode with cached `data_`/`size_`, avoiding `shared_ptr` dereference on every access.
- **MemorySlice**: `ReadSlice` now returns a zero-copy non-owning view.
- **VarLengthIntUtils**: unify varint encode/decode into `static inline` functions with 1-byte fast path for values 0–127.
- **MemorySliceInput**: inline all hot-path methods into header; delete `memory_slice_input.cpp`.
- **BTreeGlobalIndexReader**: compare keys at byte level via `SliceComparator` instead of deserializing to `Literal` per entry.
- **KeySerializer**: add fast-path comparator for integer and string.
- **BlockIterator**: return `BlockEntry` by value instead of `unique_ptr`.

### Tests
MemorySegmentTest.TestNonOwningWrapView

### Generative AI tooling
Partially Generated-by: Claude-4.6-Opus


src/paimon/common/utils/var_length_int_utils.h
```diff
@@ -34,67 +41,85 @@ class VarLengthIntUtils {
...
+        return Status::Invalid("Malformed varint32: too many continuation bytes");
+    }
 
-            if ((b & 0x80u) == 0) {
+    /// Decodes a varint64 from `data` at `*offset`, advancing `*offset` past the consumed bytes.
+    /// Inlines a 1-byte fast path (values 0-127), which is the most common case.
+    static inline Result<int64_t> DecodeLong(const char* data, int32_t* offset) {
+        auto first_byte = static_cast<uint8_t>(data[*offset]);
+        if (__builtin_expect((first_byte & 0x80) == 0, 1)) {
+            ++(*offset);
```
lxy-9602, line 108: PAIMON_LIKELY

lszskye (approved):

PR #256, opened 2026-04-28
fix: Use fully qualified namespace for Literal in std::hash specialization

Smith-Cruise:
### Purpose

In #248 I have already fixed part of it, but I forgot to fix the function's parameters.


lxy-9602 (commented):
+1

lxy-9602 (approved):
+1

PR #258, opened 2026-04-29
feat(parquet): add metrics for parquet reader observability

duanyyyyyyy:
Add row groups, rows, batch count, and latency metrics to ParquetFileBatchReader, matching the observability level of the ORC reader.

### Purpose

Linked issue: close #xxx

### Generative AI tooling
Opus 4.6


src/paimon/format/parquet/parquet_format_defs.h
```diff
@@ -61,6 +61,13 @@ static constexpr uint32_t DEFAULT_PARQUET_READ_PREDICATE_NODE_COUNT_LIMIT = 512;
...
  public:
     static inline const char WRITE_RECORD_COUNT[] = "parquet.write.record.count";
+
+    // read
+    static inline const char READ_ROW_GROUPS_TOTAL[] = "parquet.read.row_groups.total";
+    static inline const char READ_ROW_GROUPS_FILTERED[] = "parquet.read.row_groups.filtered";
+    static inline const char READ_ROWS[] = "parquet.read.rows";
+    static inline const char READ_BATCH_COUNT[] = "parquet.read.batch.count";
+    static inline const char READ_NEXT_BATCH_LATENCY_MS[] = "parquet.read.next_batch.latency.ms";
 };
```
lxy-9602, line 70: Suggestion: for high-level categories, consider using `.` as the separator to represent hierarchy (e.g., `parquet.read`), while using `-` to separate words within the same level (e.g., `batch-count`).

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -243,6 +248,14 @@ Result<BatchReader::ReadBatch> ParquetFileBatchReader::NextBatch() {
     std::unique_ptr<ArrowArray> c_array = std::make_unique<ArrowArray>();
     std::unique_ptr<ArrowSchema> c_schema = std::make_unique<ArrowSchema>();
     PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::ExportArray(*array, c_array.get(), c_schema.get()));
+
+    read_rows_ += array->length();
+    read_batch_count_++;
+    read_next_batch_latency_ms_ += timer.Get();
+    metrics_->SetCounter(ParquetMetrics::READ_ROWS, read_rows_);
+    metrics_->SetCounter(ParquetMetrics::READ_BATCH_COUNT, read_batch_count_);
+    metrics_->SetCounter(ParquetMetrics::READ_NEXT_BATCH_LATENCY_MS, read_next_batch_latency_ms_);
```
lucasfang: Using a counter for READ_NEXT_BATCH_LATENCY_MS seems inappropriate. Since the prefetch reader aggregates metrics across all sub-readers, the resulting value is not very meaningful for a latency metric. If the goal is to capture end-to-end latency, I think it would be more appropriate to record it as a histogram at the framework level, rather than aggregating sub-reader metrics.

lucasfang: This bug seems also exist in the orc format, we will fix this in the future

duanyyyyyyy: As you will also fix the ORC next batch issue
I will remove it.

lxy-9602 (approved):
+1

lucasfang (approved):
+1

Issue #141, opened 2026-02-12
Support querying system tables in paimon-cpp


suxiaogang223:
According to the official Paimon docs, system tables are first-class query interfaces for table metadata, changelog analysis, and catalog-wide observability:
https://paimon.apache.org/docs/master/concepts/system-tables/

`paimon-cpp` currently lacks system table query support.
This request is to add system table query support in `paimon-cpp`, including:

- Data system tables (for example):
  `snapshots`, `schemas`, `options`, `audit_log`, `binlog`, `ro`, `files`, `tags`, `branches`, `consumers`, `manifests`, `aggregation_fields`, `partitions`, `buckets`, `statistics`, `table_indexes`
- Global system tables under `sys` (for example):
  `tables`, `partitions`, `all_table_options`, `catalog_options`

Expected outcome:
- Users can query system tables in C++ environments with behavior aligned to Paimon documentation.
- Core changelog observability use cases (including `audit_log` and `binlog`) are covered.
- Metadata and global catalog inspection via system tables are available in `paimon-cpp`.

lxy-9602:
Very useful issue — thank you!

The system table part overlaps with catalog logic, so we can wait until the catalog refactor (#84 , #92 ) is complete before moving forward. Changelog support isn't available in paimon-cpp yet, but this would be a great addition.

Would you be open to contributing this feature? We’re happy to help with design and review.

suxiaogang223:
I’d like to start with a conservative phased implementation, keeping the first PR small and leaving room for design adjustments.

TODO:

- [x] Add a lightweight system table routing layer for read-only system tables.
- [x] First support metadata-oriented system tables that can be built from existing catalog, schema, snapshot, tag, branch, and manifest metadata.
- [x] Add `audit_log` and `binlog` after the changelog read path is clarified and supported.
- [ ] Add global `sys` tables after the catalog interface, especially REST catalog support, is stable enough.
- [ ] Keep system tables read-only and align schemas/behavior with Apache Paimon documentation as much as possible.

This should let us deliver useful metadata observability first, while avoiding over-specifying details that may change during implementation.

PR #261, opened 2026-04-30
feat: add system table framework with options table

suxiaogang223:
Related to #141.

## Summary

- Add Java-aligned system table identifier parsing for `table$system_table` and reserved branch form `table$branch_xxx$system_table`
- Add a lightweight system table framework with loader, scan, read, schema, and in-memory split support
- Implement the first metadata system table: `table$options`
- Route catalog schema/existence and scan/read paths through the system table framework
- Keep system table DDL unsupported

## Tests

- Added identifier parsing tests for normal tables, system tables, branch tables, and invalid `$` combinations
- Added catalog tests for `table$options` existence/schema and system table DDL rejection
- Added scan/read integration coverage for reading `table$options`
- Verified on Fedora:
  - `cmake --build build-codex-system-table-pr1 --target paimon-core-test -j$(nproc)`
  - `./build-codex-system-table-pr1/release/paimon-core-test --gtest_filter="IdentifierTest.*:FileSystemCatalogTest.TestOptionsSystemTableCatalog:TableReadTest.TestReadOptionsSystemTable"`
- Ran formatting checks:
  - `clang-format`
  - `cmake-format`
  - pre-commit `clang-format`
  - pre-commit `cmake-format`


suxiaogang223:
@lxy-9602 Hi, could you please help review this PR when you have time? This is the first step for system table support, adding the shared framework and the minimal `table$options` implementation. I’d like to continue with the follow-up system table work after this direction is reviewed. Thanks!🤓

src/paimon/core/catalog/identifier.cpp
```diff
@@ -40,8 +49,71 @@ const std::string& Identifier::GetTableName() const {
...
+        data_table = splits[0];
+        if (StringUtils::StartsWith(splits[1], kSystemBranchPrefix, /*start_pos=*/0)) {
+            branch = splits[1].substr(std::strlen(kSystemBranchPrefix));
+        } else {
+            system_table = splits[1];
+        }
+    } else if (splits.size() == 3) {
+        if (!StringUtils::StartsWith(splits[1], kSystemBranchPrefix, /*start_pos=*/0)) {
+            throw std::invalid_argument(fmt::format(
+                "System table can only contain one '$' separator, but this is: {}", table_));
```
lxy-9602, line 103: In paimon-cpp, we return `Status` or `Result<T>` instead of throwing exceptions, to avoid potential issues caused by exceptions. Could you please refactor this to use status checking instead of exception handling?

src/paimon/core/catalog/identifier_test.cpp
```diff
@@ -52,4 +54,60 @@ TEST(IdentifierTest, EmptyDatabaseRemainsEmpty) {
     EXPECT_EQ(id.GetTableName(), "my_table");
 }
 
+TEST(IdentifierTest, ParseDataTable) {
+    Identifier id("db", "tbl");
+    EXPECT_EQ(id.GetTableName(), "tbl");
+    EXPECT_EQ(id.GetDataTableName(), "tbl");
```
lxy-9602: Could you use `ASSERT_*` instead of `EXPECT_*` here? We prefer `ASSERT` in paimon-cpp to fail early and avoid potential undefined behavior after failures.

src/paimon/core/catalog/identifier_test.cpp
```diff
@@ -52,4 +54,60 @@ TEST(IdentifierTest, EmptyDatabaseRemainsEmpty) {
...
+    ASSERT_TRUE(id.GetBranchName());
+    EXPECT_EQ(id.GetBranchName().value(), "dev");
+    ASSERT_TRUE(id.GetSystemTableName());
+    EXPECT_EQ(id.GetSystemTableName().value(), "options");
+    EXPECT_TRUE(id.IsSystemTable());
+}
+
+TEST(IdentifierTest, InvalidSystemTableName) {
+    Identifier invalid_middle("db", "tbl$bad$options");
+    EXPECT_THROW(invalid_middle.IsSystemTable(), std::invalid_argument);
```
lxy-9602: When testing error paths with `Status` or `Result`, you might want to use `ASSERT_NOK_WITH_MSG` to check the failure and provide a descriptive error message. It improves clarity when tests fail.

See `testharness.h` for more details on the available assertion macros.

src/paimon/core/table/source/split.cpp
```diff
@@ -217,6 +223,21 @@ Result<std::shared_ptr<Split>> Split::Deserialize(const char* buffer, size_t len
...
+        }
+        PAIMON_ASSIGN_OR_RAISE(std::string table_path, in.ReadString());
+        PAIMON_ASSIGN_OR_RAISE(int64_t pos, in.GetPos());
+        PAIMON_ASSIGN_OR_RAISE(int64_t stream_length, in.Length());
+        if (pos != stream_length) {
+            return Status::Invalid(
+                fmt::format("invalid SystemTableSplit, remaining {} bytes after deserializing",
+                            stream_length - pos));
+        }
+        return std::make_shared<SystemTableSplit>(table_path);
```
lxy-9602: `DataSplit` and `IndexedSplit` have cross-language serialization compatibility guarantees with the Java side.

However, I'm not sure if I missed something — I couldn't find the corresponding serialization/deserialization logic for `SystemTableSplit` in the Java codebase.

Before we support serializing/deserializing `SystemTableSplit` in C++, could we confirm that the Java side already has compatible support for it? Without a shared, well-defined protocol, there's a risk of breaking cross-language compatibility.

If Java doesn’t yet support it, perhaps we should treat `SystemTableSplit` as an in-memory-only type for now — meaning it won't be serializable or deserializable until both sides are aligned.

src/paimon/core/table/system/system_table.h
```diff
@@ -0,0 +1,66 @@
...
+    const std::string& system_table_name, const std::shared_ptr<FileSystem>& fs,
+    const std::string& table_path, const std::shared_ptr<TableSchema>& table_schema);
+
+bool IsSupportedSystemTable(const std::string& system_table_name);
+
+Result<std::optional<SystemTablePath>> TryParseSystemTablePath(const std::string& path);
+
+Result<std::shared_ptr<SystemTable>> LoadSystemTableFromPath(const std::shared_ptr<FileSystem>& fs,
+                                                             const std::string& path);
+
```
lxy-9602, line 70: To improve organization and maintainability, we generally avoid free functions in the paimon namespace.
Could you consider moving these functions into a class as a static method, or into a utility class?

src/paimon/core/table/source/table_read_test.cpp
```diff
@@ -153,4 +164,62 @@ TEST(TableReadTest, TestMergeOptions) {
...
+    ASSERT_OK_AND_ASSIGN(auto table_read, TableRead::Create(std::move(read_context)));
+    std::vector<std::shared_ptr<Split>> splits = {deserialized_split};
+    ASSERT_OK_AND_ASSIGN(auto batch_reader, table_read->CreateReader(splits));
+    ASSERT_NOK_WITH_MSG(table_read->CreateReader(std::vector<std::shared_ptr<Split>>{}),
+                        "single split");
+    ASSERT_NOK_WITH_MSG(table_read->CreateReader(std::make_shared<Split>()), "unsupported split");
+    ASSERT_OK_AND_ASSIGN(auto result, ReadResultCollector::CollectResult(batch_reader.get()));
+    ASSERT_TRUE(result);
+    ASSERT_EQ(result->type()->id(), arrow::Type::STRUCT);
+
```
lxy-9602: Consider using `arrow::ipc::internal::json::ChunkedArrayFromJSON` combined with `arrow::Array::Equals` for this test.
It allows you to verify both data content and type correctness in one go, and makes the test more concise and easier to read.

lxy-9602: Since this test covers scanning and reading system tables, would it be better placed into `test/inte/`.  
This helps treat it as an integration scenario and paves the way for organizing future system table tests in one dedicated place.

src/paimon/core/table/system/system_table.h
```diff
@@ -0,0 +1,66 @@
...
+ public:
+    virtual ~SystemTable() = default;
+
+    virtual std::string Name() const = 0;
+    virtual std::shared_ptr<arrow::Schema> ArrowSchema() const = 0;
+    virtual Result<std::unique_ptr<TableScan>> NewScan() const = 0;
+    virtual Result<std::unique_ptr<BatchReader>> NewReader(
+        const std::vector<std::shared_ptr<Split>>& splits,
+        const std::shared_ptr<MemoryPool>& pool) const = 0;
+};
```
lxy-9602, line 55: Since `TableScan` and `TableRead` are now better aligned, would it make sense to refactor `NewReader()` to `NewRead()` and return `SystemTableRead`?

src/paimon/core/table/system/system_table.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+Result<std::shared_ptr<SystemTable>> CreateSystemTable(
+    const std::string& system_table_name, const std::shared_ptr<FileSystem>& /*fs*/,
+    const std::string& table_path, const std::shared_ptr<TableSchema>& table_schema) {
+    std::string normalized_name = StringUtils::ToLowerCase(system_table_name);
+    if (normalized_name == OptionsSystemTable::NAME) {
+        return std::make_shared<OptionsSystemTable>(table_path, table_schema);
+    }
+    return Status::NotExist("unsupported system table: ", system_table_name);
```
lxy-9602: May return `Status::NotImplement` for other system tables.

src/paimon/core/table/system/system_table.cpp
```diff
@@ -0,0 +1,83 @@
...
+        }
+        std::string parent = PathUtil::GetParentDirPath(path);
+        SystemTablePath system_table_path;
+        system_table_path.table_path = PathUtil::JoinPath(parent, identifier.GetDataTableName());
+        system_table_path.branch = identifier.GetBranchName();
+        system_table_path.system_table_name = identifier.GetSystemTableName().value();
+        return std::optional<SystemTablePath>(std::move(system_table_path));
+    } catch (const std::exception& e) {
+        return Status::Invalid(e.what());
+    }
```
lxy-9602, line 61: The `try-catch` can be removed — when `Identifier` returns errors via status rather than exceptions.

src/paimon/core/table/source/table_read_test.cpp
```diff
@@ -153,4 +164,62 @@ TEST(TableReadTest, TestMergeOptions) {
...
+    auto struct_array = std::static_pointer_cast<arrow::StructArray>(result->chunk(0));
+    auto key_array = std::static_pointer_cast<arrow::StringArray>(struct_array->field(0));
```
lxy-9602: One more thing — could we add a test case for `OptionsSystemTable` that includes a branch?
This would help verify correct behavior when reading options from non-default branches.

You can use existing test databases under `test/test_data/`, such as `append_table_with_rt_branch`, which already contains multiple branches. That should make it easy to set up the scenario.

src/paimon/core/table/system/system_table.cpp
```diff
@@ -0,0 +1,83 @@
...
+    PAIMON_ASSIGN_OR_RAISE(std::optional<SystemTablePath> system_table_path,
+                           TryParseSystemTablePath(path));
+    if (!system_table_path) {
+        return Status::Invalid("path is not a system table path: ", path);
+    }
+    const auto& parsed = system_table_path.value();
+    SchemaManager schema_manager(fs, parsed.table_path, parsed.branch.value_or(""));
+    PAIMON_ASSIGN_OR_RAISE(std::optional<std::shared_ptr<TableSchema>> latest_schema,
+                           schema_manager.Latest());
+    if (!latest_schema) {
```
lxy-9602, line 85: Just a question: should the default branch be set to `BranchManager::DEFAULT_MAIN_BRANCH` instead of `""` (empty string)?

src/paimon/core/table/system/system_table_scan.h
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+ private:
+    std::string table_path_;
+};
+
+class SystemTableScan : public TableScan {
+ public:
+    explicit SystemTableScan(std::string table_path);
+
```
lxy-9602, line 44: Could we use `const std::string& table_path` for param? It matches the parameter style used elsewhere.

src/paimon/core/table/system/system_table_schema.h
```diff
@@ -0,0 +1,55 @@
...
+    std::vector<std::string> FieldNames() const override;
+    Result<FieldType> GetFieldType(const std::string& field_name) const override;
+    int64_t Id() const override;
+    const std::vector<std::string>& PrimaryKeys() const override;
+    const std::vector<std::string>& PartitionKeys() const override;
+    const std::vector<std::string>& BucketKeys() const override;
+    int32_t NumBuckets() const override;
+    int32_t HighestFieldId() const override;
+    const std::map<std::string, std::string>& Options() const override;
+    std::optional<std::string> Comment() const override;
```
lxy-9602, line 39: Functions like `Id()` and `PrimaryKeys()` don't seem to apply well to system tables.

We might want to consider splitting the schema into a base class with separate e.g., `DataSchema` and `SystemSchema` types — this would let each expose only the relevant APIs.

This refactoring could be done together with future system table work, so we don’t grow the scope of this PR too much.

src/paimon/core/table/system/options_system_table.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include "paimon/core/table/system/system_table.h"
+
+namespace paimon {
+class TableSchema;
+
+class OptionsSystemTable : public SystemTable {
+ public:
+    static constexpr const char* NAME = "options";
+
```
lxy-9602, line 32: Consider using `kName` for static constants in the new code.

src/paimon/core/table/system/options_system_table.cpp
```diff
@@ -0,0 +1,120 @@
...
+    Result<ReadBatch> NextBatch() override {
+        if (emitted_) {
+            return BatchReader::MakeEofBatch();
+        }
+        emitted_ = true;
+
+        arrow::StringBuilder key_builder;
+        arrow::StringBuilder value_builder;
+        for (const auto& [key, value] : options_) {
+            PAIMON_RETURN_NOT_OK_FROM_ARROW(key_builder.Append(key));
```
lxy-9602: Consider using `arrow::MakeBuilder` with `arrow_pool` to construct the `ArrayBuilder`.  
This guarantees that the generated array data is allocated from the pool, as expected for consistent memory management.

lxy-9602:
Thanks for this great contribution — the code is clean, well-structured, and clearly reflects the effort you've made to bridge C++ and Java differences.  

Looking forward to your follow-up updates and more system table support!

src/paimon/core/table/system/system_table_read.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace paimon {
+
+SystemTableRead::SystemTableRead(std::shared_ptr<SystemTable> system_table,
+                                 const std::shared_ptr<MemoryPool>& memory_pool)
+    : TableRead(memory_pool), system_table_(std::move(system_table)) {}
+
+Result<std::unique_ptr<BatchReader>> SystemTableRead::CreateReader(
+    const std::vector<std::shared_ptr<Split>>& splits) {
+    return system_table_->CreateBatchReader(splits, GetMemoryPool());
```
lszskye, line 33: return system_table_->CreateBatchReader(splits, pool_);

suxiaogang223: Thanks for the suggestion. Since `pool_` is private in `TableRead`, `SystemTableRead` cannot access it directly. I therefore used `GetMemoryPool()` here, which still passes the same pool to `CreateBatchReader()` and keeps the encapsulation of `TableRead` intact. This felt like a slightly better fit for the current structure.`

lxy-9602 (approved):
+1

PR #268, opened 2026-05-09
feat: add audit_log and binlog system tables

suxiaogang223:
## Summary
- Add `table$audit_log` and `table$binlog` on top of the system table framework.
- Route catalog, scan, and read paths through context-aware `SystemTable` interfaces.
- Output `rowkind`, optional `_SEQUENCE_NUMBER`, and batch-mode `binlog` singleton-list columns.
- Reuse the primary-key table read path with delete records preserved; non-primary-key tables return a clear unsupported status.
- Remove the legacy `SystemTableRead` fallback so system tables use the same scan/read context path.

Part of #141.

## Scope
- This PR focuses on batch read correctness for `audit_log` and `binlog`.
- Streaming before/after pairing and full predicate pushdown are intentionally left for follow-up work.

## Validation
- Fedora: `cmake --build build-codex-pr2 --target paimon-core-test paimon-read-inte-test -j4`
- Fedora: `./build-codex-pr2/release/paimon-core-test --gtest_filter="FileSystemCatalogTest.TestAuditLogAndBinlogSystemTableCatalog:CoreOptionsTest.*"`
- Fedora: `./build-codex-pr2/release/paimon-read-inte-test --gtest_filter="SystemTableReadInteTest.*"`


suxiaogang223:
Hi @lxy-9602 , could you please help review this PR when you have time?

This PR adds the next batch of system table support on top of the merged system table framework: `table$audit_log` and `table$binlog`. It includes catalog routing/schema exposure, scan/read integration, rowkind output, optional `_SEQUENCE_NUMBER`, and related tests for catalog and read paths.

The next PR will continue this work by adding the remaining metadata system tables. :)

include/paimon/defs.h
```diff
@@ -374,6 +374,12 @@ struct PAIMON_EXPORT Options {
     /// "aggregation.remove-record-on-delete" - Whether to remove the whole row in aggregation
     /// engine when delete records are received. Default value is "false".
     static const char AGGREGATION_REMOVE_RECORD_ON_DELETE[];
+    /// "table-read.sequence-number.enabled" - Whether audit_log/binlog system tables expose the
+    /// sequence number column. Default value is "false".
+    static const char TABLE_READ_SEQUENCE_NUMBER_ENABLED[];
+    /// "key-value.sequence_number.enabled" - Internal option for key-value system table reads.
+    /// Default value is "false".
+    static const char KEY_VALUE_SEQUENCE_NUMBER_ENABLED[];
```
lxy-9602, line 385: I noticed that the parameter comment here is inconsistent with the Java version. 
Could you please align them? This will help users compare and understand the API across languages.

src/paimon/core/io/key_value_projection_consumer.cpp
```diff
@@ -72,6 +72,15 @@ Result<BatchReader::ReadBatch> KeyValueProjectionConsumer::NextBatch(
         array_builder_->AppendValues(key_value_vec.size(), /*valid_bytes=*/nullptr));
     for (size_t i = 0; i < target_to_src_mapping_.size(); i++) {
         for (const auto& row : key_value_vec) {
+            if (target_to_src_mapping_[i] == kSequenceNumberProjection) {
+                auto* builder = dynamic_cast<arrow::Int64Builder*>(
+                    array_builder_->field_builder(static_cast<int>(i)));
+                if (builder == nullptr) {
```
lxy-9602, line 78: To improve type consistency, could you please replace all `int` with `int32_t`?

src/paimon/core/io/key_value_projection_consumer.h
```diff
@@ -40,6 +40,8 @@ struct KeyValue;
 class KeyValueProjectionConsumer
     : public RowToArrowArrayConverter<KeyValue, BatchReader::ReadBatch> {
  public:
+    static constexpr int32_t kSequenceNumberProjection = -1;
+
```
lxy-9602, line 46: I wonder would it be better to set `kSequenceNumberProjection` to `SpecialFieldIds::SEQUENCE_NUMBER` for stronger type consistency and intent clarity?

src/paimon/core/operation/merge_file_split_read.cpp
```diff
@@ -305,6 +318,14 @@ Status MergeFileSplitRead::GenerateKeyValueReadSchema(
     // 1. add user raw read schema to need_fields
     PAIMON_ASSIGN_OR_RAISE(std::vector<DataField> need_fields,
                            DataField::ConvertArrowSchemaToDataFields(raw_read_schema));
+    if (options.KeyValueSequenceNumberEnabled()) {
```
lxy-9602: Could you please add comments for remove `SequenceNumber` from need_fields.

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+    ChangelogBatchReader(std::unique_ptr<BatchReader> reader,
+                         std::shared_ptr<arrow::Schema> output_schema, bool include_sequence_number,
+                         std::shared_ptr<const ChangelogBatchConverter> converter,
+                         const std::shared_ptr<MemoryPool>& /*pool*/)
+        : reader_(std::move(reader)),
+          output_schema_(std::move(output_schema)),
+          include_sequence_number_(include_sequence_number),
+          converter_(std::move(converter)),
+          arrow_pool_(arrow::default_memory_pool()) {}
+
```
lxy-9602, line 72: Please construct `arrow_pool_` using the passed-in memory `pool` parameter to ensure all allocations are made from the user-specified pool.

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+            PAIMON_ASSIGN_OR_RAISE(array, CopyToStablePool(array));
+            output_arrays.push_back(array);
+        }
+
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+            std::shared_ptr<arrow::StructArray> output_array,
+            arrow::StructArray::Make(output_arrays, output_schema_->fields(),
+                                     struct_array->null_bitmap(), struct_array->null_count(),
+                                     struct_array->offset()));
+        auto output_c_array = std::make_unique<ArrowArray>();
```
lxy-9602, line 118: Consider using the overload of `arrow::StructArray::Make` that takes `field_names` instead of a full schema. 

When reading formatted data, string-type columns may be stored as dictionary arrays (e.g., due to shallow-copy optimizations). In such cases, the logical schema expects STRING, but the physical array is DICTIONARY — causing a mismatch if schema is enforced strictly.

By letting Arrow infer types from the arrays directly via the `field_names` overload, we ensure consistency between the actual data layout and the created struct array, avoiding potential type conflicts.

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+    builder.SetOptions(options_)
+        .WithStreamingMode(context->IsStreamingMode())
+        .WithMemoryPool(context->GetMemoryPool())
+        .WithExecutor(context->GetExecutor())
+        .WithFileSystem(context->GetSpecificFileSystem());
+    if (context->GetLimit()) {
+        builder.SetLimit(context->GetLimit().value());
+    }
+    PAIMON_ASSIGN_OR_RAISE(auto data_context, builder.Finish());
+    return TableScan::Create(std::move(data_context));
```
lxy-9602, line 262: Consider avoid `auto` with PAIMON_ASSIGN_OR_RAISE — please use explicit types for better readability and type clarity.

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+    if (context->GetScanFilters() && context->GetScanFilters()->GetPredicate()) {
+        return Status::NotImplemented("audit_log system table predicate pushdown is not supported");
+    }
+    ScanContextBuilder builder(table_path_);
+    builder.SetOptions(options_)
+        .WithStreamingMode(context->IsStreamingMode())
+        .WithMemoryPool(context->GetMemoryPool())
+        .WithExecutor(context->GetExecutor())
+        .WithFileSystem(context->GetSpecificFileSystem());
+    if (context->GetLimit()) {
```
lxy-9602, line 258: I wonder should we also pass `partition_filters` or `bucket_filters` through the context?

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+    void Close() override {
+        reader_->Close();
+    }
+
+ private:
+    Result<std::shared_ptr<arrow::Array>> CopyToStablePool(
+        const std::shared_ptr<arrow::Array>& array) const {
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Array> result,
+                                          arrow::Concatenate({array}, arrow_pool_));
+        return result;
```
lxy-9602, line 140: Given that paimon-cpp expects the reader to outlive any data it produces — ensuring the memory pool remains valid — I'm wondering if `CopyToStablePool` still needs to perform a deep copy in all cases. Could there be an opportunity to avoid unnecessary copies when the lifetime contract is respected?

src/paimon/core/table/system/audit_log_system_table.cpp
lxy-9602: There might be similar cases elsewhere in the codebase. Could you please take a moment to check and address them all?

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+    PAIMON_ASSIGN_OR_RAISE(auto data_context, builder.Finish());
+    PAIMON_ASSIGN_OR_RAISE(auto data_read, TableRead::Create(std::move(data_context)));
+    auto* key_value_read = dynamic_cast<KeyValueTableRead*>(data_read.get());
+    if (!key_value_read) {
+        return Status::Invalid("audit_log system table requires key-value table read");
+    }
+    key_value_read->ForceKeepDelete(true);
+    auto core_options = CoreOptions::FromMap(options_);
+    bool include_sequence_number =
```
lxy-9602: `auto core_options = CoreOptions::FromMap(options_);` -> `PAIMON_ASSIGN_OR_RAISE`

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+        core_options.ok() && core_options.value().TableReadSequenceNumberEnabled();
+    return std::make_unique<ChangelogTableRead>(std::move(data_read), ArrowSchema(),
+                                                include_sequence_number, std::move(converter),
+                                                context->GetMemoryPool());
+}
+
+std::shared_ptr<arrow::Schema> AuditLogSystemTable::BaseReadSchema() const {
+    arrow::FieldVector fields;
+    auto core_options = CoreOptions::FromMap(options_);
+    bool include_sequence_number =
```
lxy-9602: Could you please return `Result` and use `PAIMON_ASSIGN_OR_RAISE` for `CoreOptions::FromMap`?

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+        core_options.ok() && core_options.value().TableReadSequenceNumberEnabled();
+    if (include_sequence_number) {
+        fields.push_back(DataField::ConvertDataFieldToArrowField(SpecialFields::SequenceNumber()));
+    }
+    for (const auto& field : table_schema_->Fields()) {
+        fields.push_back(field.ArrowField());
+    }
```
lxy-9602, line 325: It looks like Java's `AuditLogRead` has a `withReadType` func to control what data is read, while C++'s `BaseReadSchema` always reads all fields.

Is the current "read all" behavior in C++ by design, or could it be extended to support finer-grained read modes?

Supporting partial field reads might improve performance and align better with the Java counterpart.

src/paimon/core/table/system/system_table.h
```diff
@@ -60,12 +58,14 @@ class SystemTableLoader {
 
     static Result<std::shared_ptr<SystemTable>> Load(
         const std::string& system_table_name, const std::shared_ptr<FileSystem>& fs,
-        const std::string& table_path, const std::shared_ptr<TableSchema>& table_schema);
+        const std::string& table_path, const std::shared_ptr<TableSchema>& table_schema,
+        const std::map<std::string, std::string>& dynamic_options = {});
```
lxy-9602: To improve clarity and maintainability, could you please avoid default parameters and prefer explicit parameter specification?

test/inte/read_inte_test.cpp
```diff
@@ -546,6 +593,174 @@ TEST(SystemTableReadInteTest, TestReadBranchOptionsSystemTable) {
...
+    ASSERT_TRUE(rowkind_array);
+    ASSERT_TRUE(pk_array);
+    ASSERT_TRUE(v_array);
+    ASSERT_EQ(rowkind_array->GetString(0), "+I");
+    ASSERT_EQ(rowkind_array->GetString(1), "+I");
+    ASSERT_EQ(pk_array->value_length(0), 1);
+    ASSERT_EQ(pk_array->value_length(1), 1);
+    ASSERT_EQ(v_array->value_length(0), 1);
+    ASSERT_EQ(v_array->value_length(1), 1);
+    auto pk_values = std::dynamic_pointer_cast<arrow::StringArray>(pk_array->values());
```
lxy-9602: The current test assertions for Binlog/AuditLog are quite detailed.  
Could we instead express expected results as Arrow array JSON and use `array->Equals()` for comparison?

test/inte/read_inte_test.cpp
```diff
@@ -546,6 +593,174 @@ TEST(SystemTableReadInteTest, TestReadBranchOptionsSystemTable) {
...
+                                                         /*is_streaming_mode=*/false));
+
+    ASSERT_NOK_WITH_MSG(
+        ReadSystemTable(PathUtil::JoinPath(dir->Str(), "foo.db/bar$audit_log"), options),
+        "only supports primary key table");
+    ASSERT_NOK_WITH_MSG(
+        ReadSystemTable(PathUtil::JoinPath(dir->Str(), "foo.db/bar$binlog"), options),
+        "only supports primary key table");
+}
+
```
lxy-9602, line 899: Could you please add tests with:
- Branch read
- A field that exists but has null values (in binlog, read batch maybe list of [null])
- Containing delete or update rows
- Read partial fields if supported.

We can reuse data from `test/test_data/parquet` or `orc`.

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,309 @@
...
+    return arrow::schema(fields);
+}
+
+std::map<std::string, std::string> AuditLogSystemTable::ReadOptions() const {
+    auto read_options = options_;
+    read_options[Options::KEY_VALUE_SEQUENCE_NUMBER_ENABLED] = "true";
+    return read_options;
+}
```
lxy-9602, line 336: Java only enables `KEY_VALUE_SEQUENCE_NUMBER_ENABLED` when `TABLE_READ_SEQUENCE_NUMBER_ENABLED` is set while C++ enables it unconditionally.  

This works now, but for cross-language consistency, should we align C++ behavior with Java?

lxy-9602:
Thanks for the submission! It's great to see not just new features, but also thoughtful cleanups and small refactorings you've made🚀.
Looking forward to your adjustments and continued contributions!

src/paimon/core/table/system/audit_log_system_table.cpp
suxiaogang223: I agree both partial field read and avoiding unnecessary deep copies are valuable optimizations, but I’d prefer not to include them in this PR to keep the scope focused on batch correctness for audit_log/binlog.

For partial field read, audit_log/binlog still need internal fields such as _VALUE_KIND, optional _SEQUENCE_NUMBER, primary keys, and sequence fields for correct key-value changelog reading. Supporting projection properly would require an explicit system-table projection layer: map requested system table fields to required base table fields, read the internal required fields, and then project back to the requested output schema. This is doable, but it is more than a small adjustment to BaseReadSchema().

For the deep copy, the current implementation is intentionally conservative. The system table reader imports batches from the underlying C Arrow interface and then exports a new system-table batch. Keeping copied arrays in the system table reader’s pool avoids depending on the lifetime details of the wrapped reader’s returned buffers. We can optimize this later once we make the ownership contract explicit and add coverage for the returned batch lifetime.

TODOs for follow-up:
- Add a system-table projection layer for audit_log/binlog.
- Support partial reads for data fields while still reading required internal fields.
- Add tests for projected audit_log/binlog reads.
- Revisit CopyToStablePool() and introduce zero-copy paths where ownership is safe.

src/paimon/core/table/system/binlog_system_table.cpp
```diff
@@ -0,0 +1,89 @@
...
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(offsets_builder.Reserve(array->length() + 1));
+        for (int64_t i = 0; i <= array->length(); ++i) {
+            PAIMON_RETURN_NOT_OK_FROM_ARROW(offsets_builder.Append(static_cast<int32_t>(i)));
+        }
+        std::shared_ptr<arrow::Array> offsets_array;
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(offsets_builder.Finish(&offsets_array));
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+            std::shared_ptr<arrow::Array> list_array,
+            arrow::ListArray::FromArrays(*offsets_array, *array, pool));
+        return list_array;
```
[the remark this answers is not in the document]
suxiaogang223: I agree that fixed-size list could reduce some per-batch overhead for the current batch-mode binlog output, where each row currently contains exactly one element.

I’d prefer not to include this optimization in this PR because it may change the exposed Arrow type from list<T> to fixed_size_list<T, 1>. The current schema is intentionally aligned with the Java binlog system table semantics, where fields are represented as arrays/lists and future streaming support may need before/after pairs rather than always a single element.

So I’d like to keep this PR focused on correctness and Java-aligned behavior for audit_log/binlog, and leave this as a follow-up TODO once we decide the public Arrow schema contract for binlog and streaming before/after output.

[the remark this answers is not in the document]
lxy-9602: I agree with you.

include/paimon/defs.h
lxy-9602: `table-read.sequence-number.enabled`- Whether to include the _SEQUENCE_NUMBER field when reading the audit_log or binlog system tables. This is only valid for primary key tables.

`key-value.sequence_number.enabled` - Whether to include the _SEQUENCE_NUMBER field when reading key-value data. This is an internal option used by AuditLogTable and BinlogTable when table.read_sequence_number_enabled is set to true.

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,336 @@
...
+            }
+            PAIMON_ASSIGN_OR_RAISE(sequence_array, CopyToStablePool(sequence_array));
+            output_arrays.push_back(sequence_array);
+        }
+
+        for (const auto& field : output_schema_->fields()) {
+            if (field->name() == "rowkind" ||
+                field->name() == SpecialFields::SequenceNumber().Name()) {
+                continue;
+            }
```
lxy-9602, line 105: I double-checked the Java implementation, and it uses `rowkind` (and optionally `_SEQUENCE_NUMBER`) as the field names in `AuditLogTable`.  
Currently, the C++ side happens to use the same names, so behavior is consistent.

To better ensure cross-language consistency, could you please:
- Add `rowkind` to `SpecialFields` and `SpecialFieldIds`
- Avoid hardcoding `"rowkind"` directly in `audit_log_system_table.cpp`

Thanks!

lxy-9602 (dismissed):
+1

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -0,0 +1,339 @@
...
+}
+
+std::shared_ptr<arrow::Schema> AuditLogSystemTable::ArrowSchema() const {
+    std::shared_ptr<arrow::Field> rowkind_field =
+        DataField::ConvertDataFieldToArrowField(SpecialFields::RowKind());
+    rowkind_field = rowkind_field->WithNullable(false);
+    arrow::FieldVector fields = {rowkind_field};
+    Result<CoreOptions> core_options = CoreOptions::FromMap(options_);
+    bool include_sequence_number =
+        core_options.ok() && core_options.value().TableReadSequenceNumberEnabled();
```
lszskye: supposed to deal with core_options not ok, return bad status

lxy-9602 (approved):
+1

PR #278, opened 2026-05-12
refactor(global_index): adapt global index id in pre-filter

lszskye:
### Purpose
change **local index id**  to **global index id** in pre-filter

### Tests
global_index_test.cpp

### DIFF
--- a/test/inte/global_index_test.cpp
+++ b/test/inte/global_index_test.cpp
@@ -2136,22 +2136,20 @@ TEST_P(GlobalIndexTest, TestWriteCommitScanReadLuceneIndexWithPartition) {
             readers[0]->VisitFullTextSearch(std::make_shared<FullTextSearch>(
                 "f0",
                 /*limit=*/10, "document", FullTextSearch::SearchType::MATCH_ALL, pre_filter)));
-        ASSERT_EQ(index_result->ToString(), index_expected);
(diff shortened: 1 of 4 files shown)


test/inte/global_index_test.cpp
```diff
@@ -2073,6 +2073,87 @@ TEST_P(GlobalIndexTest, TestLuceneWriteCommitScanReadIndexWithScore) {
...
+        scan_and_check_result({{"f1", "10"}}, row_range_index, filter,
+                              "row ids: {0}, scores: {0.59}");
+    }
+    {
+        // test scan and read for f1=20
+        auto filter = RoaringBitmap64::From(std::vector<int64_t>({2l}));
+        ASSERT_OK_AND_ASSIGN(RowRangeIndex row_range_index, RowRangeIndex::Create({Range(2, 3)}));
+        scan_and_check_result({{"f1", "20"}}, row_range_index, filter,
+                              "row ids: {2}, scores: {2.00}");
+    }
```
lxy-9602, line 2153: Let’s avoid asserting on FTS score unless we have a strong reason.
It’s often non-deterministic or hard to interpret, and such checks can make tests brittle.

lszskye: ok

lxy-9602 (approved):
+1

PR #282, opened 2026-05-13
enhance: improve dependency source resolution

suxiaogang223:
## Summary

- Disable `SYSTEM` resolution for Arrow and ORC while paimon-cpp depends on project-specific patches.
- Make `AUTO` resolution for Arrow and ORC fall back to bundled builds.
- Add shared helper logic for validating that imported CMake targets expose required headers.
- Apply system dependency header validation to GTest and other `FindXXXAlt` config-package paths.

## Why

Arrow and ORC currently carry project-specific patches in paimon-cpp, so resolving them from system packages can miss required behavior. Some system dependency config packages can also expose library targets without the corresponding development headers, which lets configure pass and build fail later.

## Validation

- `cmake -S . -B build/codex-deps-check -DPAIMON_BUILD_TESTS=OFF -DPAIMON_ENABLE_ORC=OFF -DPAIMON_DEPENDENCY_SOURCE=BUNDLED`
- `cmake -S . -B build/codex-auto-check -DPAIMON_BUILD_TESTS=OFF -DPAIMON_ENABLE_ORC=OFF`
- `cmake -S . -B build/codex-arrow-system-check -DArrow_SOURCE=SYSTEM -DPAIMON_ENABLE_ORC=OFF` fails as expected.
- `cmake -S . -B build/codex-gtest-system-check -DPAIMON_BUILD_TESTS=ON -DPAIMON_ENABLE_ORC=ON -DPAIMON_DEPENDENCY_SOURCE=BUNDLED -DGTest_SOURCE=SYSTEM` fails at configure time when the full GTest development package is unavailable.


cmake_modules/FindGTestAlt.cmake
```diff
@@ -65,8 +66,20 @@ if(NOT TARGET GTest::gtest
...
+                               gtest/gtest.h
+                               ${_PAIMON_GTEST_FIND_ARGS})
+    paimon_find_target_headers(GMOCK_INCLUDE_DIR
+                               GTest::gmock
+                               NAMES
+                               gmock/gmock.h
+                               ${_PAIMON_GTEST_FIND_ARGS})
+
+    include(FindPackageHandleStandardArgs)
+    find_package_handle_standard_args(GTestAlt REQUIRED_VARS GTEST_INCLUDE_DIR
```
zjw1111, line 81: **[P1] `AUTO` fallback can fail after a partial config package creates targets.**  
  In `AUTO` mode, `resolve_dependency()` calls `find_package(... QUIET MODULE)`, and if `*_Alt_FOUND` is false it falls back to bundled. But the new config-package path can create imported targets before header validation fails. Those targets remain globally defined, so bundled builders that create the same target will fail with “target already exists”.  
  Example: `FindRE2Alt.cmake` can leave `re2::re2` defined while `RE2Alt_FOUND=false`; then `build_re2()` calls `add_library(re2::re2 ...)`. Same shape exists for `GTest::gtest`/`GTest::gmock`, and possibly `RapidJSON`, `tbb`, `glog`, `zstd` when the system config target name matches the bundled compatibility target. This hits exactly the “config target exists but dev headers are incomplete” case the PR is trying to handle.  
  Suggested fix: for `AUTO`, avoid falling back after a config package has created conflicting targets, or make the `Find*Alt` modules validate headers before accepting/creating compatibility targets and return a hard configure error for partial packages.

------

The `GTest::gtest` target has already been generated on line 22 (`find_package(GTest CONFIG QUIET ${_PAIMON_GTEST_FIND_ARGS})`). An error will occur indicating that the `GTest::gtest` target already exists when the system library is found to be missing header files and source installation is required.

You can reproduce this issue by running `yum install gtest gmock` (but not installing `gtest-devel` and `gmock-devel`).

Image

zjw1111 (dismissed):
+1

zjw1111:
Thanks for enhancing the dependency build module. Great work! 👍🏻

zjw1111:
@suxiaogang223 Hi, I also found a minor issue that needs adjustment. The configuration item regarding `-DArrow_SOURCE=SYSTEM` in the document needs to be updated.

lucasfang (approved):
+1

lucasfang:
Thank you for your contribution, nice work!

PR #68, opened 2026-01-20
feat: Add global config API and optimize Parquet read thread conf

Eyizoha:
### Purpose

- Add global_config.h/cpp for managing global configurations
- Change Parquet read thread configuration from thread count to boolean switch
    The current configuration approach modifies global settings at the reader level, which is not reasonable. It affects all code using the Arrow thread pool in the process and is not always effective in concurrent scenarios. Additionally, its default value is too small, impacting Parquet read performance under default settings.

### Tests

Update related test cases to adapt to the new configuration

### API and Format

Added a new global configuration module
Modified Parquet read configuration

### Documentation

None

PR #284, opened 2026-05-15
fix: Restore direct Arrow thread pool control inside Parquet format library

lxy-9602:
### Purpose

Linked issue: #68 

### Motivation

When Arrow is statically linked into multiple shared libraries (`libpaimon.so` and `libpaimon_parquet_file_format.so`), each `.so` gets its own copy of Arrow's `CpuThreadPool` singleton. 

Commit d3bb3a9 introduced `paimon::SetArrowCpuThreadPoolCapacity()` in `libpaimon.so` as a wrapper around `arrow::SetCpuThreadPoolCapacity()`, and removed the direct `arrow::SetCpuThreadPoolCapacity()` call from `parquet_file_batch_reader.cpp` (which lives in `libpaimon_parquet_file_format.so`). This means setting the thread pool capacity through the wrapper only affects the singleton inside `libpaimon.so`, **never** the one inside `libpaimon_parquet_file_format.so` — making Parquet read thread control completely ineffective.

On a 96-core machine, this caused Arrow to spawn 96 CpuThreadPool workers + 8 IOThreadPool workers inside `libpaimon_parquet_file_format.so`, regardless of any capacity setting by the user.

### Changes

This PR reverts commit d3bb3a9 to restore the original behavior where `parquet_file_batch_reader.cpp` directly calls `arrow::SetCpuThreadPoolCapacity()`, ensuring the call targets the correct singleton within `libpaimon_parquet_file_format.so`.


lxy-9602:
@Eyizoha I came across some unexpected behavior related to the earlier change #68 and opened this pr to fix.
The root cause and solution are detailed in the PR description.
Would you mind taking a quick look to see if you agree with the approach?

lucasfang (approved):
+1

Eyizoha:
+1

PR #285, opened 2026-05-16
feat: Add metadata system tables

suxiaogang223:
## Background

This PR is part of #141 and continues the system table work after the previously merged `options`, `audit_log`, and `binlog` support.

The scope of this PR is table-level metadata system tables. It adds read-only support for metadata that can be served from existing table metadata files, so users can inspect snapshots, schemas, tags, branches, and consumers through the existing system table query path.

## Architecture

- Introduce `InMemorySystemTable` for system tables whose output can be materialized as a single in-memory Arrow `RecordBatch`.
  - It centralizes singleton split scan, one-shot table read, and Arrow C data export.
  - `options`, `snapshots`, `schemas`, `tags`, `branches`, and `consumers` all use this execution model.

- Keep metadata table classes Java-aligned and avoid an extra metadata base class.
  - `OptionsSystemTable`, `SnapshotsSystemTable`, `SchemasSystemTable`, `TagsSystemTable`, `BranchesSystemTable`, and `ConsumersSystemTable` directly inherit `InMemorySystemTable`.
  - Shared metadata location state is held by a small `MetadataSystemTableContext` instead of a thin inheritance layer.
  - `audit_log` and `binlog` intentionally remain outside this hierarchy because they are data/changelog-backed system tables, not in-memory metadata tables.

- Group pure metadata system tables in `metadata_system_tables`.
  - `options` is now colocated with the other in-memory metadata system tables because it has the same scan/read shape and row-to-Arrow conversion path.
  - `options` differs only in data source: it reads latest schema options, while the new tables read metadata through managers.

- Reuse row-to-Arrow conversion for in-memory system table output.
  - Add `GenericRowToArrowArrayConverter` on top of the existing `RowToArrowArrayConverter` infrastructure.
  - In-memory metadata system tables build `GenericRow` values and convert them through the shared converter instead of maintaining table-local Arrow builders.

- Refactor `SystemTableLoader` to use a registry/factory table.
  - `IsSupported` and `Load` now use the same registry entry list.
  - This avoids adding each new system table name in multiple places.

- Add metadata managers/helpers where needed.
  - `BranchManager::ListBranches` is the shared branch listing entry point.
  - `ConsumerManager` owns consumer path/list/read logic for `consumer/consumer-*` files.

## System Tables Added

This PR follows the Apache Paimon table-scoped system table model, where table metadata is queried with names like `table$snapshots`. Reference semantics: https://paimon.apache.org/docs/master/concepts/system-tables/

- `table$snapshots`
  - Purpose: exposes snapshot commit history for the data table.
  - Source: snapshot metadata files through `SnapshotManager`.
  - Columns: `snapshot_id`, `schema_id`, `commit_user`, `commit_identifier`, `commit_kind`, `commit_time`, `base_manifest_list`, `delta_manifest_list`, `changelog_manifest_list`, `total_record_count`, `delta_record_count`, `changelog_record_count`, `watermark`, `next_row_id`.
  - Ordering: sorted by `snapshot_id` ascending.

- `table$schemas`
  - Purpose: exposes schema evolution history and allows snapshot schema ids to be resolved to concrete schema versions.
  - Source: schema metadata files through `SchemaManager`.
  - Columns: `schema_id`, `fields`, `partition_keys`, `primary_keys`, `options`, `comment`, `update_time`.
  - Note: `fields`, `partition_keys`, `primary_keys`, and `options` are returned as JSON strings in this first C++ version.
  - Ordering: sorted by `schema_id` ascending.

- `table$tags`
  - Purpose: exposes table tags and the snapshots they reference.
  - Source: tag metadata through `TagManager`.
  - Columns: `tag_name`, `snapshot_id`, `schema_id`, `commit_time`, `record_count`, `create_time`, `time_retained`.

- `table$branches`
  - Purpose: exposes available table branches, including `main`.
  - Source: branch metadata paths through `BranchManager`.
  - Columns: `branch_name`, `create_time`.

- `table$consumers`
  - Purpose: exposes persisted streaming consumer progress by consumer id.
  - Source: consumer metadata files through `ConsumerManager`.
  - Columns: `consumer_id`, `next_snapshot_id`.
  - Note: supports both plain numeric consumer files and Java-style JSON files with `nextSnapshot`.

## Scope Notes

- This PR does not implement heavier file/manifest-backed metadata tables such as `files` and `manifests`; those are planned for a follow-up PR.
- Existing `options`, `audit_log`, and `binlog` behavior is preserved.

## Tests

- Fedora: `cmake --build build-codex-metadata-pr3 -j$(nproc)`
- Fedora: `cmake --build build-codex-metadata-pr3 --target paimon-core-test paimon-read-inte-test -j2`
- Fedora: `./build-codex-metadata-pr3/debug/paimon-core-test --gtest_filter=FileSystemCatalogTest.TestMetadataSystemTableCatalog`
- Fedora: `./build-codex-metadata-pr3/debug/paimon-read-inte-test --gtest_filter=SystemTableReadInteTest.TestReadMetadataSystemTables:SystemTableReadInteTest.TestReadTagBranchAndConsumerSystemTables`
- pre-commit passed


src/paimon/core/table/system/in_memory_system_table.cpp
```diff
@@ -0,0 +1,120 @@
...
+        if (emitted_) {
+            return BatchReader::MakeEofBatch();
+        }
+        emitted_ = true;
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::RecordBatch> record_batch,
+                               table_.BuildRecordBatch(arrow_pool_.get()));
+        std::shared_ptr<arrow::StructArray> struct_array =
+            std::make_shared<arrow::StructArray>(arrow::struct_(record_batch->schema()->fields()),
+                                                 record_batch->num_rows(), record_batch->columns());
+        auto c_array = std::make_unique<ArrowArray>();
```
lxy-9602: Could you please use `StructArray::Make` instead? It ensures that any Arrow-level validation errors are properly propagated up the call stack, rather than being silently ignored or coredump.

src/paimon/core/table/system/in_memory_system_table.cpp
```diff
@@ -0,0 +1,120 @@
...
+        return std::make_shared<MetricsImpl>();
+    }
+
+    void Close() override {
+        emitted_ = true;
+    }
+
+ private:
+    const InMemorySystemTable& table_;
+    std::unique_ptr<arrow::MemoryPool> arrow_pool_;
```
lxy-9602, line 64: I'm not entirely confident that holding a reference to the `Table` is sufficient.
Is there a scenario where the `Table` object could be destroyed while a `Reader` is still alive?

To be safe, should the `Reader` hold a ptr instead, to ensure the table remains valid for the lifetime of the reader?

src/paimon/core/table/system/metadata_system_table.h
```diff
@@ -0,0 +1,39 @@
...
+
+namespace paimon {
+class FileSystem;
+
+// Base class for table metadata system tables backed by Paimon metadata files.
+// Subclasses read metadata from the selected table branch and materialize it in memory.
+class MetadataSystemTable : public InMemorySystemTable {
+ public:
+    MetadataSystemTable(std::shared_ptr<FileSystem> fs, std::string table_path, std::string branch);
+
```
lxy-9602: Could you please use const T& for this parameter? It aligns with our existing coding style.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,394 @@
...
+        builders_.reserve(schema_->num_fields());
+        for (const auto& field : schema_->fields()) {
+            std::unique_ptr<arrow::ArrayBuilder> builder;
+            init_status_ = ToPaimonStatus(arrow::MakeBuilder(pool, field->type(), &builder));
+            if (!init_status_.ok()) {
+                return;
+            }
+            builders_.push_back(std::move(builder));
+        }
+    }
```
lxy-9602, line 84: The design of `InitStatus` seems a bit inconsistent with patterns used in paimon-cpp.
If we need to report errors during initialization, consider using a static `Create` factory function instead of relying on post-construction status checks.

Alternatively, we could lazily initialize the `builders_` object (e.g., on first use) to avoid partial construction states.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,394 @@
...
+            PAIMON_RETURN_NOT_OK_FROM_ARROW(builder->Append(value.value()));
+        } else {
+            PAIMON_RETURN_NOT_OK_FROM_ARROW(builder->AppendNull());
+        }
+        return Status::OK();
+    }
+
+    Status AppendOptionalDouble(int32_t index, const std::optional<double>& value) {
+        auto* builder = dynamic_cast<arrow::DoubleBuilder*>(builders_.at(index).get());
+        if (builder == nullptr) {
```
lxy-9602: Consider templating the append operation based on the builder type.
Arrow’s dispatch macros (like those used in `orc_adapter.cpp`'s Appender) could help eliminate repetitive type-switching code.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,394 @@
...
+            if (i == 0) {
+                num_rows = array->length();
+            } else if (array->length() != num_rows) {
+                return Status::Invalid("metadata field ", schema_->field(i)->name(),
+                                       " length does not match previous fields");
+            }
+            arrays.push_back(std::move(array));
+        }
+        return arrow::RecordBatch::Make(schema_, num_rows, std::move(arrays));
+    }
```
lxy-9602, line 128: Could you clarify whether this `Result` is from `Arrow` or `Paimon`?
If it's `Arrow`'s, please qualify it with arrow:: to make the origin explicit.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,394 @@
...
+        arrow::field("schema_id", arrow::int64(), /*nullable=*/false),
+        arrow::field("commit_user", arrow::utf8(), /*nullable=*/false),
+        arrow::field("commit_identifier", arrow::int64(), /*nullable=*/false),
+        arrow::field("commit_kind", arrow::utf8(), /*nullable=*/false),
+        arrow::field("commit_time", arrow::int64(), /*nullable=*/false),
+        arrow::field("total_record_count", arrow::int64(), /*nullable=*/true),
+        arrow::field("delta_record_count", arrow::int64(), /*nullable=*/true),
+        arrow::field("changelog_record_count", arrow::int64(), /*nullable=*/true),
+        arrow::field("watermark", arrow::int64(), /*nullable=*/true),
+    });
```
lxy-9602, line 213: I compared the meta table schema between the Java and C++ implementations, and noticed a few discrepancies. For example:

- `SnapshotsTable` has 14 fields in Java, but only 10 in `paimon-cpp`.
- The `commit_time` field is represented as `TIMESTAMP` in Java, but as `INT64` in C++.

Are these differences intentional (e.g., due to language-specific optimizations), or should we aim to align the schemas for better cross-language compatibility?  
It would be great to clarify the expected behavior so we can ensure consistent metadata interpretation.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+Result<std::shared_ptr<arrow::Schema>> SchemasSystemTable::ArrowSchema() const {
+    return arrow::schema({
+        arrow::field("schema_id", arrow::int64(), /*nullable=*/false),
+        arrow::field("fields", arrow::utf8(), /*nullable=*/false),
+        arrow::field("partition_keys", arrow::utf8(), /*nullable=*/false),
+        arrow::field("primary_keys", arrow::utf8(), /*nullable=*/false),
+        arrow::field("options", arrow::utf8(), /*nullable=*/false),
+        arrow::field("comment", arrow::utf8(), /*nullable=*/true),
+    });
```
lxy-9602, line 268: I also noticed that the `SchemasSystemTable` in C++ might be missing the update_time field, which exists in the Java implementation.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+Result<std::shared_ptr<arrow::Schema>> TagsSystemTable::ArrowSchema() const {
+    return arrow::schema({
+        arrow::field("tag_name", arrow::utf8(), /*nullable=*/false),
+        arrow::field("snapshot_id", arrow::int64(), /*nullable=*/false),
+        arrow::field("schema_id", arrow::int64(), /*nullable=*/false),
+        arrow::field("tag_create_time", arrow::utf8(), /*nullable=*/true),
+        arrow::field("tag_time_retained", arrow::float64(), /*nullable=*/true),
+    });
+}
```
lxy-9602, line 326: Similar to other `SystemTable`, the `TagsSystemTable` and `BranchesSystemTable`'s schema also seems inconsistent with the Java side.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,394 @@
...
+    for (const auto& name : tag_names) {
+        PAIMON_ASSIGN_OR_RAISE(Tag tag, tag_manager.GetOrThrow(name));
+        PAIMON_ASSIGN_OR_RAISE(std::optional<std::string> tag_create_time_json,
+                               OptionalJson(tag.TagCreateTime()));
+        PAIMON_RETURN_NOT_OK(rows.AppendString(0, name));
+        PAIMON_RETURN_NOT_OK(rows.AppendInt64(1, tag.Id()));
+        PAIMON_RETURN_NOT_OK(rows.AppendInt64(2, tag.SchemaId()));
+        PAIMON_RETURN_NOT_OK(rows.AppendOptionalString(3, tag_create_time_json));
+        PAIMON_RETURN_NOT_OK(rows.AppendOptionalDouble(4, tag.TagTimeRetained()));
+    }
```
lxy-9602, line 350: While we could define a custom row type and corresponding append logic, I wonder if it's simpler to reuse existing infrastructure.

For example, we could construct a `GenericRow` and use the already-implemented `RowToArrowArrayConverter` to convert it into an Arrow array.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,394 @@
...
+
+std::string ConsumersSystemTable::Name() const {
+    return kName;
+}
+
+Result<std::shared_ptr<arrow::Schema>> ConsumersSystemTable::ArrowSchema() const {
+    return arrow::schema({
+        arrow::field("consumer_id", arrow::utf8(), /*nullable=*/false),
+        arrow::field("next_snapshot_id", arrow::int64(), /*nullable=*/true),
+    });
```
lxy-9602, line 407: In Java, `next_snapshot_id` is non-nullable (nullable = false). Is it intended to be nullable here?

src/paimon/core/table/system/audit_log_system_table.h
```diff
@@ -27,6 +27,7 @@ namespace paimon {
 class FileSystem;
 class TableSchema;
 
+// Converts data columns when wrapping the base table changelog reader.
 class ChangelogBatchConverter {
```
lxy-9602, line 31: Could you please update the comment style to follow Doxygen format?
For example, for single-line comments, we generally use `///` instead of `//` to ensure they are picked up by documentation generators.

zjw1111:
@suxiaogang223 Hi, due to some temporary issues with GitHub Copilot, the Copilot review results could not be displayed directly. I extracted the review results from the Copilot logs for your reference.
image
image
image

src/paimon/core/utils/branch_manager.cpp
```diff
@@ -0,0 +1,54 @@
...
+            continue;
+        }
+        std::string dir_name = PathUtil::GetName(file_status->GetPath());
+        if (StringUtils::StartsWith(dir_name, branch_prefix, /*start_pos=*/0)) {
+            branches.push_back(dir_name.substr(branch_prefix.length()));
+        }
+    }
+    std::sort(branches.begin(), branches.end());
+    branches.erase(std::unique(branches.begin(), branches.end()), branches.end());
+    return branches;
```
lxy-9602, line 50: I wonder are duplicate branch paths actually possible?
If the system ensures uniqueness (e.g., via naming rules), do we still need to deduplicate here?

src/paimon/core/utils/consumer_manager.h
```diff
@@ -0,0 +1,48 @@
...
+#include "paimon/result.h"
+
+namespace paimon {
+
+class FileSystem;
+
+class ConsumerManager {
+ public:
+    static constexpr char CONSUMER_PREFIX[] = "consumer-";
+
```
lxy-9602, line 39: Could you please change `CONSUMER_PREFIX` to `kConsumerPrefix`?

src/paimon/core/utils/consumer_manager.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+Result<std::optional<int64_t>> ConsumerManager::GetNextSnapshotId(
+    const std::string& consumer_id) const {
+    std::string content;
+    Status read_status = fs_->ReadFile(ConsumerPath(consumer_id), &content);
+    if (!read_status.ok()) {
+        return std::optional<int64_t>();
+    }
+    return StringUtils::StringToValue<int64_t>(content);
+}
```
lxy-9602, line 118: If an I/O error occurs or the string fails to parse as `int64`, maybe we should return an explicit error status instead of proceeding silently.

Silently ignoring these issues can mask problems and make debugging much harder later.

lxy-9602: I noticed that in the Java implementation, `Consumer::fromPath()` includes a retry mechanism when reading consumer metadata.

If the file does not exist initially or all retries fail, the consumer is silently skipped in `ConsumersTable`.

To ensure consistent behavior across runtimes, could we align the C++ version with this pattern?

src/paimon/core/utils/tag_manager.cpp
```diff
@@ -53,8 +55,35 @@ Result<std::optional<Tag>> TagManager::Get(const std::string& tag_name) const {
...
+
 std::string TagManager::TagPath(const std::string& tag_name) const {
-    return PathUtil::JoinPath(BranchManager::BranchPath(root_path_, branch_),
-                              "/tag/" + std::string(TAG_PREFIX) + tag_name);
+    return PathUtil::JoinPath(TagDirectory(), std::string(TAG_PREFIX) + tag_name);
+}
+
+std::string TagManager::TagDirectory() const {
+    return PathUtil::JoinPath(BranchManager::BranchPath(root_path_, branch_), "/tag");
 }
```
lxy-9602, line 88: Could you please omit the leading or trailing `/tag` in string literals — `PathUtil` handles path separators automatically.

lxy-9602:
Thanks for the pr! The system table implementation is shaping up well — great work!

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,419 @@
...
+VariantType TimestampMillisValue(int64_t value) {
+    return Timestamp::FromEpochMillis(value);
+}
+
+VariantType OptionalTimestampMillisValue(const std::optional<int64_t>& value) {
+    if (!value) {
+        return NullType();
+    }
+    return TimestampMillisValue(value.value());
+}
```
lxy-9602, line 148: It seems that Java may be performing a timezone conversion, interpreting the UTC time as local time based on the system's timezone.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -0,0 +1,447 @@
...
+    }
+    return std::string(buffer.GetString(), buffer.GetSize());
+}
+
+Result<int64_t> LocalDateTimePartsToTimestampMillis(const std::vector<int64_t>& parts) {
+    if (parts.size() < 6) {
+        return Status::Invalid("tag create time requires at least 6 date-time fields");
+    }
+
+    int64_t year = parts[0];
```
lxy-9602, line 68: I reviewed the code, and I wonder if we could reuse the pattern in `DateTimeUtils::GetCurrentLocalTimeUs` for this time conversion. Perhaps we could introduce a helper such as `ToLocalTimestamp`, which would let us rely on Arrow's timezone conversion and reduce duplicated logic. Thanks!

src/paimon/core/core_options.cpp
```diff
@@ -639,7 +639,7 @@ struct CoreOptions::Impl {
         // Parse aggregation_remove_record_on_delete
         PAIMON_RETURN_NOT_OK(parser.Parse<bool>(Options::AGGREGATION_REMOVE_RECORD_ON_DELETE,
                                                 &aggregation_remove_record_on_delete));
```
lszskye, line 668: remain // in function

src/paimon/core/utils/tag_manager.cpp
```diff
@@ -53,8 +55,35 @@ Result<std::optional<Tag>> TagManager::Get(const std::string& tag_name) const {
     return std::optional<Tag>(std::move(tag));
 }
```
lszskye, line 57: Could you please add UT for TagManager::ListTagNames()

lxy-9602 (approved):
+1

PR #302, opened 2026-05-26
feat(benchmark): add append/pk table benchmark

### LINKED ISSUE #100: [Feature] Add micro benchmarks for paimon c++.
### Search before asking

### Motivation

Add benchmarks for table reading and writing.

### Solution

Add benchmarks using google benchmark framework.

### Anything else?

_No response_

- [ ] I'm willing to submit a PR!

lucasfang:
### Purpose

Linked issue: close #100

- Add a Google Benchmark based performance suite covering:
  1. BM_Write (append table write)
  2. BM_Read (append table read with prefetch_parallel variants)
  3. BM_PK_Write (primary key table write)
  4. BM_MOR_Read (primary key MOR read with prefetch_parallel variants)

- Add a benchmark entrypoint and custom CLI parsing for:
  1. --paimon_source_parquet
  2. --paimon_source_table_path
  3. --paimon_pk_columns
  4. --paimon_option (repeatable)

- Add benchmark CLI behavior and validation improvements:
  1. Parse benchmark-specific CLI arguments without throwing exceptions
  2. Return paimon::Status / paimon::Result<T> for parse and setup failures
  3. Handle --help before benchmark-specific argument parsing
  4. Trim paimon_option key/value segments
  5. Configure file format through --paimon_option file.format:<format>

- Add benchmark helper utilities for:
  1. Validation and skip behavior for invalid configurations
  2. Shared read iteration execution
  3. Source table read mode (skip pre-write stage)
  4. Source parquet loading and cache reuse

- Build integration updates:
  1. Add PAIMON_BUILD_BENCHMARKS option (default OFF)
  2. Add benchmark subdirectory and benchmark target
  3. Add add_paimon_benchmark macro and benchmark labels
  4. Add benchmark dependency discovery via FindbenchmarkAlt
  5. Add PAIMON_BENCHMARK_BUILD_VERSION=1.9.1 in third_party versions

### Tests

UT:
1. CliOptionParsingTest.ConsumeCliOptionWorks
2. CliOptionParsingTest.ParseCsvColumnsWorks
3. CliOptionParsingTest.ParseCsvColumnsRejectsInvalidInput
4. CliOptionParsingTest.ParseDelimitedOptionsWorks
5. CliOptionParsingTest.ParseDelimitedOptionsRejectsInvalidInput
6. CliOptionParsingTest.ParseStringOptionArgWorksForEqualsAndSeparatedForms
7. CliOptionParsingTest.ParseStringOptionArgRejectsMissingValue
8. CliOptionParsingTest.ParseCsvOptionArgAndDelimitedRepeatableOptionArgWorks

Benchmark smoke:
1. BM_Write
2. BM_Read/1, BM_Read/2, BM_Read/4
3. BM_PK_Write
4. BM_MOR_Read/1, BM_MOR_Read/2, BM_MOR_Read/4

Build:
1. paimon-cli-option-parsing-test
2. paimon-read-write-benchmark

IT:
1. No new integration tests in this PR (benchmark and build integration scope)

### API and Format

- No change to public API under include.
- No storage format or protocol compatibility changes.
- Changes are limited to benchmark executable behavior, benchmark helpers, tests, documentation, and build wiring.

### Documentation

- Add benchmark usage documentation.
- Document source parquet mode, source table mode, primary key columns, prefetch_parallel variants, and paimon_option usage.
- Document file format override through --paimon_option file.format:<format>.

### Generative AI tooling


benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+
+if(NOT PAIMON_BUILD_BENCHMARKS AND NOT PAIMON_BUILD_TESTS)
+    return()
+endif()
+
+find_package(Threads REQUIRED)
+
+set(PAIMON_BENCHMARK_STATIC_LINK_LIBS
+    paimon_shared
+    "-Wl,--whole-archive"
```
zjw1111: use variable such as `PAIMON_LOCAL_FILE_SYSTEM_SHARED_LINK_LIBS`

benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+                         benchmark_case_pk_write.cpp
+                         benchmark_case_mor_read.cpp
+                         read_write_benchmark.cpp
+                         STATIC_LINK_LIBS
+                         arrow
+                         parquet
+                         ${PAIMON_BENCHMARK_STATIC_LINK_LIBS}
+                         Threads::Threads
+                         ${CMAKE_DL_LIBS}
+                         rt
```
zjw1111: why need `rt` here? maybe it's cannot found at MacOS

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -1810,6 +1869,9 @@ if(PAIMON_ENABLE_ORC)
     resolve_dependency(Protobuf)
     resolve_dependency(ORC)
 endif()
+if(PAIMON_BUILD_BENCHMARKS)
+    resolve_dependency(Benchmark FIND_PACKAGE_NAME benchmark)
```
zjw1111: maybe dependency name can change from `Benchmark` to `benchmark`, then no need `FIND_PACKAGE_NAME benchmark`

zjw1111:
I found two issues in the current PR head:

- [P1] `benchmark/CMakeLists.txt` unconditionally links GNU ld flags and `rt` (`-Wl,--whole-archive`, `-Wl,--no-as-needed`, `-Wl,--as-needed`, `rt`). This makes `PAIMON_BUILD_BENCHMARKS=ON` fail on non-Linux/ELF toolchains such as macOS, where those linker flags and `librt` are not available. Please guard these flags/libraries by platform or linker capability.

- [P2] The PK/MOR benchmarks create tables with `bucket=4`, but `MakeRecordBatch()` always calls `SetBucket(0)`, so all benchmark data is written into bucket 0. `FileStoreWrite` only validates that the bucket is in range; it does not re-hash rows by primary key. As a result, the PK/MOR benchmark layout does not represent a normal multi-bucket PK table. Please either distribute rows to buckets by PK hash or make the PK benchmarks explicitly use `bucket=1`.

The earlier checksum, `URL_HASH`, temp-directory cleanup, and help-exit concerns appear to be fixed in the current head.

benchmark/benchmark_suite.cpp
```diff
@@ -0,0 +1,851 @@
...
+
+namespace paimon::benchmark {
+
+namespace {
+
+using DataBatches = std::vector<std::shared_ptr<arrow::StructArray>>;
+
+struct BenchmarkCliOptions {
+    std::string source_parquet;
+    std::string external_table_path;
```
lxy-9602: Could `source_parquet` be renamed to `source_data_file`? I’m wondering whether we intend to support only Parquet as the source format here.

benchmark/benchmark_suite.cpp
```diff
@@ -0,0 +1,851 @@
...
+        argv[write_idx++] = argv[i];
+    }
+
+    *argc = write_idx;
+    argv[write_idx] = nullptr;
+}
+
+bool HasHelpFlagImpl(int argc, char** argv) {
+    for (int i = 1; i < argc; ++i) {
+        const std::string arg(argv[i]);
```
lxy-9602: int -> int32_t

benchmark/benchmark_suite.cpp
```diff
@@ -0,0 +1,851 @@
...
+        << "Example:\n"
+        << "  paimon-read-write-benchmark --paimon_source_parquet /path/data.parquet \\\n"
+        << "      --paimon_file_format parquet --paimon_pk_columns=id \\\n"
+        << "      --paimon_option \"read.batch-size:8192;bucket:4\" --benchmark_filter=BM_Read\n"
+        << std::endl;
+}
+
+struct BenchmarkWorkspace {
+    explicit BenchmarkWorkspace(const std::string& prefix) {
+        std::error_code ec;
```
lxy-9602: Can we use `UniqueTestDirectory`?

benchmark/benchmark_suite.cpp
```diff
@@ -0,0 +1,851 @@
...
+    }
+    return {"", ""};
+}
+
+int64_t GetParquetSourceBatchMaxRows() {
+    const char* value = std::getenv("PAIMON_BENCHMARK_SOURCE_BATCH_MAX_ROWS");
+    if (value == nullptr || std::strlen(value) == 0) {
+        return 4096;
+    }
+
```
lxy-9602, line 260: Consider remove env.

benchmark/benchmark_suite.cpp
```diff
@@ -0,0 +1,851 @@
...
+                                     read_batch_status.ToString());
+        }
+        if (record_batch == nullptr) {
+            break;
+        }
+        if (record_batch->num_rows() <= 0) {
+            continue;
+        }
+        batches.push_back(BuildStructArrayFromRecordBatch(record_batch));
+        total_rows += record_batch->num_rows();
```
lxy-9602: Is all the data kept in memory before being written to Paimon? If so, this could lead to noticeable memory growth. It may also introduce significant memory access overhead during writing, which does not seem very consistent with the actual usage pattern.

benchmark/benchmark_suite.cpp
```diff
@@ -0,0 +1,851 @@
...
+
+    paimon::ReadContextBuilder read_builder(table_path);
+    constexpr int32_t kPrefetchBatchCount = 600;
+    read_builder.SetOptions(options)
+        .EnablePrefetch(true)
+        .SetPrefetchBatchCount(kPrefetchBatchCount)
+        .SetPrefetchMaxParallelNum(prefetch_parallel_num)
+        .EnableMultiThreadRowToBatch(false)
+        .SetRowToBatchThreadNumber(1);
+    auto read_ctx = ValueOrThrow(read_builder.Finish(), "create read context");
```
lxy-9602: It would be better to make `SetRowToBatchThreadNumber` configurable, since in production it is usually set to 3 rather than 1.

benchmark/cli_option_parsing.h
```diff
@@ -0,0 +1,171 @@
...
+    }
+    *value_out = argv[++(*i)];
+    return true;
+}
+
+inline bool ParseCsvOptionArg(int* i, int argc, char** argv, const std::string& arg,
+                              const std::string& option_name,
+                              std::vector<std::string>* columns_out) {
+    std::string parsed_value;
+    if (ConsumeCliOption(arg, option_name, &parsed_value)) {
```
lxy-9602, line 135: Let’s keep output parameters consistently at the end and int->int32_t

benchmark/cli_option_parsing_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    int argc() const {
+        return static_cast<int>(argv.size());
+    }
+};
+
+TEST(CliOptionParsingTest, ConsumeCliOptionWorks) {
+    std::string value;
+    EXPECT_TRUE(paimon::benchmark::ConsumeCliOption("--foo=bar", "--foo", &value));
```
lxy-9602: Please prefer ASSERT* rather than EXPECT*

benchmark/cli_option_parsing_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    EXPECT_EQ(value, "bar");
+
+    value.clear();
+    EXPECT_FALSE(paimon::benchmark::ConsumeCliOption("--foo", "--foo", &value));
+}
+
+TEST(CliOptionParsingTest, ParseCsvColumnsWorks) {
+    const auto parsed = paimon::benchmark::ParseCsvColumns("id, name\tage", "--cols");
```
lxy-9602: Why do we need to use CSV format here? If this is only for separation, would using , as the delimiter be sufficient?

CMakeLists.txt
```diff
@@ -472,3 +489,4 @@ add_subdirectory(src/paimon/global_index/lucene)
 add_subdirectory(src/paimon/testing/mock)
 add_subdirectory(src/paimon/testing/utils)
 add_subdirectory(test/inte)
+add_subdirectory(benchmark)
```
lxy-9602: Please also add a `.rst` for usage.

benchmark/benchmark_suite.cpp
lucasfang: We can add support for the ORC format later if needed. For now,  I think we can keep it simpler.

benchmark/benchmark_suite.cpp
```diff
@@ -0,0 +1,863 @@
...
+        if (record_batch == nullptr) {
+            break;
+        }
+        if (record_batch->num_rows() <= 0) {
+            continue;
+        }
+
+        auto struct_array = BuildStructArrayFromRecordBatch(record_batch);
+        PAIMON_ASSIGN_OR_RAISE(auto batch, MakeRecordBatch(struct_array));
+        PAIMON_RETURN_NOT_OK(AddContext(writer->Write(std::move(batch)), "write batch"));
```
lxy-9602, line 420: Please rm auto in `PAIMON_ASSIGN_OR_RAISE`.

lxy-9602 (approved):
+1

PR #306, opened 2026-05-26
fix(build): make package config relocatable

zjw1111:
### Purpose

Linked issue: none

Fix the generated CMake package config so downstream `find_package(Paimon)` uses
relocatable installed targets instead of hard-coded install-time library paths.

The previous `PaimonConfig.cmake.in` manually declared static/shared imported
targets with `@CMAKE_INSTALL_PREFIX@`, `@CMAKE_INSTALL_LIBDIR@`, and
Linux-specific library filenames. This could expose targets for artifacts that
were not installed by the selected build options, and it could resolve to stale
paths after relocation.

This PR switches the package config to `@PACKAGE_INIT@` and an installed
`PaimonTargets.cmake` export set. The exported target set is populated only by
the targets actually installed through `add_paimon_lib()`.

It also updates the examples CMake file and build-system documentation to link
against exported package targets such as `Arrow::arrow_shared` and
`Paimon::paimon_shared`, instead of relying on global include directories or
unqualified target names.

Finally, it hardens `ci/scripts/build_paimon.sh` by quoting path expansions and
using `mkdir -p` for the build directory.

### Tests

- `git diff --cached --check`
- `git diff --check`
- `git diff --check HEAD~1..HEAD`
- `bash -n ci/scripts/build_paimon.sh`
- `cmake-format --check CMakeLists.txt cmake_modules/BuildUtils.cmake PaimonConfig.cmake.in`
- `cmake-format --check examples/CMakeLists.txt`
- `cmake -S . -B /tmp/paimon-cpp-package-config-check -DPAIMON_BUILD_TESTS=OFF`

### API and Format

No public API or storage format changes.

### Documentation

Updates `docs/source/build_system.rst` to use exported CMake package targets.

### Generative AI tooling

Generated-by: OpenAI Codex


lucasfang (dismissed):
+1

lucasfang (approved):

PR #307, opened 2026-05-27
fix: add ScopeGuard to wait async tasks before early return in OrphanFilesCleanerImpl::Clean()

zjw1111:
### Purpose

Linked issue: N/A

Fix intermittent segmentation fault in `OrphanFilesCleanerImpl::Clean()`. When `GetUsedFiles()` returns an early error (e.g. "do not support clean index manifest"), the function returns immediately while thread pool tasks submitted via `Via(executor_.get(), ...)` are still running. These tasks capture `this` and access members like `fs_`, causing use-after-free when the `OrphanFilesCleanerImpl` object is destroyed shortly after.

Changes:
1. Added a `ScopeGuard` in `Clean()` that calls `CollectAll(file_statuses_futures)` to ensure all submitted async tasks complete before the function returns on early error exit.
2. Added `future.valid()` check in `CollectAll()` (in `future.h`) to safely skip already-consumed futures, so the guard is a no-op on the normal path where `CollectAll` is explicitly called later. This is consistent with the existing `Wait()` function which already has the same `valid()` check.
3. Added a TODO comment in `Via()` noting that the exception capture logic will be removed in the future since paimon-cpp uses Status/Result for error handling throughout.

### Tests

Existing test `OrphanFilesCleanerTest.TestTableWithIndexManifest` covers this path — it was the one experiencing the intermittent segfault.

### API and Format

No.

### Documentation

No.

### Generative AI tooling

Generated-by: Aone Copilot (Claude 4.6 Opus)

### DIFF
--- a/src/paimon/common/executor/future.h
+++ b/src/paimon/common/executor/future.h
@@ -99,7 +99,7 @@ std::vector<T> CollectAll(std::vector<std::future<T>>& futures) {
     std::vector<T> results;
     results.reserve(futures.size());  // Reserve space to avoid reallocation.
     for (auto& future : futures) {
-        results.push_back(future.get());
+        results.push_back(future.get());  // Wait for each future and collect the result.
     }
 
     return results;
--- a/src/paimon/common/executor/future.h
+++ b/src/paimon/common/executor/future.h
@@ -45,6 +45,9 @@ namespace paimon {
 /// execution.
 ///
 /// @note If `func` returns `void`, the returned future is of type `std::future<void>`.
+///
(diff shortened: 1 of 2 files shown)


lxy-9602 (commented):
+1

zjw1111:
@Eyizoha, the exception handling logic (`try/catch + set_exception`) in `Via()` is a legacy pattern. Since paimon-cpp consistently uses `Status`/`Result` for error handling, we plan to remove this exception capture logic from `Via()` in the future. Another reason for this is that calling functions that may throw exceptions inside `ScopeGuard` can result in unexpected crashes.

lucasfang (approved):
+1

PR #308, opened 2026-05-27
fix: fix undefined behavior in bitwise left-shift operations across codebase

lxy-9602:
### Purpose

No Linked issue.

Fix all signed left-shift undefined behavior (UB) across the codebase - signed left-shift that overflows or operates on negative values is UB per the C++ standard.

### Generative AI tooling
Generated-by: Claude-4.6-Opus

### DIFF
--- a/src/paimon/common/utils/var_length_int_utils.h
+++ b/src/paimon/common/utils/var_length_int_utils.h
@@ -88,13 +88,13 @@ class VarLengthIntUtils {
             return static_cast<int32_t>(first_byte);
         }
         // Multi-byte: fall through to generic loop
-        int32_t result = 0;
+        uint32_t result = 0;
         for (int32_t shift = 0; shift < 32; shift += 7) {
             auto byte_val = static_cast<uint8_t>(data[*offset]);
             ++(*offset);
-            result |= static_cast<int32_t>(byte_val & 0x7F) << shift;
+            result |= static_cast<uint32_t>(byte_val & 0x7F) << shift;
             if ((byte_val & 0x80) == 0) {
-                return result;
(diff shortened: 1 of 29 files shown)


src/paimon/common/utils/var_length_int_utils.h
```diff
@@ -88,13 +88,13 @@ class VarLengthIntUtils {
             return static_cast<int32_t>(first_byte);
         }
         // Multi-byte: fall through to generic loop
-        int32_t result = 0;
+        uint32_t result = 0;
         for (int32_t shift = 0; shift < 32; shift += 7) {
             auto byte_val = static_cast<uint8_t>(data[*offset]);
             ++(*offset);
-            result |= static_cast<int32_t>(byte_val & 0x7F) << shift;
+            result |= static_cast<uint32_t>(byte_val & 0x7F) << shift;
```
zjw1111, line 97: This avoids the signed left-shift UB, but the decoder now accepts unsigned values that are outside the signed non-negative range produced by the encoder. For example, a malformed varint32 such as `ff ff ff ff 0f` is assembled as `0xffffffff` and then `static_cast<int32_t>` returns `-1`, even though `EncodeInt` rejects negative values. `DecodeLong` has the same issue for inputs that require the 10th payload bit and cast back to a negative `int64_t`. Please reject values greater than `INT32_MAX` / `INT64_MAX` before casting back, or validate the final payload byte so these malformed encodings return an error instead of a valid negative length/position.

zjw1111 (commented):
I found one correctness issue in the varint decoder overflow handling.

lszskye (approved):

PR #309, opened 2026-05-27
feat: Add manifests and files system tables

suxiaogang223:
## Background

This PR is part of #141 and continues the system table work after the previously merged `options`, `audit_log`, `binlog`, and pure metadata system table support.

The scope of this PR is manifest/file-backed table-level metadata system tables. It adds read-only support for `table$manifests` and `table$files`, so users can inspect latest snapshot manifest and data file metadata through the existing system table query path.

## Architecture

- Keep `manifests` and `files` on the existing `InMemorySystemTable` execution path.
  - Both tables materialize latest-snapshot metadata into `GenericRow` values.
  - The existing in-memory system table scan/read path handles singleton split planning and Arrow export.

- Split system table implementations into Java-aligned per-table files.
  - `OptionsSystemTable`, `SnapshotsSystemTable`, `SchemasSystemTable`, `TagsSystemTable`, `BranchesSystemTable`, `ConsumersSystemTable`, `ManifestsSystemTable`, and `FilesSystemTable` now live in separate source/header pairs.
  - The previous `metadata_system_tables.{h,cpp}` aggregation file is removed.

- Introduce `SystemTableUtils` for shared system table helpers.
  - It owns shared context creation, option merging, branch loading, timestamp/value conversion, JSON conversion, manifest reading, data file merging, and metadata formatting helpers.
  - This keeps concrete system table classes focused on schema declaration and row materialization.

- Keep `SystemTableLoader` registry behavior unchanged.
  - The loader still creates system tables through the registry/factory table.
  - It now includes concrete system table headers directly after the file split.

## System Tables Added

This PR follows the Apache Paimon table-scoped system table model, where table metadata is queried with names like `table$files`. Reference semantics: https://paimon.apache.org/docs/master/concepts/system-tables/

- `table$manifests`
  - Purpose: exposes data manifest metadata for the latest snapshot.
  - Source: latest snapshot data manifests through `SnapshotManager`, `ManifestList`, and `ManifestFileMeta`.
  - Columns: `file_name`, `file_size`, `num_added_files`, `num_deleted_files`, `schema_id`, `min_partition_stats`, `max_partition_stats`, `min_row_id`, `max_row_id`.
  - Note: reads data manifests only in this PR.

- `table$files`
  - Purpose: exposes data file metadata for the latest snapshot.
  - Source: latest snapshot data manifest entries merged with `FileEntry::MergeEntries`.
  - Columns: `partition`, `bucket`, `file_path`, `file_format`, `schema_id`, `level`, `record_count`, `file_size_in_bytes`, `min_key`, `max_key`, `null_value_counts`, `min_value_stats`, `max_value_stats`, `min_sequence_number`, `max_sequence_number`, `creation_time`, `deleteRowCount`, `file_source`, `first_row_id`, `write_cols`.
  - Note: `file_path` uses external path when present, otherwise the table bucket path plus file name.

## Scope Notes

- This PR reads the latest snapshot only; `scan.snapshot-id`, tag, and timestamp time-travel options are left for follow-up work.
- Predicate pushdown for manifest/file metadata is not implemented in this PR.
- Existing `options`, `audit_log`, `binlog`, `snapshots`, `schemas`, `tags`, `branches`, and `consumers` behavior is preserved.
- Remaining metadata system tables such as `partitions`, `buckets`, `statistics`, `table_indexes`, and `aggregation_fields` are planned for later PRs.

## Tests

- Local: `pre-commit run --all-files`
- Fedora: `cmake --build build-codex-metadata-pr3 --target paimon-core-test paimon-read-inte-test -j2`
- Fedora: `./build-codex-metadata-pr3/debug/paimon-core-test --gtest_filter=FileSystemCatalogTest.TestMetadataSystemTableCatalog`
- Fedora: `./build-codex-metadata-pr3/debug/paimon-read-inte-test --gtest_filter=SystemTableReadInteTest.TestReadMetadataSystemTables:SystemTableReadInteTest.TestReadFilesSystemTableForPartitionedTable:SystemTableReadInteTest.TestReadManifestAndFilesSystemTablesForEmptyTable`


src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -140,6 +166,12 @@ Result<VariantType> LocalTimestampMillisValue(int64_t epoch_millis) {
     return TimestampMillisValue(local_timestamp.GetMillisecond());
 }
 
+Result<VariantType> LocalTimestampMillisValue(const Timestamp& local_timestamp) {
+    PAIMON_ASSIGN_OR_RAISE(Timestamp utc_timestamp, DateTimeUtils::ToUTCTimestamp(local_timestamp));
+    int64_t epoch_millis = utc_timestamp.GetMillisecond();
+    return LocalTimestampMillisValue(epoch_millis);
+}
```
lxy-9602: I’m a bit curious about `LocalTimestampMillisValue`: why does it first convert local time to UTC and then convert it back to local time? Is there any difference compared with reading the value directly from the local timestamp?

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -150,12 +182,215 @@ VariantType OptionalTimestampMillisValue(const std::optional<int64_t>& value) {
...
+Result<std::vector<ManifestFileMeta>> ReadDataManifests(
+    const MetadataSystemTableContext& context, const Snapshot& snapshot,
+    const std::shared_ptr<FileStorePathFactory>& path_factory, const CoreOptions& core_options,
+    const std::shared_ptr<MemoryPool>& pool) {
+    PAIMON_ASSIGN_OR_RAISE(
+        std::unique_ptr<ManifestList> manifest_list,
+        ManifestList::Create(context.fs, core_options.GetManifestFormat(),
+                             core_options.GetManifestCompression(), path_factory, pool));
+    std::vector<ManifestFileMeta> manifests;
+    PAIMON_RETURN_NOT_OK(manifest_list->ReadDataManifests(snapshot, &manifests));
```
lxy-9602, line 237: It looks like the Java implementation calls `ReadAllManifests`, which also supports changelog manifests. If paimon-cpp does not plan to support that for now, could you please leave a TODO in the code.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -150,12 +182,215 @@ VariantType OptionalTimestampMillisValue(const std::optional<int64_t>& value) {
...
+    std::vector<ManifestEntry> merged_entries;
+    PAIMON_RETURN_NOT_OK(FileEntry::MergeEntries(entries, &merged_entries));
+    return merged_entries;
+}
+
+std::optional<std::string> OptionalBinaryRowString(const BinaryRow& row) {
+    if (row.GetFieldCount() <= 0) {
+        return std::nullopt;
+    }
+    return row.ToString();
```
lxy-9602: `BinaryRow::ToString()` returns a hash value, but my understanding is that min/max keys are expected to use the actual field data (`BinaryRow@0x12ab34cd` -> `[1, 100]`). It may need something like FieldsValueMapString() instead. Could you please adjust the corresponding logic?

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -150,12 +182,215 @@ VariantType OptionalTimestampMillisValue(const std::optional<int64_t>& value) {
...
+    const BinaryRow& row, const std::shared_ptr<arrow::Schema>& partition_schema) {
+    PAIMON_ASSIGN_OR_RAISE(std::optional<std::string> value,
+                           OptionalPartitionString(row, partition_schema));
+    return OptionalStringValue(value);
+}
+
+Result<std::string> PartitionString(const std::shared_ptr<FileStorePathFactory>& path_factory,
+                                    const BinaryRow& partition) {
+    PAIMON_ASSIGN_OR_RAISE(std::string value, path_factory->GetPartitionString(partition));
+    return value;
```
lxy-9602: The partition field seems to have a little formatting issue. The current C++ result is `dt=20260527`, while the actual Java result is `{20260527}`. You may want to refer to the test in `FilesTableTest.java`.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -150,12 +182,215 @@ VariantType OptionalTimestampMillisValue(const std::optional<int64_t>& value) {
...
+    PAIMON_ASSIGN_OR_RAISE(std::vector<InternalRow::FieldGetterFunc> getters,
+                           InternalRowUtils::CreateFieldGetters(schema, /*use_view=*/false));
+    std::vector<std::string> values;
+    values.reserve(fields.size());
+    for (size_t i = 0; i < fields.size(); ++i) {
+        std::string value = "null";
+        if (!row.IsNullAt(i)) {
+            VariantType field_value = getters[i](row);
+            if (std::holds_alternative<std::string_view>(field_value)) {
+                value = std::string(std::get<std::string_view>(field_value));
```
lxy-9602: I noticed that `DataDefine::VariantValueToString` supports `string_view`. I wonder why do we need to handle `string_view` as a special case here?

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -150,12 +182,215 @@ VariantType OptionalTimestampMillisValue(const std::optional<int64_t>& value) {
...
+            } else {
+                value = DataDefine::VariantValueToString(field_value);
+            }
+        }
+        values.emplace_back(fmt::format("{}:{}", fields[i].Name(), value));
```
lxy-9602: I’d like to confirm whether the separator between the key and value should be `:` or `=` in java.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -424,4 +659,189 @@ Result<std::vector<GenericRow>> ConsumersSystemTable::BuildRows() const {
...
+        if (!(entry.Kind() == FileKind::Add())) {
+            continue;
+        }
+
+        const std::shared_ptr<DataFileMeta>& file = entry.File();
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<TableSchema> data_schema,
+                               LoadDataSchema(context_, file->schema_id));
+        PAIMON_ASSIGN_OR_RAISE(std::vector<DataField> value_stats_fields,
+                               ValueStatsFields(context_, file->schema_id));
+        std::shared_ptr<SimpleStatsEvolution> stats_evolution =
```
lxy-9602: `value_stats_fields` uses the field types corresponding to the `schema_id` of the data file, while stats is based on the latest evolved schema, so the two do not match. My understanding is that `value_stats_fields` needs to be converted to the latest schema as well. Also, could you please add a test covering schema evolution?

test/inte/read_inte_test.cpp
```diff
@@ -699,6 +699,151 @@ TEST(SystemTableReadInteTest, TestReadMetadataSystemTables) {
     auto branch_create_time_array =
         std::dynamic_pointer_cast<arrow::TimestampArray>(branches_array->field(1));
     ASSERT_TRUE(branch_create_time_array);
+
```
lxy-9602, line 702: Could you please add a test based on the existing `test_data` and verify the full Arrow array returned by the system table? This would avoid the issue where temporarily constructed data makes the manifest file names unpredictable. I think having the complete array output would be more intuitive, and it could also cover the schema evolution scenario. Thanks!

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -791,10 +831,16 @@ Result<std::vector<GenericRow>> FilesSystemTable::BuildRows() const {
...
-        PAIMON_ASSIGN_OR_RAISE(std::vector<DataField> value_stats_fields,
-                               ValueStatsFields(context_, file->schema_id));
-        std::shared_ptr<SimpleStatsEvolution> stats_evolution =
-            stats_evolutions.GetOrCreate(data_schema);
+        PAIMON_ASSIGN_OR_RAISE(std::vector<DataField> data_stats_fields,
+                               ProjectWriteFields(data_schema, *file));
+        PAIMON_ASSIGN_OR_RAISE(std::vector<DataField> key_fields,
+                               data_schema->TrimmedPrimaryKeyFields());
+        if (key_fields.empty()) {
+            key_fields = data_schema->Fields();
```
lxy-9602: When `key_fields` is empty, why do we repopulate it with `data_schema->fields()`? Is there any problem with preserving the original empty `key_fields`?

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -373,11 +397,29 @@ Result<std::shared_ptr<TableSchema>> LoadDataSchema(const MetadataSystemTableCon
...
+                                                  const DataFileMeta& file) {
+    if (!file.write_cols) {
+        return data_schema->Fields();
+    }
+
+    std::vector<DataField> fields;
+    fields.reserve(file.write_cols->size() + data_schema->PartitionKeys().size());
+    for (const auto& write_col : file.write_cols.value()) {
+        if (write_col == SpecialFields::RowId().Name() ||
+            write_col == SpecialFields::SequenceNumber().Name()) {
```
lxy-9602: Would it be simpler to use `SpecialFields::IsSpecialFieldName` here?

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -373,11 +397,29 @@ Result<std::shared_ptr<TableSchema>> LoadDataSchema(const MetadataSystemTableCon
...
+        }
+        PAIMON_ASSIGN_OR_RAISE(DataField field, data_schema->GetField(write_col));
+        fields.push_back(std::move(field));
+    }
+
+    for (const auto& partition_key : data_schema->PartitionKeys()) {
+        if (!ObjectUtils::Contains(file.write_cols.value(), partition_key)) {
+            PAIMON_ASSIGN_OR_RAISE(DataField field, data_schema->GetField(partition_key));
+            fields.push_back(std::move(field));
+        }
```
lxy-9602, line 440: I’m a bit confused why partition fields need to be handled separately in addition to `write_col`.

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -150,12 +177,262 @@ VariantType OptionalTimestampMillisValue(const std::optional<int64_t>& value) {
...
+}
+
+Result<std::optional<std::string>> OptionalPartitionString(
+    const BinaryRow& row, const std::shared_ptr<arrow::Schema>& partition_schema) {
+    if (row.GetFieldCount() <= 0) {
+        return std::optional<std::string>();
+    }
+    PAIMON_ASSIGN_OR_RAISE(std::string value,
+                           BinaryRowPartitionComputer::PartToSimpleString(
+                               partition_schema, row, ",", kMaxPartitionStatsLength));
```
lszskye: It looks like Java Paimon uses CastExecutor to convert partition values to strings. We hope cpp Paimon can also use the CastExecutor class; otherwise, there will be inconsistencies in the conversion of certain special types, such as the DATE type.

lxy-9602 (approved):
+1

PR #310, opened 2026-05-28
feat(lumina): support null values in vector column during index building

lxy-9602:
### Purpose

Support writing null vector values in LuminaGlobalIndex. Previously, any null in the vector column would cause AddBatch to fail. Now, null rows are automatically skipped during index building — their row IDs are not indexed, and they will never be recalled by vector search.

- `AddBatch`: splits contiguous non-null rows into segments, skips null rows
- `Finish`: returns empty metas when all rows are null
- `LuminaDataset`: uses per-segment start IDs to generate correct non-contiguous vector IDs

Linked issue #5 

### Tests
LuminaGlobalIndexTest. TestWriteWithNullRows
LuminaGlobalIndexTest.TestWriteWithMultipleNullSegments
LuminaGlobalIndexTest.TestWriteWithAllNullRows
LuminaGlobalIndexTest.TestWriteWithNullAndFilter
GlobalIndexTest .TestWriteCommitScanReadIndexWithScore

### Generative AI tooling
Generated-by: Claude-4.6-Opus

### DIFF
--- a/src/paimon/global_index/lumina/lumina_global_index.cpp
+++ b/src/paimon/global_index/lumina/lumina_global_index.cpp
@@ -177,8 +177,12 @@ Result<std::shared_ptr<GlobalIndexReader>> LuminaGlobalIndex::CreateReader(
 class LuminaDataset : public ::lumina::api::Dataset {
  public:
     LuminaDataset(int64_t element_count, uint32_t dimension,
-                  const std::vector<std::shared_ptr<arrow::FloatArray>>& array_vec)
-        : element_count_(element_count), dimension_(dimension), array_vec_(array_vec) {}
+                  const std::vector<std::shared_ptr<arrow::FloatArray>>& array_vec,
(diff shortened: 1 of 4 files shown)


jerry-024:
Nice work! The null-skipping approach is clean and compatible with the Java implementation.

**One suggestion for test coverage:**

Consider adding a test that exercises multiple `AddBatch` calls where **both batches contain null rows**. This would verify that `count_` accumulation across batches produces correct global vector IDs. For example:

- Batch 1: `[vec0, null, vec2]` → expected IDs: {0, 2}
- Batch 2: `[null, vec4, vec5]` → expected IDs: {4, 5} (since count_=3 after batch 1)

The current tests cover single-batch null patterns well, but cross-batch ID correctness is the most likely place a subtle regression could appear in future refactoring.

jerry-024 (approved):
+1

lszskye (approved):

PR #314, opened 2026-05-28
chore: modify comment in RecordBatch & add check in GenericRow

lszskye:
### Purpose
modify comment in RecordBatch & add check in GenericRow


lxy-9602 (approved):
+1

PR #315, opened 2026-05-28
fix: date validation reject invalid dates instead of silent normalization

dalingmeng:
### Purpose

 no linked issue:

`std::get_time` only validates field ranges (e.g., `tm_mday` in [1, 31]) but does not check actual date validity (e.g., Feb 29 in non-leap year, Apr 31). After parsing, `timegm`/`mktime` silently normalizes these invalid dates instead of returning an error:

- `StringToDate("2019-02-29")` → returns 17956 (2019-03-01) instead of error
- `StringToTimestampMillis("2023-02-30 00:00:00")` → returns 1677686400000 (2023-03-02) instead of error

This change adds post-normalization validation: after calling `timegm`/`mktime`, we compare `tm_mday` and `tm_mon` with the original values. If they differ, the input date was invalid and an error is returned.

This aligns with Java Paimon behavior, where `LocalDate.parse("2019-02-29")` throws `DateTimeParseException`.

### Tests

- `StringUtilsTest.TestStringToDateWithInvalidLeapYear` — verifies Feb 29 in non-leap year, Feb 30, Apr 31 are rejected; valid leap year date (2020-02-29) succeeds
- `StringUtilsTest.TestStringToTimestampMillisWithInvalidDate` — verifies Feb 30, Feb 29 in non-leap year, Apr 31 are rejected; valid leap year timestamp (2024-02-29) succeeds

### API and Format

No API or format changes.

### Documentation

No new features introduced.


src/paimon/common/utils/string_utils.cpp
```diff
@@ -140,8 +140,10 @@ Result<int32_t> StringUtils::StringToDate(const std::string& str) {
     if (ss.fail()) {
         return Status::Invalid(fmt::format("failed to convert string '{}' to date", str));
     }
+    int orig_mon = timeinfo.tm_mon;
+    int orig_mday = timeinfo.tm_mday;
     std::time_t time = timegm(&timeinfo);
```
lxy-9602, line 145: int -> int32_t

lxy-9602 (approved):
+1

PR #319, opened 2026-05-29
fix: fix ub in Blob

lszskye:
### Purpose
fix ub in Blob

### Generative AI tooling
Generated-by: Claude-4.7-Opus


lxy-9602 (dismissed):
+1

lxy-9602 (approved):
+1

PR #320, opened 2026-05-29
fix: Like::TestString to align with Java LIKE semantics

lxy-9602:
### Purpose

No Linked issue.

**Description:**

- Validate LIKE escape sequences like Java Paimon
- Treat trailing `\` and invalid escapes as errors
- Make `_` match one UTF-8 character instead of one byte
- Keep `%` matching any sequence
- Remove `alloca` and use `std::vector<bool>` instead

### Tests

- Add invalid escape sequence tests
- Add escaped `\` and `%` tests
- Add UTF-8 multi-byte `_` matching tests
- Add line terminator semantics tests

### Generative AI tooling
Aone Copilot (Claude)


src/paimon/common/predicate/like.cpp
```diff
@@ -16,79 +16,143 @@
...
+    std::vector<std::string> pat_chars;  // each element is a literal string segment or wildcard
     std::vector<bool> is_wild;
-    for (size_t i = 0; i < pattern.size(); ++i) {
-        if (pattern[i] == '\\' && i + 1 < pattern.size()) {
-            pat.push_back(pattern[i + 1]);
+
+    for (size_t i = 0; i < pattern.size();) {
+        if (pattern[i] == '\\') {
+            if (i + 1 >= pattern.size()) {
+                return Status::Invalid("Invalid escape sequence '" + pattern + "', " +
```
zjw1111: use `fmt::format` for message output

zjw1111 (approved):
+1

PR #323, opened 2026-06-01
fix: fix thread-safe problem for file store path factory

lucasfang:
### Purpose

N/A

This change fixes a thread-safety issue in `FileStorePathFactory` cache access paths.

- Protect cache reads/writes in `ToBinaryRow` and `GetPartitionString` with `std::shared_mutex`.
- Use read lock for fast-path cache hits and write lock for insertions, with double-check under write lock.
- Update API note to reflect that `GetPartitionString` is now thread-safe for concurrent calls.

### Tests

UT:
- `FileStorePathFactoryTest.TestConcurrentToBinaryRowAndGetPartitionString`
  - Runs 8 threads with 500 iterations per thread.
  - Concurrently calls `ToBinaryRow` and `GetPartitionString` and validates deterministic outputs.

Existing UTs in the same suite still pass for non-concurrent paths.

IT:
- N/A

### API and Format

No API behavior change for external callers, and no storage format/protocol change.

Internal update only:
- Added `cache_mutex_` (`std::shared_mutex`) to guard in-memory caches.
- Updated method comment for thread-safety semantics.

### Documentation

No new feature introduced. No documentation update required.

### Generative AI tooling


lxy-9602 (commented):
+1

zjw1111 (approved):
+1

PR #328, opened 2026-06-02
fix: skip object store check when UseRESTCatalogCommit is enabled

liujiayi771:
### Purpose

When `UseRESTCatalogCommit(true)` is set, `FileStoreCommit` delegates the final snapshot commit to the REST catalog server instead of writing the snapshot file to the filesystem. The existing object store restriction (which blocks commit on `oss://`, `s3://`, etc.) exists because atomic rename is not supported on object stores — but this restriction does not apply in REST commit mode since no snapshot file needs to be atomically written by the client.

Currently, external catalog integrations that use `UseRESTCatalogCommit(true)` with object store table paths are forced to set the internal test flag `enable-object-store-commit-in-inte-test` to bypass this check. This PR relaxes the constraint so that `UseRESTCatalogCommit(true)` alone is sufficient.

### Changes

- `file_store_commit.cpp`: Added `!ctx->UseRESTCatalogCommit()` to the object store check condition, so REST commit mode skips the restriction.

### Tests

- `TestObjectStoreBlockedWithoutRESTCommit`: Verifies that non-REST commit mode still correctly identifies object store vs. local paths via `FileSystem::IsObjectStore()`.
- `TestRESTCatalogCommitWithObjectStorePath`: Verifies that REST commit mode works end-to-end (Create → Commit → GetLastCommitTableRequest) and does not write snapshot files to the filesystem.

All 30 existing `FileStoreCommitImplTest` tests continue to pass.

### DIFF
--- a/src/paimon/core/operation/file_store_commit.cpp
+++ b/src/paimon/core/operation/file_store_commit.cpp
@@ -81,7 +81,8 @@ Result<std::unique_ptr<FileStoreCommit>> FileStoreCommit::Create(
     assert(options.GetFileSystem());
     assert(options.GetFileFormat());
     PAIMON_ASSIGN_OR_RAISE(bool is_object_store, FileSystem::IsObjectStore(root_path));
-    if (is_object_store && opts.find("enable-object-store-commit-in-inte-test") == opts.end()) {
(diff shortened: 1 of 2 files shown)


lxy-9602 (approved):
+1

PR #329, opened 2026-06-02
fix: avoid signed integer overflow UB in BloomFilter and DeltaVarintCompressor

lxy-9602:
### Purpose

No Linked issue.
Fixes signed integer overflow undefined behavior (UB) in utility classes (`bloom_filter.cpp`, `bloom_filter64.cpp`, `delta_varint_compressor.cpp` ). Both issues stem from porting Java integer arithmetic (which wraps mod 2^N) directly to C++ signed types (where signed overflow is UB and may fail under UBSan or compiler optimization).


zjw1111 (approved):
+1

PR #333, opened 2026-06-03
feat: remove support for FLOAT/DOUBLE type partition field

lszskye:
### Purpose
remove support for FLOAT/DOUBLE type partition field

### DIFF
--- a/src/paimon/common/utils/binary_row_partition_computer_test.cpp
+++ b/src/paimon/common/utils/binary_row_partition_computer_test.cpp
@@ -43,16 +43,14 @@ TEST(BinaryRowPartitionComputerTest, TestToAndFromBinaryRow) {
                                  arrow::field("f6", arrow::int32()),
                                  arrow::field("f7", arrow::int64()),
                                  arrow::field("f8", arrow::int64()),
-                                 arrow::field("f9", arrow::float32()),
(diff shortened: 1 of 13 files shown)


lucasfang (dismissed):
+1

zjw1111 (dismissed):
Requesting changes for one scope issue:

- `src/paimon/testing/utils/data_generator.cpp`: removing `FLOAT` / `DOUBLE` support from `DataGenerator::WriteBinaryRow` also affects bucket-row construction, not only partition-row construction. `SplitArrayByPartitionAndBucket()` uses the same `ExtractPartialRow(binary_row, bucket_fields)` path to build `bucket_row` for bucket id calculation. The schema validation added in this PR only rejects `FLOAT` / `DOUBLE` partition keys, while `FLOAT` / `DOUBLE` bucket keys are still allowed and the production `BucketIdCalculator::WriteBucketRow` still supports them. As a result, a legal table with a `FLOAT` / `DOUBLE` bucket key that is not a partition key can become unsupported in the test data generation path. Please keep the `FLOAT` / `DOUBLE` branches in `DataGenerator::WriteBinaryRow`, or split the helper so the new restriction applies only to the partition converter / schema validation path.

lucasfang (approved):
+1

PR #337, opened 2026-06-03
Allow FileStoreCommit for PK tables with postpone bucket mode

liujiayi771:
## Summary

- Allow `FileStoreCommit::Create` to proceed for PK tables when using postpone bucket mode (`bucket=-2`)
- In postpone bucket mode, all records are first stored in the `bucket-postpone` directory of each partition and are not available to readers. To move the records into the correct bucket and make them readable, a compaction job needs to be run (via `compact` procedure or triggered by the catalog server)
- The commit logic (manifest and snapshot generation) is identical to append tables, so there is no reason to block it
- PK tables with fixed bucket (`bucket > 0`) remain blocked as before

## Test plan

- [x] `TestPostponeBucketPKTableCommitAllowed`: PK table with `bucket=-2` → `FileStoreCommit::Create` succeeds without workaround flag
- [x] `TestFixedBucketPKTableCommitRejected`: PK table with `bucket=4` → `FileStoreCommit::Create` returns `NotImplemented`

## Reference

- [Paimon Postpone Bucket Documentation](https://paimon.apache.org/docs/master/primary-key-table/data-distribution/#postpone-bucket)

### DIFF
--- a/src/paimon/core/operation/file_store_commit.cpp
+++ b/src/paimon/core/operation/file_store_commit.cpp
@@ -69,15 +69,10 @@ Result<std::unique_ptr<FileStoreCommit>> FileStoreCommit::Create(
     const auto& schema = table_schema.value();
     if (!schema->PrimaryKeys().empty() &&
         ctx->GetOptions().find("enable-pk-commit-in-inte-test") == ctx->GetOptions().end()) {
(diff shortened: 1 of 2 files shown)


lucasfang (approved):
+1

PR #338, opened 2026-06-04
fix: remap predicate field index to projected read schema in InternalReadContext

duanyyyyyyy:
### Purpose

  `InternalReadContext::Create` performs a strict field-id check that compares
  the predicate's `FieldIndex` against the field's position in `read_schema`.
  Predicates are constructed upstream (Java SDK / FE) against the *latest table
  schema*, so when the query projects a subset of columns the leaf's
  `FieldIndex` no longer matches the column's position in the projected
  `read_schema` and `TableRead::Create` fails:

      Paimon TableRead::Create error: Invalid: field obs_index has field
      idx 0 in input schema, mismatch field idx 1 in predicate

  The same mismatch is also unsafe downstream: `LeafPredicateImpl::Test` uses
  `field_index_` directly to index into the arrow array / `InternalRow` built
  from `read_schema`, so even with the validation relaxed the predicate would
  silently read the wrong column.

  This is the C++ analogue of paimon Java's `AbstractDataTableRead.executeFilter`
  calling `PredicateProjectionConverter.fromProjection` before applying
  row-level filters; we do it once at context construction so every downstream
  reader sees a predicate already aligned with `read_schema`.

  Walks the predicate tree once at `InternalReadContext::Create` and rebuilds
  each `LeafPredicate` with its `field_index` resolved from `read_schema` by
  field name. `CompoundPredicate` is reconstructed only when any descendant
  actually changed, otherwise the original `shared_ptr` is reused. The
  remapped predicate is stored on `InternalReadContext`; `GetPredicate()`
  returns it (the original FE predicate stays accessible via the wrapped
  `ReadContext` if ever needed). `CreateWithSchema` remaps against its
  `new_read_schema` for the same reason — the previously-remapped predicate
  is aligned with the *original* read schema, not the new one.

  Reproduces on a DE table with btree global index, query pattern:

      SELECT clip_id, obs_index, ...
      FROM t
      WHERE collected_date = ... AND clip_id = ... AND obs_index BETWEEN 0 AND 10;                                                                                                            

  ### Tests

  `InternalReadContext` gains four cases (in `internal_read_context_test.cpp`):

  - `TestPredicateFieldIdxRemappedWhenProjected` — projected `read_schema`
    has the predicate's field at a different position; leaf's `FieldIndex`
    is rewritten to the projected position.
  - `TestPredicateUnchangedWhenAligned` — when already aligned, the
    returned `shared_ptr` is the input pointer (zero-copy reuse).
  - `TestCompoundPredicateRemap` — every leaf inside a `CompoundPredicate`
    is recursively rewritten.
  - `TestPredicateOnFieldMissingFromReadSchema` — a leaf naming a field
    not in `read_schema` surfaces as `Status::Invalid`.

  ### API and Format
  
  No public-include / storage-format / protocol change. The private
  constructor of `InternalReadContext` gains a `remapped_predicate`
  parameter; `GetPredicate()`'s return value now reflects the
  read-schema-aligned indices rather than forwarding to the wrapped
  `ReadContext`. Behavior change is correctness-only (previously failing
  paths now succeed; previously succeeding paths return identical results
  since the remap is a no-op when already aligned).

  ### Documentation
  
  No new user-facing feature; no documentation changes.

  ### Generative AI tooling
  
  Generated-by: Claude Code (claude-opus-4-7)

### DIFF
--- a/src/paimon/core/operation/internal_read_context.cpp
+++ b/src/paimon/core/operation/internal_read_context.cpp
@@ -16,12 +16,16 @@
 
 #include "paimon/core/operation/internal_read_context.h"
 
+#include <optional>
 #include <utility>
 
+#include "paimon/common/predicate/compound_predicate_impl.h"
+#include "paimon/common/predicate/leaf_predicate_impl.h"
 #include "paimon/common/predicate/predicate_validator.h"
 #include "paimon/common/table/special_fields.h"
 #include "paimon/common/types/data_field.h"
 #include "paimon/core/schema/arrow_schema_validator.h"
+#include "paimon/predicate/function.h"
 #include "paimon/status.h"
 
 namespace arrow {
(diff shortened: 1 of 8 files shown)


src/paimon/core/operation/internal_read_context.h
```diff
@@ -62,8 +62,13 @@ class InternalReadContext {
...
+    // Returns the predicate with each leaf's field index already remapped to its position
+    // in `read_schema_`. Upstream constructs predicates against the latest table schema,
+    // so when the query projects a subset of columns the original indices no longer
+    // match the projected schema. The remapping is done once at context construction; all
+    // subsequent reader code can use this index directly.
     const std::shared_ptr<Predicate>& GetPredicate() const {
-        return read_context_->GetPredicate();
+        return remapped_predicate_;
     }
     bool EnablePredicateFilter() const {
```
lxy-9602, line 76: The class comment in `predicate_builder.h` clearly states that field_index is “the index of the field in the read schema (0-based)”. If we need to change its behavior, could you please also review the comments carefully and update them all accordingly?

lxy-9602: Is the current change based on the assumption that the `field_idx` provided in the user predicate is not meaningful, since it will be rebuilt in `InternalReadContext` anyway?

src/paimon/core/operation/internal_read_context.cpp
```diff
@@ -29,6 +31,50 @@ class Schema;
...
+        }
+        return std::static_pointer_cast<Predicate>(leaf->NewLeafPredicate(new_index));
+    }
+    if (auto compound = std::dynamic_pointer_cast<CompoundPredicateImpl>(predicate)) {
+        std::vector<std::shared_ptr<Predicate>> remapped_children;
+        remapped_children.reserve(compound->Children().size());
+        bool any_changed = false;
+        for (const auto& child : compound->Children()) {
+            PAIMON_ASSIGN_OR_RAISE(auto remapped, RemapPredicateFieldIndex(read_schema, child));
+            if (remapped != child) {
```
lxy-9602: In production code, could we please make the type explicit in `PAIMON_ASSIGN_OR_RAISE` to improve readability?

src/paimon/core/operation/internal_read_context_test.cpp
```diff
@@ -191,4 +194,99 @@ TEST(InternalReadContext, TestReadWithFieldIdsAndSpecialFields) {
...
+    ASSERT_OK_AND_ASSIGN(auto table_schema, schema_manager.ReadSchema(0));
+    ASSERT_OK_AND_ASSIGN(auto internal_context,
+                         InternalReadContext::Create(std::move(read_context), table_schema,
+                                                     table_schema->Options()));
+    auto leaf = std::dynamic_pointer_cast<LeafPredicate>(internal_context->GetPredicate());
+    ASSERT_NE(leaf, nullptr);
+    EXPECT_EQ(leaf->FieldName(), "f3");
+    // f3 is the first column in the projected read schema.
+    EXPECT_EQ(leaf->FieldIndex(), 0);
+}
```
lxy-9602, line 223: Please prefer `ASSERT_*` over `EXPECT_*` here.

src/paimon/core/table/source/table_read_test.cpp
```diff
@@ -67,18 +67,6 @@ TEST(TableReadTest, TestReadWithInvalidContext) {
...
-                                                 FieldType::DOUBLE, Literal(15.0));
-        ReadContextBuilder context_builder(path);
-        context_builder.SetReadSchema({"f3", "f0", "f1"});
-        context_builder.SetPredicate(predicate);
-        ASSERT_OK_AND_ASSIGN(auto read_context, context_builder.Finish());
-        ASSERT_NOK_WITH_MSG(
-            TableRead::Create(std::move(read_context)),
-            "field f3 has field idx 0 in input schema, mismatch field idx 2 in predicate");
-    }
     {
```
lxy-9602, line 70: Could you please add an integration test, or alternatively modify field_idx so it no longer matches the read schema, to verify end-to-end correctness? Thanks!

src/paimon/core/operation/internal_read_context.cpp
```diff
@@ -29,6 +31,50 @@ class Schema;
...
+    if (auto leaf = std::dynamic_pointer_cast<LeafPredicateImpl>(predicate)) {
+        int32_t new_index = read_schema.GetFieldIndex(leaf->FieldName());
+        if (new_index == -1) {
+            return Status::Invalid(
+                fmt::format("field {} does not exist in schema", leaf->FieldName()));
+        }
+        if (new_index == leaf->FieldIndex()) {
+            return predicate;
+        }
+        return std::static_pointer_cast<Predicate>(leaf->NewLeafPredicate(new_index));
```
lszskye: why std::static_pointer_cast here

src/paimon/core/operation/internal_read_context.cpp
```diff
@@ -29,6 +31,50 @@ class Schema;
...
+            }
+            remapped_children.push_back(std::move(remapped));
+        }
+        if (!any_changed) {
+            return predicate;
+        }
+        return std::static_pointer_cast<Predicate>(
+            compound->NewCompoundPredicate(remapped_children));
+    }
+    return predicate;
```
lszskye: if neither LeafPredicateImpl or CompoundPredicateImpl, should return Invalid state

PR #340, opened 2026-06-05
fix: Avoid signed overflow UB in BSI index and null-deref in bloom filter index

lxy-9602:
### Purpose

No Linked issue.

Fix two undefined-behavior bugs in file index readers:

1. **BSI reader signed overflow**: negative branches computed `-value`, which is
   UB when value is `INT64_MIN`. Now use a `SafeAbs()` helper that clamps
   `INT64_MIN` to `INT64_MIN`.

2. **Bloom filter null deref**: `VisitEqual` hashed the literal before checking
   `IsNull()`, dereferencing a null value. Now returns `Remain()`
   for null literals before hashing.

Cross-checked with Java `BitSliceIndexBitmapFileIndex`: Java's `Math.abs` wraps
`Long.MIN_VALUE` back to negative, so its writer rejects `MIN_VALUE` outright —
the value can never reach a BSI index. The C++ fix is defensive and matches Java
results for all valid values.

### Generative AI tooling
Generated-by: Aone Copilot (Claude)

### DIFF
--- a/src/paimon/common/file_index/bloomfilter/bloom_filter_file_index.cpp
+++ b/src/paimon/common/file_index/bloomfilter/bloom_filter_file_index.cpp
@@ -78,9 +78,15 @@ BloomFilterFileIndexReader::BloomFilterFileIndexReader(const FastHash::HashFunct
 
 Result<std::shared_ptr<FileIndexResult>> BloomFilterFileIndexReader::VisitEqual(
     const Literal& literal) {
+    // This returns `Remain` to align with the current Java implementation in BF index, even though
+    // its predicate semantics are inconsistent here. In practice, equality tests in predicate
+    // evaluation always return false when the literal is null. See
+    // `null_false_leaf_binary_function.h`.
+    if (literal.IsNull()) {
+        return FileIndexResult::Remain();
+    }
     int64_t hash = hash_function_(literal);
-    return literal.IsNull() || filter_.TestHash(hash) ? FileIndexResult::Remain()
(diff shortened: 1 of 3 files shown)


lxy-9602:
The `SafeAbs(INT64_MIN)` change avoids C++ signed-overflow UB, but it still passes a negative value into the negative-side `BitSliceIndexRoaringBitmap`. That bitmap represents absolute values and expects non-negative literals, so `INT64_MIN` needs explicit handling instead of being returned from `SafeAbs`.

This affects predicate pruning when the query literal is `INT64_MIN`:

- `x > INT64_MIN` should keep all non-null rows, but the current code can drop negative rows.
- `x >= INT64_MIN` should keep all non-null rows, but the current code can drop negative rows.
- `x < INT64_MIN` should return an empty result for int64 values, but the current code can keep negative rows.
- `x <= INT64_MIN` should also be empty if the index never contains `INT64_MIN`, but the current code can keep negative rows.

I checked Java Paimon as well. Java’s BSI writer cannot successfully write `Long.MIN_VALUE`, because `Math.abs(Long.MIN_VALUE)` remains negative and `BitSliceIndexRoaringBitmap.Appender` rejects negative `min`. However, Java’s reader also does not special-case a query literal of `Long.MIN_VALUE`, so this appears to be a Java-side boundary bug too rather than behavior that C++ should copy exactly.

lxy-9602:
Already fixed in https://github.com/apache/paimon/pull/8150.

zjw1111 (approved):
+1

PR #344, opened 2026-06-08
fix: LZ4 block compressor/decompressor safety issues

lxy-9602:
### Purpose

No Linked issue.
- **Decompressor:** Add `src_length < HEADER_LENGTH` guard before `ReadIntLE` to prevent out-of-bounds read on truncated/corrupted blocks.
- **Compressor:** Change `compressed_size < 0` to `<= 0` since `LZ4_compress_default` returns 0 (not negative) on failure, which previously produced silent corrupt blocks.

### Tests
- `TestDecompressTruncatedHeader`: Verifies truncated input (3 bytes / 0 bytes) returns `Invalid` instead of reading out-of-bounds.
- `TestCompressInsufficientOutputBuffer`: Verifies insufficient output buffer returns error instead of writing a corrupt block.

### DIFF
--- a/src/paimon/common/compression/block_compression_factory_test.cpp
+++ b/src/paimon/common/compression/block_compression_factory_test.cpp
@@ -79,6 +79,37 @@ TEST_P(CompressionFactoryTest, TESTCompressThenDecompress) {
     ASSERT_EQ(0, decompressor->ReadIntLE(read_write_le.data()));
 }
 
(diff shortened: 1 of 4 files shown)


zjw1111 (approved):
Re-reviewed the latest head and the previous LZ4 output-buffer issue is fixed. No remaining actionable issues from my side.

PR #345, opened 2026-06-08
feat(cache): support manifest file cache

gripleaf:
### Purpose

  Add an optional process-local manifest file bytes cache to reduce repeated manifest file reads during scan/planning paths.

  This change:
  - Adds `manifest.cache-max-memory-size`, defaulting to `64MB`; `0` disables the cache.
  - Reuses the existing `CacheManager` to cache raw manifest file bytes.
  - Wires the shared manifest cache manager through `CoreOptions`, `ManifestList`, `ManifestFile`, and `IndexManifestFile`.
  - Reads cached manifest bytes through `MemorySegmentInputStream`, which keeps the underlying `MemorySegment` alive without copying bytes.
  - Uses a whole-file cache key to avoid extra file-status lookups on cache hits.
  - Bypasses the cache when `IOHook` is active so fault-injection tests still exercise real IO.

  ### Tests

  - `git diff --check HEAD~1 HEAD`
  - `cmake -S . -B /tmp/paimon-cpp-manifest-cache-build -DPAIMON_BUILD_TESTS=ON -DPAIMON_DEPENDENCY_SOURCE=AUTO ...`

  Added/updated UT coverage:
  - `CoreOptionsTest`
    - verifies default manifest cache size is `64MB`
    - verifies setting `manifest.cache-max-memory-size=0` disables the cache
  - `ManifestFileTest.TestReadUsesManifestCache`
    - verifies repeated reads hit cache and avoid repeated file opens
    - verifies filtering still works after reading from cached bytes
  - `ManifestFileTest.TestManifestCacheCanBeDisabled`
    - verifies disabled cache preserves original repeated-open behavior
  - `ManifestFileTest.TestManifestCacheBypassesWhenIOHookActive`
    - verifies cache is bypassed while `IOHook` is active

  Note: local `paimon-core-test` build was blocked by third-party dependency downloads because the configured proxy could not connect to GitHub.

  ### API and Format

  This change adds a new public table option:

  - `manifest.cache-max-memory-size`

  It does not change storage format or protocol compatibility. The cache stores raw manifest file bytes in process memory only.

  ### Documentation

  Yes. Added `docs/source/user_guide/manifest_cache.rst` and linked it from the user guide.

  ### Generative AI tooling

  Generated-by: OpenAI Codex


docs/source/user_guide/manifest_cache.rst
```diff
@@ -0,0 +1,85 @@
+.. Copyright 2024-present Alibaba Inc.
+
```
lxy-9602, line 2: Could you please change Copyright 2024 to Copyright 2026.

docs/source/user_guide/manifest_cache.rst
```diff
@@ -0,0 +1,85 @@
...
+- The cache is process-local LRU storage and is not shared across processes.
+- Reads bypass the cache while ``IOHook`` is active, so IO fault-injection tests
+  still observe real IO operations.
+- ``Read()`` keeps its existing append semantics, which is required when base
+  and delta manifests are loaded into the same result vector.
+
+Test Coverage
+-------------
+
+- ``CoreOptionsTest`` validates the ``64MB`` default and the ``0`` disable
```
lxy-9602: It might be enough to include only usage examples in the user guide doc, without adding tests.

include/paimon/defs.h
```diff
@@ -150,6 +150,10 @@ struct PAIMON_EXPORT Options {
     /// compaction of manifest, default value is 16MB.
     static const char MANIFEST_FULL_COMPACTION_FILE_SIZE[];
 
+    /// "manifest.cache-max-memory-size" - Maximum process-local memory used to cache raw manifest
+    /// file bytes. Set to 0 to disable. Default value is 64MB.
+    static const char MANIFEST_CACHE_MAX_MEMORY_SIZE[];
+
```
lxy-9602: Could you please clarify in the comments that this configuration differs from the Java version, and that the cache is file-granular instead of record-granular? Additionally, it would be better to align the config name with the Java convention (it calls `cache.manifest.max-memory`).

lxy-9602: For configs differing from Java, could we default them to disabled and let users opt-in during reads when needed? This might also eliminate the need to adjust the IO hook. We could also add this option to the integration tests to ensure the corresponding code path is covered.

src/paimon/core/manifest/index_manifest_file.h
```diff
@@ -60,7 +61,8 @@ class IndexManifestFile : public ObjectsFile<IndexManifestEntry> {
                       const std::shared_ptr<WriterBuilder>& writer_builder,
                       const std::string& compression,
                       const std::shared_ptr<PathFactory>& path_factory, int32_t bucket_mode,
-                      const std::shared_ptr<MemoryPool>& pool);
+                      const std::shared_ptr<MemoryPool>& pool,
+                      const std::shared_ptr<CacheManager>& cache_manager);
```
lxy-9602: Could you move pool to the last parameter and put `cache_manager` before it?

src/paimon/core/manifest/manifest_list.h
```diff
@@ -50,7 +51,8 @@ class ManifestList : public ObjectsFile<ManifestFileMeta> {
         const std::shared_ptr<FileSystem>& file_system,
         const std::shared_ptr<FileFormat>& file_format, const std::string& compression,
         const std::shared_ptr<FileStorePathFactory>& path_factory,
-        const std::shared_ptr<MemoryPool>& pool);
+        const std::shared_ptr<MemoryPool>& pool,
+        const std::shared_ptr<CacheManager>& cache_manager = nullptr);
```
lxy-9602, line 55: Please try to avoid default arguments in production code where possible, so that callers can specify them explicitly.

src/paimon/core/utils/objects_file.h
```diff
@@ -50,7 +58,8 @@ class ObjectsFile {
                 const std::shared_ptr<WriterBuilder>& writer_builder,
                 std::unique_ptr<ObjectSerializer<T>>&& serializer, const std::string& compression,
                 const std::shared_ptr<PathFactory>& path_factory,
-                const std::shared_ptr<MemoryPool>& pool);
+                const std::shared_ptr<MemoryPool>& pool,
+                const std::shared_ptr<CacheManager>& cache_manager = nullptr);
```
lxy-9602, line 59: Also please try to avoid default arguments in production code.

src/paimon/core/utils/objects_file.h
```diff
@@ -117,8 +129,29 @@ Status ObjectsFile<T>::Read(const std::string& file_name,
                             const std::function<Result<bool>(const T&)>& filter,
                             std::vector<T>* result) const {
     std::string file_path = path_factory_->ToPath(file_name);
-    PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<InputStream> file_input_stream,
-                           file_system_->Open(file_path));
+    bool should_cache = cache_manager_ && !IOHook::GetInstance()->IsActive();
+    std::shared_ptr<InputStream> file_input_stream;
```
lxy-9602, line 128: Please avoid adding IOHook in tests. For IO exception-related tests, it might be better to simply disable the cache instead.

src/paimon/core/utils/objects_file.h
```diff
@@ -154,6 +187,39 @@ Status ObjectsFile<T>::Read(const std::string& file_name,
     return Status::OK();
 }
 
+template <typename T>
+Result<MemorySegment> ObjectsFile<T>::ReadFileBytes(const std::string& file_path) const {
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<InputStream> input_stream,
+                           file_system_->Open(file_path));
+    Result<uint64_t> input_length_result = input_stream->Length();
+    if (!input_length_result.ok()) {
```
lxy-9602: Consider use `PAIMON_ASSIGN_OR_RAISE`.

src/paimon/core/manifest/manifest_file_test.cpp
```diff
@@ -181,6 +235,123 @@ TEST_F(ManifestFileTest, TestSimple) {
     ASSERT_EQ(expected_manifest_entries, manifest_entries);
 }
 
+TEST_F(ManifestFileTest, TestReadUsesManifestCache) {
+    auto pool = GetDefaultPool();
+    auto counting_file_system = std::make_shared<CountingFileSystem>();
+    EXPECT_OK_AND_ASSIGN(std::shared_ptr<FileFormat> file_format,
+                         FileFormatFactory::Get("orc", {}));
```
lxy-9602, line 241: Could we prefer ASSERT_* rather than EXPECT_* in tests?

src/paimon/core/manifest/manifest_file_test.cpp
```diff
@@ -181,6 +235,123 @@ TEST_F(ManifestFileTest, TestSimple) {
...
+    ASSERT_OK(manifest_file->Read("manifest-3ea5ee21-d399-4f1c-a749-2fc63dbf0852-1",
+                                  /*filter=*/nullptr, &first_read));
+    std::vector<ManifestEntry> second_read;
+    ASSERT_OK(manifest_file->Read("manifest-3ea5ee21-d399-4f1c-a749-2fc63dbf0852-1",
+                                  /*filter=*/nullptr, &second_read));
+
+    ASSERT_EQ(first_read, second_read);
+    ASSERT_EQ(2, counting_file_system->open_count);
+    ASSERT_EQ(0, counting_file_system->get_file_status_count);
+}
```
lxy-9602, line 315: I’m a bit curious where `GetFileStatus` is used here. `get_file_status_count` seems to stay 0 whether the cache is enabled or disabled.

lxy-9602:
The logic in this PR is very clear, and it does a great job of reusing the existing utilities. Nice work!👍 I’m curious about the original motivation for this PR — was it driven by a scan performance bottleneck caused by repeatedly scanning the same table in production? Will this cache bring a noticeable performance improvement? If possible, could you also share some benchmark or test results?

gripleaf:
> The logic in this PR is very clear, and it does a great job of reusing the existing utilities. Nice work!👍 I’m curious about the original motivation for this PR — was it driven by a scan performance bottleneck caused by repeatedly scanning the same table in production? Will this cache bring a noticeable performance improvement? If possible, could you also share some benchmark or test results?

yep, we expect Paimon to have stronger online serving capabilities, especially in point lookups and short scans. (Parquet metadata cache, page index, and file handler cache are all what we need.)

follow is what we poc above features(cold start, and without parquet page index, it would read 120MB data):
```
Analyze report:
Manifest:
  files: 6
  cache_hits: 0
  cache_misses: 6
  cache_disabled: 0
  read_bytes: 7389018
  open_time: 102.240 ms
Schema:
    read_file_time: 160.743 ms
Parquet:
  files: 1
  meta_cache_hits: 0
  meta_cache_misses: 1
  meta_cache_disabled: 0
  read_bytes: 3366722
  read_ops: 0
  read_at_ops: 55
  read_async_ops: 0
  read_at_bytes: 3366722
  read_async_bytes: 0
  next_batches: 1
  read_at_time: 1004.593 ms
    read_at_max_time: 212.223 ms
    read_at_max_bytes: 1016131
    read_async_max_bytes: 0
  next_batch_time: 819.574 ms
  page_index:
    attempts: 1
    disabled: 0
    plan_created: 1
    plan_fallback: 0
    no_index: 0
    no_column_index: 0
    no_offset_index: 0
    unsupported_predicate: 0
    fast_path_enabled: 1
    fast_path_fallback: 0
    candidate_row_groups: 1
    selected_row_groups: 1
    candidate_pages: 4
    selected_pages: 1
    skipped_pages: 3
    selected_rows: 29700
    selected_bytes: 8705
    fast_next_time: 819.539 ms
      candidate_rows_after_filter: 49
      matched_rows: 1
      filter_column_read_time: 105.639 ms
      predicate_column_read_time: 221.298 ms
      output_column_read_time: 489.303 ms
Read path:
  create_plan_time: 1782.933 ms
    starting_scan_time: 1782.929 ms
      snapshot_raw_plan_time: 1389.236 ms
        file_store_read_manifest_entries_time: 1237.905 ms
          read_file_entries_time: 1237.897 ms
            task_time: 1497.192 ms
            task_max_time: 1237.780 ms
              manifest_file_read_time: 1497.108 ms
              manifest_file_read_max_time: 1237.730 ms
                build_reader_time: 623.999 ms
                  build_reader_max_time: 409.022 ms
                next_batch_time: 273.773 ms
                deserialize_time: 489.439 ms
                records: 61265
                max_records: 1024
            collect_time: 1237.869 ms
  create_reader_time: 388.457 ms
    split_time: 388.456 ms
      raw_split_create_raw_file_readers_time: 388.430 ms
        reader_loop_time: 388.427 ms
          field_mapping_reader_time: 388.413 ms
            open_file_time: 10.986 ms
```

include/paimon/cache/file_page_cache.h
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright 2024-present Alibaba Inc.
```
lucasfang: 2026

gripleaf:
### Update after latest commit

The latest commit changes the manifest cache design from an internal table-option-driven cache to an externally injected cache abstraction.

The motivation for the design adjustment is
- The `manifest.cache-max-memory-size` is only in readcontext scope.
- Custom cache implementations can support a wider variety of cache forms and strategies, such as remote cache, local disk cache, etc.

Instead of adding `manifest.cache-max-memory-size`, paimon-cpp now exposes a public `FilePageCache` interface in `include/paimon/cache/file_page_cache.h`. Embedding applications can provide their own cache implementation and inject it through `ScanContextBuilder::WithManifestCache` / `ReadContextBuilder::
WithManifestCache`. If no cache is provided, manifest reads keep the original behavior and go directly to the underlying `FileSystem`.

This also means:

- `manifest.cache-max-memory-size` has been removed; there is no built-in 64MB default cache and no `0` option-based disable switch.
- `CoreOptions` no longer creates or owns a `CacheManager`; it only carries an optional `std::shared_ptr<FilePageCache>`.
- Manifest caching still happens at `ObjectsFile<T>::Read()` and still uses a whole-file key, now represented as `FilePageKey { path, offset = 0, size = -1 }`.
- Cached values are now `Bytes` buffers supplied by the external cache loader and read via `ByteArrayInputStream`; the previously added `MemorySegmentInputStream` was removed.
- `ManifestFile`, `ManifestList`, and `IndexManifestFile` now receive `FilePageCache` instead of `CacheManager`.
- Read/scan paths propagate the injected cache through `ReadContext`, `ScanContext`, `CoreOptions`, `TableRead`, `TableScan`, and `BlobViewLookup`.
 - Write/commit/clean paths explicitly pass `nullptr`, so they do not use the manifest cache.
- The previous IOHook-specific cache bypass behavior was removed together with the internal cache ownership model.

### Updated tests

The unit tests were adjusted to match the external-injection model:

- `CoreOptionsTest` now verifies that the default manifest cache is `nullptr`.
- `ManifestFileTest.TestReadUsesManifestCache` uses a test `FilePageCache` implementation to verify repeated manifest reads hit the cache and avoid repeated file opens.
- `ManifestFileTest.TestManifestCacheCanBeDisabled` verifies that omitting the cache preserves repeated-open behavior.
- The previous IOHook bypass test was removed because cache bypass is no longer managed inside `ObjectsFile`.
- `ManifestList` callers in tests were updated to pass an explicit `file_page_cache` argument.

### Updated docs

The documentation page was updated from “Manifest Cache” to “File Page Cache” and now describes the API-based injection model instead of the removed `manifest.cache-max-memory-size` table option.

include/paimon/cache/file_page_cache.h
```diff
@@ -0,0 +1,94 @@
...
+using PageLoader = std::function<Result<std::shared_ptr<Bytes>>(const FilePageKey& key,
+                                                                MemoryPool* memory_pool)>;
+
+/// A pluggable cache for byte pages backed by files.
+///
+/// `FilePageCache` is the public abstraction injected by embedding applications into paimon-cpp.
+/// It is intentionally minimal so that callers can map it onto whatever caching framework they
+/// already maintain (LRU, LFU, off-heap, distributed, etc.). paimon-cpp itself does not provide a
+/// default implementation; each integrator is expected to supply its own.
+class PAIMON_EXPORT FilePageCache {
```
lucasfang: Why was a new FilePageCache added? Can we use the existing ./common/io/cache/cache.h instead?

src/paimon/core/operation/read_context.cpp
```diff
@@ -207,6 +211,12 @@ ReadContextBuilder& ReadContextBuilder::WithFileSystem(
     return *this;
 }
 
+ReadContextBuilder& ReadContextBuilder::WithManifestCache(
```
lucasfang: I think the WithManifestCache interface could be further abstracted. Otherwise, every time a new file type is added, we might need to add a corresponding WithXXXCache interface for that file type.

include/paimon/cache/file_page_cache.h
gripleaf: Custom cache implementations can support a wider variety of cache forms and strategies, such as remote cache, local disk cache, etc.

src/paimon/core/operation/read_context.cpp
gripleaf: Perhaps we could replace it with a `FileCacheFactory` (similar to `FileFormatFactory`), providing a `Get` interface to retrieve cache instances of different types?

lucasfang: What are the benefits of separating caches for different file types, and why not just share a single cache? Also, if we split the caches too granularly, how do we prevent passing the wrong cache type during internal handoffs, which could lead to cache misses

include/paimon/cache/file_page_cache.h
lucasfang: Well, Cache.h can also be customized, you can move it to the include directory.

src/paimon/core/operation/read_context.cpp
gripleaf: - Different file types have different caching approaches. For example, memory obviously cannot hold many data files, so I may need to cache them on the local disk, while files like Manifest have a limited total data size and can be fully stored in memory.
- Cache eviction for data files might affect meta files like Manifest, making fine-grained management impossible.

lucasfang: I think your strategies make perfect sense. However, I'd suggest keeping them encapsulated within the cache itself to maintain a clean interface and ensure the internals are easy to manage. Having a unified entry point for the cache is crucial; the underlying implementation and eviction strategies can still be fully customized behind the scenes.

lucasfang: The current cache should already handle similar cases. We could probably extend the abstraction here

gripleaf: I agree. What I'm thinking is to make the LRUCache implementation used in CacheManager customizable, so that the cache can be uniformly managed externally.

gripleaf:
## Update after latest commit

  The latest commit changes manifest cache from the previous `FilePageCache` injection model to a generic `Cache` abstraction.

 ### Main changes:

  - Removed the public `FilePageCache` interface.
  - Moved the cache abstraction to `include/paimon/cache/cache.h`.
  - Exposed `Cache`, `CacheKey`, `CacheValue`, `CacheCallback`, and `CacheKind`.
  - Currently `CacheKind` only supports `CacheKind::Manifest`.
  - Moved `MemorySegment` to `include/paimon/memory/memory_segment.h` because `CacheValue` exposes it.
  - Manifest cache is now injected through:
    - `ScanContextBuilder::WithCache(CacheKind::Manifest, cache)`
    - `ReadContextBuilder::WithCache(CacheKind::Manifest, cache)`

  This also means:

  - There is still no built-in default manifest cache.
  - If no cache is injected, manifest reads keep the original behavior and go directly to the underlying `FileSystem`.
  - `CoreOptions::FromMap()` no longer accepts a cache argument.
  - Read/scan contexts carry a map of `CacheKind -> Cache`, and pass the manifest cache into `CoreOptions` with `CoreOptions::WithCache()`.
  - Manifest caching still happens in `ObjectsFile<T>::Read()`.
  - The cache key remains a whole-file key, now created by `CacheKey::ForPosition(path, 0, -1, false)`.
  - Cached manifest bytes are stored as `CacheValue` with a `MemorySegment`.
  - On cache hits, `ObjectsFile::Read()` keeps the cached bytes alive while `ByteArrayInputStream` reads from them.

  The motivation of design:
  - ScanContext/ReadContextare not strongly coupled with Catalog/Table.
  - There is a certain need for managing caches at the granularity of Catalog or Table, and simply replacing the underlying cache implementation is not enough.

###  Updated tests:

  - `ScanContextTest` and `ReadContextTest` now verify `WithCache(CacheKind::Manifest, cache)` injection.
  - `ManifestFileTest.TestReadUsesManifestCache` now uses `LruCache` through the generic `Cache` interface.
  - `ManifestFileTest.TestManifestCacheIsDisabledWithoutInjectedCache` verifies no injected cache preserves repeated file opens.
  - Existing manifest cache disabled behavior is kept for the no-cache path.

###  Updated docs:

  - The manifest cache doc now describes the generic `Cache` / `CacheKind::Manifest` injection model instead of the removed `FilePageCache` API.

@lxy-9602 @lucasfang  I have updated the implementation of the Manifest cache. Could you please help review the proposal?

include/paimon/cache/cache.h
```diff
@@ -49,31 +70,21 @@ class PAIMON_EXPORT Cache {
...
+    const MemorySegment& GetSegment() const;
+
+    void OnEvict(const std::shared_ptr<CacheKey>& key) const;
+
+    bool operator==(const CacheValue& other) const;
 
  private:
-    MemorySegment segment_;
+    std::shared_ptr<MemorySegment> segment_;
     CacheCallback callback_;
```
lxy-9602, line 99: Since `MemorySegment` already holds a `Byte` pointer, introducing too many pointers here may add noticeable `shared_ptr` construction/destruction overhead. I’d suggest switching back to `MemorySegment` objects instead.

include/paimon/scan_context.h
```diff
@@ -178,6 +186,12 @@ class PAIMON_EXPORT ScanContextBuilder {
     /// @note If not set, use default file system (configured in `Options::FILE_SYSTEM`)
     ScanContextBuilder& WithFileSystem(const std::shared_ptr<FileSystem>& file_system);
 
+    /// Inject a cache for the given cache kind.
+    /// @param kind The cache kind to set. Currently only CacheKind::Manifest is used.
+    /// @param cache The cache instance to use. Passing nullptr disables this cache kind.
+    /// @return Reference to this builder for method chaining.
+    ScanContextBuilder& WithCache(CacheKind kind, const std::shared_ptr<Cache>& cache);
+
```
lxy-9602, line 191: I’d suggest keeping only `WithCache(const std::shared_ptr<Cache>& cache)`. The cache can be implemented as a forwarding cache that owns a set of specialized caches. `CacheKey` could include a `CacheKind`, allowing the forwarding cache to route requests based on the kind.

src/paimon/core/manifest/manifest_file_test.cpp
lxy-9602: This test case might not be necessary, since it looks the same as `TestManifestCacheIsDisabledWithoutInjectedCache`. Could we remove it to avoid duplication?

src/paimon/core/utils/objects_file.h
```diff
@@ -117,8 +127,33 @@ Status ObjectsFile<T>::Read(const std::string& file_name,
...
+            PAIMON_ASSIGN_OR_RAISE(MemorySegment segment, ReadFileSegment(file_path));
+            return std::make_shared<CacheValue>(segment, CacheCallback());
+        };
+        Result<std::shared_ptr<CacheValue>> cache_result = cache_->Get(cache_key, supplier);
+        if (cache_result.ok() && cache_result.value() &&
+            cache_result.value()->GetSegment().Data() != nullptr) {
+            cached_bytes = cache_result.value()->GetSegment().GetOrCreateHeapMemory(pool_.get());
+            file_input_stream =
+                std::make_shared<ByteArrayInputStream>(cached_bytes->data(), cached_bytes->size());
+        }
```
lxy-9602, line 147: I’d like to confirm whether there may be a concurrency issue here. If multiple scans share the same cache, and `file_input_stream` does not retain a smart pointer to the underlying bytes, could this lead to a memory issue if the cached entry is replaced or evicted while it is still in use?

src/paimon/core/utils/objects_file.h
```diff
@@ -154,6 +189,34 @@ Status ObjectsFile<T>::Read(const std::string& file_name,
...
+        return Status::Invalid(
+            fmt::format("file {}, length {} is too large", file_path, input_length));
+    }
+
+    PAIMON_RETURN_NOT_OK(input_stream->Seek(0, FS_SEEK_SET));
+    auto bytes = std::make_shared<Bytes>(static_cast<size_t>(input_length), pool_.get());
+    int64_t offset = 0;
+    while (offset < static_cast<int64_t>(input_length)) {
+        uint32_t read_size = static_cast<uint32_t>(std::min<int64_t>(
+            static_cast<int64_t>(input_length) - offset, std::numeric_limits<uint32_t>::max()));
```
lxy-9602: It should be okay to read the entire data in one call and just verify that `actual_read_size == read_size`, without a loop, since the current filesystem implementations guarantee that the actual read size matches the requested size.

src/paimon/core/utils/objects_file.h
```diff
@@ -154,6 +189,34 @@ Status ObjectsFile<T>::Read(const std::string& file_name,
...
+        if (actual_read_size <= 0) {
+            return Status::IOError(fmt::format(
+                "Unexpected EOF while reading manifest file {}, expected {} bytes, got {} bytes",
+                file_path, input_length, offset));
+        }
+        offset += actual_read_size;
+    }
+    return MemorySegment::Wrap(bytes);
+}
+
```
lxy-9602, line 207: Also, could you modify the scan/scan-and-read integration test that runs all with/without cache modes to verify end-to-end correctness?

src/paimon/core/utils/objects_file.h
gripleaf: cached_bytes hold the smart pointer, it has the same lifecycle with file_input_stream

include/paimon/scan_context.h
gripleaf: done

docs/source/user_guide/manifest_cache.rst
```diff
@@ -0,0 +1,93 @@
...
+
+.. code-block:: cpp
+
+   auto manifest_cache = std::make_shared<MyCache>();
+
+   paimon::ScanContextBuilder scan_builder(table_path);
+   scan_builder.WithCache(manifest_cache);
+
+   paimon::ReadContextBuilder read_builder(table_path);
+   read_builder.WithCache(manifest_cache);
```
lxy-9602: Since the read path here does not read manifest files or other metadata files at all, I’d suggest using a dedicated read cache for this example.

docs/source/user_guide/manifest_cache.rst
```diff
@@ -0,0 +1,93 @@
...
+  read or scan contexts.
+- On a cache hit, ``ObjectsFile::Read()`` keeps the cached ``Bytes`` alive while
+  ``ByteArrayInputStream`` reads from it without copying. Caller filters are
+  applied after records are decoded.
+- ``Read()`` keeps its existing append semantics, which is required when base
+  and delta manifests are loaded into the same result vector.
+
+Test Coverage
+-------------
+
```
lxy-9602: Could you please remove the “Implementation Notes” and “Test Coverage” sections from the documentation? The documentation should primarily focus on user-facing usage, and these code-level implementation details may be too detailed for it.

lxy-9602: That said, it’s really great that the documentation is being updated together with the feature work.👍

include/paimon/cache/cache.h
```diff
@@ -15,25 +15,58 @@
...
+    Manifest,
+};
+
+class PAIMON_EXPORT CacheKey {
+ public:
+    static std::shared_ptr<CacheKey> ForPosition(const std::string& file_path, int64_t position,
+                                                 int32_t length, bool is_index);
+
+ public:
+    virtual ~CacheKey() = default;
```
lxy-9602, line 45: It would be nice to add another static create helper/function (may be`ForKind`) to simplify construction.

src/paimon/core/operation/scan_context.cpp
```diff
@@ -143,6 +147,11 @@ ScanContextBuilder& ScanContextBuilder::WithFileSystem(
     return *this;
 }
 
+ScanContextBuilder& ScanContextBuilder::WithCache(const std::shared_ptr<Cache>& cache) {
+    impl_->cache_ = Cache::WarpKind(CacheKind::Manifest, cache);
+    return *this;
+}
```
lxy-9602, line 153: It looks like `WarpKind` has already been called in places like the manifest. Should we avoid calling it again here to prevent wrapping twice?

include/paimon/cache/cache.h
```diff
@@ -15,25 +15,58 @@
...
+    CacheKind GetKind() const {
+        return kind_;
+    }
+
+    virtual bool Equals(const CacheKey& other) const = 0;
+
+    virtual size_t HashCode() const = 0;
+
+ private:
+    CacheKind kind_ = CacheKind::Manifest;
```
lxy-9602: Could we add a default value, such as `CacheKind::DEFAULT`? Using `Manifest` as the default feels a bit odd.

include/paimon/cache/cache.h
```diff
@@ -15,25 +15,58 @@
...
+#include "paimon/memory/memory_segment.h"
 #include "paimon/result.h"
+#include "paimon/visibility.h"
 
 namespace paimon {
 
 class CacheValue;
+enum class CacheKind {
+    Manifest,
+};
```
lxy-9602, line 35: We usually use all-uppercase names for enum class values in our codebase. (e.g., `MANIFEST`)

include/paimon/cache/cache.h
```diff
@@ -15,25 +15,58 @@
...
+    CacheKind kind_ = CacheKind::Manifest;
+};
 
-/// Callback invoked when a cache entry is evicted by the LRU policy.
 using CacheCallback = std::function<void(const std::shared_ptr<CacheKey>&)>;
 
 class PAIMON_EXPORT Cache {
  public:
+    static std::shared_ptr<Cache> WarpKind(CacheKind kind, const std::shared_ptr<Cache>& cache);
+
```
lxy-9602: Also, I think introducing `WrapKind` makes the design a bit more complicated. In `ObjectsFile`, we could instead set the kind on `cache_key` directly when calling `cache->Get(cache_key)`, for example via the `ForKind` method mentioned above. That would avoid wrapping in multiple places.

include/paimon/cache/cache.h
gripleaf: Currently, CacheKeys are created using static functions like ForPosition. If we want to add CacheKind to the constructor, we would need to add a parameter to ForPosition. Therefore, I prefer to keep CacheKind independent externally (encapsulated by Cache::Wrap).

include/paimon/cache/cache.h
gripleaf: ObjectsFile is located at the bottom layer of the interface invocation. To support ForKind, CacheKind would need to be passed down all the way from the upper layers. The purpose of using WrapKind is to combine CacheKind and Cache together.

lxy-9602: Thank you for your reply! Maybe I wasn’t very clear earlier. What we’re expecting is for `CacheKey` to provide another static `ForKind` helper, for example:

```cpp
static std::shared_ptr<CacheKey> ForKind(
    const std::string& file_path, int64_t position, int32_t length, CacheKind kind);
```

With that, we would only need to construct the correct key when getting/putting cache entries in `ObjectsFile`, instead of introducing an intermediate `WrapKind` step.

For the parquet footer cache later on, we could define dedicated `CacheKind` values such as `DATA_FOOTER` or `INDEX`, and then explicitly specify the cache kind when putting/getting the parquet cache.

gripleaf: Okay, now we need to solve the problem of where the `CacheKind kind` parameter comes from in `ObjectsFile`.

If We assume only Manifest could use `ObjectsFile`, then we use CacheKind::MANIFEST by default, or we need add `CacheKind kind` to the constructor of ObjectsFile(what I am working on).

src/paimon/core/utils/objects_file.h
```diff
@@ -154,6 +187,28 @@ Status ObjectsFile<T>::Read(const std::string& file_name,
...
+    if (input_length > std::numeric_limits<int32_t>::max()) {
+        return Status::Invalid(
+            fmt::format("file {}, length {} is too large", file_path, input_length));
+    }
+
+    PAIMON_RETURN_NOT_OK(input_stream->Seek(0, FS_SEEK_SET));
+    auto bytes = std::make_shared<Bytes>(static_cast<size_t>(input_length), pool_.get());
+    PAIMON_ASSIGN_OR_RAISE(int32_t actual_read_size,
+                           input_stream->Read(bytes->data(), static_cast<uint32_t>(input_length)));
+    if (actual_read_size != static_cast<int32_t>(input_length)) {
```
lxy-9602: The filesystem has recently gone through a round of refactoring, and most interfaces now use int64_t, such as `Read` and `Length`. This way, we can avoid so many casts here.

test/inte/scan_inte_test.cpp
```diff
@@ -241,6 +242,32 @@ class ScanInteTest : public testing::Test {
...
+    ASSERT_GT(manifest_cache->SupplierCallCount(), 0);
+    int64_t supplier_calls_after_first_scan = manifest_cache->SupplierCallCount();
+
+    ASSERT_OK_AND_ASSIGN(auto second_plan, run_scan());
+    ASSERT_EQ(first_plan->SnapshotId(), second_plan->SnapshotId());
+    ASSERT_EQ(first_plan->Splits().size(), second_plan->Splits().size());
+    ASSERT_GT(manifest_cache->GetCount(), supplier_calls_after_first_scan);
+    ASSERT_EQ(supplier_calls_after_first_scan, manifest_cache->SupplierCallCount());
+}
+
```
lxy-9602, line 284: I recall that in a previous change, you added cache/no-cache `TEST_P` variants for all scan_inte tests, which I thought was a great approach. Would you be open to bringing that logic back? We can keep the current `TestRepeatedScanReusesManifestCache` case as well. This is the last remaining blocker for me.

gripleaf: I have added back CacheMode and TEST_P in scan_inte_test.cpp

lxy-9602 (approved):
+1

lucasfang:
+1

PR #349, opened 2026-06-08
feat(cache): support parquet metadata cache

gripleaf:
### Purpose

  Support caching parsed Parquet FileMetaData to avoid repeatedly reading and parsing Parquet footers when creating readers for the same file.

  This change:

  - Adds ParquetMetadataCache, backed by GenericLruCache.
  - Adds option parquet.read.metadata-cache.max-bytes.
  - Initializes one metadata cache per ParquetReaderBuilder when the configured size is greater than 0.
  - Uses cached metadata in ParquetFileBatchReader::Create through Parquet FileReaderBuilder::Open(..., file_metadata).
  - Keeps cache entries keyed by file URI.

  ### Tests

  Added/updated UT cases in parquet_file_batch_reader_test:

  - TestReadRewrittenFileWithoutMetadataCache
  - TestMetadataCacheGetOrLoad
  - TestMetadataCacheNullLoaderResultIsNotCached

  Not run locally: full CMake build/tests, because the local build blocks on bundled Arrow/Boost dependency download.

  ### API and Format

  No storage format or protocol changes.

  No public include/ API changes.

  Adds a Parquet read option:

  - parquet.read.metadata-cache.max-bytes

  ### Documentation

  This introduces a new Parquet reader optimization. No standalone documentation changes are included.

  ### Generative AI tooling


src/paimon/format/parquet/parquet_file_batch_reader.h
```diff
@@ -64,7 +64,8 @@ class ParquetFileBatchReader : public PrefetchFileBatchReader {
     static Result<std::unique_ptr<ParquetFileBatchReader>> Create(
         std::shared_ptr<arrow::io::RandomAccessFile>&& input_stream,
         const std::shared_ptr<arrow::MemoryPool>& pool,
-        const std::map<std::string, std::string>& options, int32_t batch_size);
+        const std::map<std::string, std::string>& options, int32_t batch_size,
+        std::shared_ptr<::parquet::FileMetaData> file_metadata = nullptr);
```
lxy-9602, line 69: Could you please try to avoid default arguments in production code.

src/paimon/format/parquet/parquet_format_defs.h
```diff
@@ -64,6 +64,15 @@ static inline const char PARQUET_READ_ENABLE_PAGE_INDEX_FILTER[] =
...
 
+// Enable reusing parsed Parquet FileMetaData (footer) across reader creations that
+// share the same ParquetReaderBuilder. Entries are keyed by file URI and file length.
+// Maximum total weight (in bytes of the thrift-encoded footer) of cached FileMetaData
+// entries; entries are evicted in LRU order once exceeded. A value of 0 disables the
+// cache. Default is 128 MiB.
+static inline const char PARQUET_READ_METADATA_CACHE_MAX_BYTES[] =
+    "parquet.read.metadata-cache.max-bytes";
+static constexpr int64_t DEFAULT_PARQUET_READ_METADATA_CACHE_MAX_BYTES = 128 * 1024 * 1024;
+
```
lxy-9602: I’d suggest disabling this by default to align with java.

src/paimon/format/parquet/parquet_reader_builder.h
```diff
@@ -46,9 +55,33 @@ class ParquetReaderBuilder : public ReaderBuilder {
...
+                    } catch (const std::exception& e) {
+                        return Status::Invalid("ParquetReaderBuilder::Build: ", e.what());
+                    } catch (...) {
+                        return Status::UnknownError("ParquetReaderBuilder::Build: unknown error");
+                    }
+                };
+                PAIMON_ASSIGN_OR_RAISE(file_metadata, metadata_cache_->GetOrLoad(file_uri, loader));
+            }
+        }
+
```
lxy-9602, line 62: I’d like to better understand in what scenarios the same file would be read multiple times. As far as I remember, the code only has prefetch-related optimizations, and prefetch should already open files concurrently. In that case, how much can the metadata cache realistically reduce read latency? Do we have any concrete data showing that the Parquet footer is a bottleneck and that this optimization provides a meaningful latency improvement? Perhaps this is mainly beneficial for extremely wide tables with a large number of fields?

src/paimon/format/parquet/parquet_metadata_cache.h
```diff
@@ -0,0 +1,77 @@
...
+    explicit ParquetMetadataCache(int64_t max_weight_bytes);
+
+    /// Look up cached metadata for the given file uri. On a cache miss, invokes
+    /// `loader` (outside the cache lock) to produce the metadata and inserts the result
+    /// into the cache before returning it. If `loader` fails, the error is propagated
+    /// and nothing is cached. Modeled after CacheManager::GetPage so callers do not
+    /// have to orchestrate Get/Put themselves.
+    Result<std::shared_ptr<::parquet::FileMetaData>> GetOrLoad(const std::string& uri,
+                                                               const MetadataLoader& loader);
+
```
lxy-9602, line 57: Because captures in `std::function` may introduce some memory issue, I’d suggest preferring pass-by-value over const Func&.

src/paimon/format/parquet/parquet_file_batch_reader_test.cpp
```diff
@@ -203,6 +205,83 @@ TEST_F(ParquetFileBatchReaderTest, TestReadBinaryWrittenFromBinaryAndLargeBinary
...
+    };
+
+    ASSERT_OK_AND_ASSIGN(auto metadata1, cache.GetOrLoad(file_path_, loader));
+    ASSERT_OK_AND_ASSIGN(auto metadata2, cache.GetOrLoad(file_path_, loader));
+    ASSERT_EQ(metadata1.get(), metadata2.get());
+    ASSERT_EQ(1, load_count);
+
+    ASSERT_OK_AND_ASSIGN(auto metadata3, cache.GetOrLoad(file_path_ + ".copy", loader));
+    ASSERT_NE(metadata1.get(), metadata3.get());
+    ASSERT_EQ(2, load_count);
```
lxy-9602: Do we actually need to copy the `file_path.copy` file here? Also, it might be better to put the cache-related tests into a separate cache_test file.

src/paimon/format/parquet/parquet_reader_builder.h
gripleaf: We provide online services on top of Paimon, primarily focused on point lookups and short scans, and we might disable the prefetch feature (to avoid wasting bandwidth on prefetching). I understand that the current paimon-cpp may be more geared towards offline analytics scenarios.

gripleaf: we also need offline analytics now, I means this features is focus on our online serving scenarios.

lxy-9602:
I wanted to confirm one point about the Parquet footer: in point lookup, is the footer becoming a performance hotspot because of parsing or because of I/O? If it is mainly I/O, could we handle it similarly to how we cache manifests? The cache added in `ReadContext` does not seem to be used yet. If the bottleneck is parsing, then we could consider caching a `::parquet::FileMetaData` object instead. That said, this optimization feels somewhat specialized and would be more intrusive to the codebase, so I think we should first confirm that the performance benefit is large enough.

gripleaf:
cache parquet footer bytes in #373 , close this PR

PR #350, opened 2026-06-08
fix: address ubsan findings

zjw1111:
### Purpose

Linked issue: N/A

This PR fixes several undefined-behavior findings exposed after enabling sanitizer flags on the object library target. It also makes Hive bucket hashing use unsigned wraparound semantics at the interface level, matching the Java-style overflow behavior without relying on C++ signed overflow.

The changes include:

- Apply sanitizer flags to `paimon_objlib` so implementation files are compiled with UBSan/ASan options.
- Add sanitizer suppressions for known third-party Arrow timestamp boundary behavior and Lance/Tokio LSAN shutdown allocations.
- Guard zero-length memory copies before pointer arithmetic.
- Avoid undefined integer operations in decimal parsing, decimal tests, global index empty appends, and ORC timestamp null handling.
- Switch Hive hashing APIs and accumulation to unsigned integer types.

### API and Format

No public API, storage format, or protocol changes. The Hive hasher internal interface now uses unsigned hash value types.

### Documentation

No user-facing documentation changes.

### Generative AI tooling

Generated-by: Codex


src/paimon/common/utils/decimal_utils.cpp
```diff
@@ -71,27 +72,51 @@ std::optional<arrow::Decimal128> DecimalUtils::RescaleDecimalWithOverflowCheck(
...
-            bool is_negative = str[0] == '-';
-            size_t posn = is_negative ? 1 : 0;
-            while (posn < length) {
-                size_t group = std::min(18ul, length - posn);
-                int64_t chunk = std::stoll(str.substr(posn, group));
-                int64_t multiple = 1;
-                for (size_t i = 0; i < group; ++i) {
-                    multiple *= 10;
+        if (length == 0) {
+            return Decimal::int128_t{0};
```
lxy-9602: return bad status when "".

lxy-9602 (approved):
+1

PR #351, opened 2026-06-09
fix: remove trailing semicolons from statement-like macro definitions

lucasfang:
### Purpose

Fix statement-like macro definitions (`PAIMON_ASSIGN_OR_RAISE` family and `ASSERT_OK_AND_ASSIGN` / `EXPECT_OK_AND_ASSIGN`) that included a trailing semicolon in the macro body. This silently allowed callers to omit the semicolon without any compiler error. Removing the trailing semicolon from the macro definitions ensures the compiler enforces correct usage at all call sites.

Also fix 11 call sites across src/ and test/ that were missing the required semicolon.

### Tests

- Full-repo static scan confirmed all `PAIMON_ASSIGN_OR_RAISE` / `ASSERT_OK_AND_ASSIGN` / `EXPECT_OK_AND_ASSIGN` call sites now have trailing semicolons
- Incremental compilation of modified `.cpp` files passed

### API and Format

No. This change only affects macro definitions in internal/test headers. The calling convention (always write a trailing `;`) is unchanged for correct existing code.

### Documentation

No.

### Generative AI tooling

Generated-by: Aone Copilot (GPT-5.5)

### DIFF
--- a/include/paimon/result.h
+++ b/include/paimon/result.h
@@ -270,11 +270,11 @@ inline Status GenericToStatus(Result<T>&& res) {
 #define PAIMON_ASSIGN_OR_RAISE_IMPL(result_name, lhs, rexpr)                                 \
     auto&& result_name = (rexpr);                                                            \
     PAIMON_RETURN_IF_(!(result_name).ok(), (result_name).status(), PAIMON_STRINGIFY(rexpr)); \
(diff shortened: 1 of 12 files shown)


lszskye (approved):
+1

PR #352, opened 2026-06-09
fix: change FileStorePathFactory::Create() to return shared_ptr

lucasfang:
fix: change FileStorePathFactory::Create() to return shared_ptr

### Purpose

FileStorePathFactory inherits `enable_shared_from_this<FileStorePathFactory>` but `Create()` returned `unique_ptr`. The following 4 public methods call `shared_from_this()` internally:

- `CreateManifestFileFactory()`
- `CreateManifestListFactory()`
- `CreateIndexManifestFileFactory()`
- `CreateGlobalIndexFileFactory()`

When the object is held by `unique_ptr`, calling any of these methods throws `std::bad_weak_ptr`. This PR changes `Create()` to return `std::shared_ptr` so the ownership model is consistent.

### Tests

- Existing `FileStorePathFactoryTest` (9 cases) and `FileStorePathFactoryCacheTest` (1 case) all pass.
- No new tests needed — this is a type-signature fix; runtime behavior is verified by existing tests that exercise `CreateManifestFileFactory()` etc.

### API and Format

- `FileStorePathFactory::Create()` return type changes from `Result<std::unique_ptr<FileStorePathFactory>>` to `Result<std::shared_ptr<FileStorePathFactory>>`. Most callers already receive the result into `std::shared_ptr` (via implicit conversion), so impact is minimal.
- No storage format or protocol changes.

### Documentation

No.

### Generative AI tooling

Generated-by: Aone Copilot (Claude 4.8 Opus)

### DIFF
--- a/src/paimon/core/append/append_compact_coordinator.cpp
+++ b/src/paimon/core/append/append_compact_coordinator.cpp
@@ -209,7 +209,7 @@ Status ValidateTable(const std::shared_ptr<TableSchema>& table_schema,
 }
 
 /// Build FileStorePathFactory from core options and table schema.
-Result<std::unique_ptr<FileStorePathFactory>> BuildPathFactory(
+Result<std::shared_ptr<FileStorePathFactory>> BuildPathFactory(
     const std::string& table_path, const std::shared_ptr<TableSchema>& table_schema,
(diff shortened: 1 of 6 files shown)


zjw1111 (approved):
+1

PR #353, opened 2026-06-10
fix: prevent zero-thread DefaultExecutor and validate FileSystem/SchemeMap conflict

zjw1111:
## Summary

Fix two categories of defensive validation issues:

### 1. Zero-thread DefaultExecutor prevention

`DefaultExecutor` previously allowed `thread_count == 0`, which could create a thread pool with no workers. Tasks added via `Add()` would then silently never execute.

**Fixes:**
- `CreateDefaultExecutor(uint32_t thread_count)`: reject `thread_count == 0` with `Status::Invalid`
- `DefaultExecutor` constructor: assert that callers have already passed a positive thread count
- `GetGlobalDefaultExecutor()`: guard against `std::thread::hardware_concurrency()` returning 0
- Updated direct `CreateDefaultExecutor(thread_count)` callers to propagate the new `Result`
- `ReadContextBuilder::Finish()`: validate `prefetch_max_parallel_num > 0` before creating a default prefetch executor
- `AbstractFileStoreWrite`: use the no-argument default executor constructor for its compact executor

### 2. WithFileSystem / WithFileSystemSchemeToIdentifierMap conflict validation

The API doc for `WithFileSystemSchemeToIdentifierMap()` states it "Cannot be used together with `WithFileSystem()`", but `ReadContextBuilder::Finish()` did not enforce this. When both were set, `CoreOptions::ParseFileSystem` would silently prioritize `specific_file_system` and ignore the scheme map, potentially causing callers to believe scheme-based routing was active when it was not.

**Fixes:**
- `ReadContextBuilder::Finish()`: reject simultaneous use with `Status::Invalid`
- Updated `read_context_test.cpp`: split the test that set both, added dedicated conflict/validation tests using `ASSERT_NOK_WITH_MSG`

## Validation

- `cmake --build build -j 64`


lxy-9602 (commented):
+1

lszskye (approved):
+1

PR #354, opened 2026-06-10
fix(scan): fix dangling reference and unsigned overflow in FileStoreScan

lxy-9602:
### Purpose

Fix `FileStoreScan`:

1. `ReadFileEntries` captures loop variable `meta` by value into async task.

2. `LAST_SCAN_SKIPPED_TABLE_FILES` computes `int64_t - size_t`, which implicitly converts to `uint64_t` and wraps around when the result is negative (e.g. DELTA scan with delete entries). Fixed by casting to signed and clamping to non-negative.


zjw1111 (approved):
+1

Issue #342, opened 2026-06-07
[Feature] Support columnar-extend storage layout for MAP columns


lxy-9602:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/paimon-cpp/issues) and found nothing similar.

### Motivation

In time-series / IoT / observability workloads, a common pattern is storing free-schema fields in a `MAP<STRING, T>` column (e.g. `metrics MAP<STRING, DOUBLE>`). The default MAP storage (two KV arrays) provides:

- No per-key columnar access
- No per-key statistics
- No predicate pushdown on individual keys

This makes queries like `SELECT ext_map['usage'] FROM metrics WHERE ext_map['usage'] > 30` scan the entire MAP column — extremely inefficient when only 1–3 keys out of thousands are needed per query.

The [PIP-43: Columnar Storage Optimization for MAP Type in Paimon](https://cwiki.apache.org/confluence/display/PAIMON/PIP-43%3A+Columnar+Storage+Optimization+for+MAP+Type+in+Paimon) proposes a new `shared-shredding` storage layout that stores MAP values in `K` reusable physical columns within a Struct, achieving near-full columnar access with per-key statistics and predicate pushdown — without changing the logical type (`MAP<STRING, T>`).

#### Physical Layout

Each `MAP<STRING, T>` column configured with `fields.<column>.map.storage-layout = shared-shredding` is physically stored as:

```
STRUCT<
  __field_mapping: FixedSizeList<Int32, K>,   -- per-row: which field_id each col holds
  __col_0: T, __col_1: T, ..., __col_{K-1}: T,  -- reusable typed columns
  __overflow: MAP<INT32, T>                    -- rare fallback for rows with > K fields
>
```

`fields.<column>.map.shared-shredding.max-columns` controls K_max, and `fields.<column>.map.shared-shredding.column-placement-policy` controls column placement.

File metadata (footer) stores: field name↔id dictionary, field_id→physical column set S, overflow set O, K, and max row width.

#### Write Path

1. **Schema conversion utilities** — Logical MAP → physical Struct schema rewriting; metadata serialization/deserialization; shared-shredding column detection via field metadata marker.

2. **`FormatWriter::AddMetadata`** — New virtual method (default no-op) for writing key-value metadata to file footer before `Finish()`. Parquet implementation calls `AddKeyValueMetadata`.

[...]

PR #355, opened 2026-06-10
feat: support read for nested type sub column

lucasfang:
Title: feat: support nested struct sub-column recall and map selected-keys recall

### Purpose

Linked issue: #342 

This PR introduces end-to-end support for read-side projection improvements focused on two capabilities:

1. Nested struct sub-column recall
- Support reading partial nested fields from STRUCT/LIST/MAP hierarchies.
- Add recursive nested projection utilities at framework level, including schema-level detection for nested sub-field projection.
- Keep format-specific behavior in format layer:
  - Avro: fail fast for unsupported nested sub-field projection with explicit error.
  - Lance: add the same fail-fast protection for unsupported nested sub-field projection.

2. Map selected-keys recall
- Support MAP_SELECTED_KEYS metadata driven key filtering during read.
- Define and implement empty-string semantic for MAP_SELECTED_KEYS:
  - metadata key present with empty value means filter all keys.
  - empty tokens in comma-separated values are ignored.
- Add utility and integration coverage for selected-keys behavior.

Additional alignment/fixes included in this commit range:
- Refactor and consolidate nested projection detection logic into NestedProjectionUtils and simplify format readers.
- Parquet read path fix for out-of-order projected schema (align output to read schema order).
- Field mapping relaxations to avoid false mismatches for valid nested/projection scenarios.
- Extend integration and unit tests around nested pruning and map selected-keys.

### Tests

Unit tests:
- NestedProjectionUtilsTest.HasNestedSubfieldProjection_NoProjection
- NestedProjectionUtilsTest.HasNestedSubfieldProjection_WithProjection
- NestedProjectionUtils map selected-keys related tests, including:
  - GetMapSelectedKeys_EmptyString
  - GetMapSelectedKeys_ContainsEmptyToken
  - FilterMapArrayBySelectedKeys_EmptyKeyMeansFilterAll

Format tests:
- AvroFileBatchReaderTest.TestSetReadSchemaRejectNestedSubFieldProjection
- Compression/AvroFileFormatTest.TestNestedMap/*

Integration tests:
- FileFormats/NestedColumnPruningInteTest.MapSelectedKeysEmptyStringMeansFilterAll/*
- Nested column pruning scenarios for parquet/orc matrix (struct/deep nested/special fields/map selected-keys)

Note:
- Lance nested sub-field projection fail-fast test was added in source and should be executed in an environment built with PAIMON_ENABLE_LANCE=ON.

### API and Format

- API impact:
  - Read projection behavior is enhanced for nested recall and map selected-keys filtering.
  - SetReadSchema((const std::vector<std::string>&) rename to SetReadFieldNames(const std::vector<std::string>&)

- Storage/protocol impact:
  - No on-disk format or protocol change.
  - Behavior change is in reader projection and validation paths.

### Documentation

- No new external feature switch introduced.
- Existing behavior is clarified by tests for nested projection and map selected-keys semantics.

### Generative AI tooling


src/paimon/core/io/field_mapping_reader.cpp
```diff
@@ -142,9 +161,9 @@ Result<BatchReader::ReadBatchWithBitmap> FieldMappingReader::NextBatchWithBitmap
     // mapping non-partition array
     PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Array> casted_non_partition_array,
                            CastNonPartitionArrayIfNeed(non_partition_array));
```
lszskye, line 249: PR #355 handles selected-key filtering for regular MAP columns in FieldMappingReader, but it does not propagate paimon.map.selected-keys to a SharedShreddingFileReader placed below it.
The lower-level read schema is built from non_partition_data_schema, which preserves file-field metadata rather than metadata from non_partition_read_schema. Therefore, the selected-keys metadata may be lost, causing SharedShredding to read the entire MAP.
The selected-keys metadata should be copied from the corresponding read field when constructing the lower-level data field or read schema.

include/paimon/read_context.h
```diff
@@ -184,12 +201,27 @@ class PAIMON_EXPORT ReadContextBuilder {
...
+    /// Set the read Arrow Schema for nested column pruning.
+    ///
+    /// The read schema is an Arrow C Data Interface schema where STRUCT types
+    /// may contain only a subset of the original sub-fields, enabling nested column
+    /// pruning to reduce I/O. Field matching is based on field name: the system
+    /// looks up each field by name in the table schema and rebuilds the aligned
+    /// schema using the table schema's type and metadata. Any "paimon.id" metadata
+    /// in the user-provided schema is ignored. Other custom metadata (except
+    /// "paimon.id") is preserved and merged into the final aligned schema.
+    ///
```
lxy-9602, line 221: Consider modify comments for `preserved`.

lxy-9602: May add `paimon.map.selected-keys` usage or helper utils to generate proper read schema.

src/paimon/core/operation/read_context.cpp
```diff
@@ -59,7 +62,13 @@ ReadContext::ReadContext(
       cache_config_(cache_config),
       cache_(cache) {}
 
-ReadContext::~ReadContext() = default;
+ReadContext::~ReadContext() {}
+
```
lxy-9602, line 70: `~ReadContext` may need to release and delete `read_schema_`?

lxy-9602: Be aware of life cycle of `ArrowSchema*`

src/paimon/core/operation/read_context_test.cpp
```diff
@@ -86,7 +86,7 @@ TEST(ReadContextTest, TestSetContent) {
     ASSERT_EQ(ctx->GetPath(), "table_root_path");
     ASSERT_TRUE(ctx->GetMemoryPool());
     ASSERT_TRUE(ctx->GetExecutor());
-    ASSERT_EQ(ctx->GetReadSchema(), std::vector<std::string>({"f1", "f2"}));
+    ASSERT_EQ(ctx->GetReadFieldNames(), std::vector<std::string>({"f1", "f2"}));
     ASSERT_EQ(ctx->GetReadFieldIds(), std::vector<int32_t>({0, 1}));
     ASSERT_EQ(*predicate, *(ctx->GetPredicate()));
```
lxy-9602, line 93: Please add tests for `SetReadSchema` and `HasReadSchema`.

src/paimon/core/operation/internal_read_context.cpp
```diff
@@ -16,19 +16,169 @@
...
+                    fmt::format("Read schema nested field '{}' does not exist in table field '{}'",
+                                read_child->name(), table_field->name()));
+            }
+            PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Field> rebased_child,
+                                   AlignReadFieldWithTableFieldIds(read_child, table_child));
+            rebased_children.push_back(rebased_child);
+        }
+        auto rebased_type = arrow::struct_(rebased_children);
+        auto aligned_field = table_field->WithType(rebased_type)->WithName(read_field->name());
+        return MergeReadFieldMetadataByWhitelist(aligned_field, read_field, kReadMetadataWhitelist);
```
lxy-9602: It looks like `WithName(...)` is called here. Does that mean `aligned_field` will also be renamed?

src/paimon/core/operation/internal_read_context_test.cpp
```diff
@@ -191,4 +192,112 @@ TEST(InternalReadContext, TestReadWithFieldIdsAndSpecialFields) {
...
+    ASSERT_OK_AND_ASSIGN(
+        auto internal_context,
+        InternalReadContext::Create(read_context, table_schema, table_schema->Options()));
+
+    std::vector<DataField> expected_fields = {
+        DataField(3, arrow::field("f3", arrow::float64())),
+        DataField(0, arrow::field("f0", arrow::utf8())),
+    };
+    auto expected_schema = DataField::ConvertDataFieldsToArrowSchema(expected_fields);
+    ASSERT_TRUE(internal_context->GetReadSchema()->Equals(expected_schema));
```
lxy-9602: If you want to check metadata, use `schema->Equals(other_schema, /*check_meta=*/true)`. Otherwise default `check_meta` is false.

src/paimon/core/utils/field_mapping.h
```diff
@@ -80,14 +80,18 @@ class FieldMappingBuilder {
...
 
     Result<NonPartitionInfo> CreateNonPartitionInfo(
         const std::vector<DataField>& data_fields, const ExistFieldInfo& exist_field_info,
         const std::map<std::string, int32_t>& partition_keys) const;
     Result<std::shared_ptr<Predicate>> CreateDataFilters(
         const std::vector<DataField>& data_fields) const;
 
+    DataField MergeFieldMetadataByWhitelist(
+        const DataField& data_field, const DataField& read_field,
+        const std::vector<std::string>& metadata_keys_whitelist) const;
```
lxy-9602: Could we add a static utility for `MergeFieldMetadataByWhitelist` in `DataField` to remove the duplicated code in `FieldMapping` and `InternalReadContext`?

include/paimon/read_context.h
```diff
@@ -173,9 +190,9 @@ class PAIMON_EXPORT ReadContextBuilder {
...
     /// @param read_field_names Vector of field names to read from the table.
     /// @return Reference to this builder for method chaining.
-    /// @note Currently supports top-level field selection. Future versions may support
-    ///       nested field selection using ArrowSchema for more granular projection
-    ReadContextBuilder& SetReadSchema(const std::vector<std::string>& read_field_names);
+    /// @note Currently supports top-level field selection. For nested field selection
+    ///       use SetReadSchema(ArrowSchema*) instead.
+    ReadContextBuilder& SetReadFieldNames(const std::vector<std::string>& read_field_names);
     /// Set the schema fields to read from the table.
     ///
```
lxy-9602, line 199: `SetReadSchema(const std::vector<std::string>&)` is part of the exported `ReadContextBuilder` API, but this PR renames it to `SetReadFieldNames(...)`. That breaks existing callers that use top-level projection, while the PR states there is no incompatible public API removal. Please update pr description.

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,289 @@
...
+        return result;
+    }
+    std::string value = get_result.ValueUnsafe();
+    StringUtils::Trim(&value);
+    if (value.empty()) {
+        // Metadata is explicitly present but empty: select the empty-string key.
+        result.push_back("");
+        return result;
+    }
+
```
lxy-9602, line 257: Why we need trim? In pr #376 , it seems only do `Split` with not ignore empty.

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+Result<std::shared_ptr<arrow::Array>> NestedProjectionUtils::FilterMapArrayBySelectedKeys(
+    const std::shared_ptr<arrow::Array>& array, const std::vector<std::string>& selected_keys) {
+    if (selected_keys.empty() || !array || array->length() == 0) {
+        return array;
+    }
+
+    auto map_array = std::static_pointer_cast<arrow::MapArray>(array);
+    auto map_type = std::static_pointer_cast<arrow::MapType>(array->type());
```
lxy-9602, line 360: assert(map_array && map_type)

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,289 @@
...
+    int64_t num_maps = map_array->length();
+
+    std::unordered_set<std::string> deduplicated;
+    deduplicated.reserve(selected_keys.size());
+    for (const auto& selected_key : selected_keys) {
+        if (!deduplicated.insert(selected_key).second) {
+            return Status::Invalid(fmt::format("Duplicate selected key '{}' in {} metadata",
+                                               selected_key, DataField::MAP_SELECTED_KEYS));
+        }
+    }
```
lxy-9602, line 376: `GetMapSelectedKeys` already handle deduplicated `selected_key`, do we need to check again here?

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+    auto key_builder = std::make_shared<arrow::StringBuilder>();
+    PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+        std::unique_ptr<arrow::ArrayBuilder> value_builder_u,
+        arrow::MakeBuilder(values_array->type(), arrow::default_memory_pool()));
+    auto value_builder = std::shared_ptr<arrow::ArrayBuilder>(std::move(value_builder_u));
+    arrow::MapBuilder map_builder(arrow::default_memory_pool(), key_builder, value_builder);
+
```
lxy-9602, line 386: `arrow::default_memory_pool()`? Add a input param for pool.

lxy-9602: Also I suggest to make a MapBuilder first and then get key and value sub builder.

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,289 @@
...
+                if (key_view == selected_key) {
+                    PAIMON_RETURN_NOT_OK_FROM_ARROW(key_builder->Append(
+                        key_view.data(), static_cast<int32_t>(key_view.size())));
+                    PAIMON_RETURN_NOT_OK_FROM_ARROW(
+                        value_builder->AppendArraySlice(*values_array->data(), entry_idx, 1));
+                }
+            }
+        }
+    }
+
```
lxy-9602, line 410: Please also add tests for orc dictionary array. Both map key and value.

src/paimon/core/utils/nested_projection_utils_test.cpp
```diff
@@ -0,0 +1,405 @@
...
+    // First map: should have "a" and "c"
+    ASSERT_EQ(result->value_length(0), 2);
+    auto keys0 = std::static_pointer_cast<arrow::StringArray>(result->keys());
+    ASSERT_EQ(keys0->GetString(result->value_offset(0)), "a");
+    ASSERT_EQ(keys0->GetString(result->value_offset(0) + 1), "c");
+
+    // Second map: should have only "a"
+    ASSERT_EQ(result->value_length(1), 1);
+    ASSERT_EQ(keys0->GetString(result->value_offset(1)), "a");
+}
```
lxy-9602, line 365: Could we use `array->Equals(...)` in the test here for better readability?

src/paimon/core/utils/nested_projection_utils_test.cpp
```diff
@@ -0,0 +1,405 @@
...
+    ASSERT_EQ(values->Value(result->value_offset(0) + 1), 1);
+}
+
+TEST(NestedProjectionUtilsTest, FilterMapArrayBySelectedKeys_DuplicateSelectedKeys) {
+    auto map_array = BuildStringInt32MapArray({{{"a", 1}, {"b", 2}}});
+    std::vector<std::string> selected = {"a", "a"};
+
+    auto result = NestedProjectionUtils::FilterMapArrayBySelectedKeys(map_array, selected);
+    ASSERT_FALSE(result.ok());
+}
```
lxy-9602, line 506: WITH_ERROR_MSG

src/paimon/core/io/field_mapping_reader.cpp
```diff
@@ -74,6 +76,23 @@ FieldMappingReader::FieldMappingReader(int32_t field_count,
...
+        // FilterMapArrayBySelectedKeys can filter out unwanted entries.
+        if (!need_mapping_ &&
+            non_partition_info_.non_partition_read_schema[i].Type()->id() == arrow::Type::MAP) {
+            auto selected_keys_or = NestedProjectionUtils::GetMapSelectedKeys(
+                non_partition_info_.non_partition_read_schema[i].ArrowField());
+            if (!selected_keys_or.ok()) {
+                // Keep mapping enabled so the parse error can be surfaced in
+                // MappingFields where Status can be returned.
+                need_mapping_ = true;
+                continue;
```
lxy-9602: Why do we delay the error until the actual I/O read happens? Shouldn’t this be validated and reported during construction instead?

lxy-9602: May add `Create` for `FieldMappingReader`

src/paimon/format/parquet/file_reader_wrapper.h
```diff
@@ -122,6 +122,14 @@ class FileReaderWrapper {
...
     Status ApplyReadRanges(const std::vector<std::pair<uint64_t, uint64_t>>& read_ranges);
 
+    /// Set the read schema for page-filtered reading. When nested column pruning
+    /// is used, the leaf-column-name-based schema inference in PrepareForReading
+    /// cannot correctly reconstruct nested types. This setter allows the caller
+    /// to provide the correct pruned schema directly.
+    void SetReadSchemaForPageFilter(const std::shared_ptr<arrow::Schema>& schema) {
+        external_read_schema_ = schema;
+    }
+
```
lxy-9602: Please refer to #371: when the read schema contains nested types, we should disable page-level filtering for now, since the current page-filtering logic has issues with offset calculation for nested pages.

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -428,4 +436,89 @@ Result<::parquet::ArrowReaderProperties> ParquetFileBatchReader::CreateArrowRead
...
+Status ParquetFileBatchReader::CollectLeafIndices(const std::shared_ptr<arrow::DataType>& read_type,
+                                                  const std::shared_ptr<arrow::DataType>& file_type,
+                                                  int32_t* leaf_index,
+                                                  std::vector<int32_t>* indices) {
+    if (file_type->id() == arrow::Type::STRUCT) {
+        for (const auto& file_child : file_type->fields()) {
+            std::shared_ptr<arrow::Field> read_child = nullptr;
+            for (const auto& candidate : read_type->fields()) {
+                if (candidate->name() == file_child->name()) {
+                    read_child = candidate;
```
lxy-9602, line 453: Could this cause issues if child columns have the same name, for example `a.sub_b` and `c.sub_b`, and we compare them by name only?

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -428,4 +436,89 @@ Result<::parquet::ArrowReaderProperties> ParquetFileBatchReader::CreateArrowRead
...
+        int32_t file_field_idx = -1;
+        for (int32_t i = 0; i < static_cast<int32_t>(file_fields.size()); ++i) {
+            if (file_fields[i]->name() == read_field->name()) {
+                file_field_idx = i;
+                break;
+            }
+        }
+        if (file_field_idx < 0) {
+            continue;
+        }
```
lxy-9602, line 521: In what cases could `read_field` fail to be found in `file_field`? Shouldn’t that be treated as an error?

src/paimon/format/parquet/parquet_file_batch_reader_test.cpp
```diff
@@ -281,6 +282,36 @@ TEST_F(ParquetFileBatchReaderTest, TestSetReadSchema) {
...
+                    arrow::struct_(read_schema->fields()), {R"([
+        ["Lucy", 1, 14.1]
+    ])"},
+                    &expected_array)
+                    .ok());
+    ASSERT_TRUE(result_array->Equals(expected_array))
+        << "expected: " << expected_array->ToString() << "\nactual: " << result_array->ToString();
+}
+
 TEST_F(ParquetFileBatchReaderTest, TestNextBatchSimple) {
```
lxy-9602, line 442: Please add a test in the Parquet reader for recalling child columns from a nested struct type, similar to the ORC case.

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -127,24 +127,26 @@ Status ParquetFileBatchReader::SetReadSchema(
...
                                           arrow::ImportSchema(schema));
 
-        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Schema> file_schema, reader_->GetSchema());
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Schema> raw_file_schema,
+                               reader_->GetSchema());
+        // Convert PARQUET:field_id to paimon.id so that nested column matching works.
+        PAIMON_ASSIGN_OR_RAISE(
+            std::shared_ptr<arrow::Schema> file_schema,
+            ParquetFieldIdConverter::GetPaimonIdsFromParquetIds(raw_file_schema));
+
```
lxy-9602, line 143: `GetPaimonIdsFromParquetIds(raw_file_schema)` converts Parquet field ids into `paimon.id` metadata, but the following nested column index computation still matches fields by name in `ComputeNestedColumnIndices` / `CollectLeafIndices`. If the downstream matching remains name-based, this conversion seems uncessary.

src/paimon/core/utils/nested_projection_utils.cpp
lucasfang: It looks like a redundant check from the current main call chain, but keeping it is valuable for function robustness.

src/paimon/format/parquet/parquet_file_batch_reader.cpp
lucasfang: This matching is performed within the scope of the current struct’s children, not as a global lookup.
During recursion, it first descends into a’s subtree and then into c’s subtree, so sub_b under a will not be confused with sub_b under c.

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,289 @@
...
+            const auto& data_map = static_cast<const arrow::MapType&>(*data_type);
+            PAIMON_ASSIGN_OR_RAISE(std::optional<std::shared_ptr<arrow::DataType>> pruned_key,
+                                   PruneDataType(read_map.key_type(), data_map.key_type()));
+            PAIMON_ASSIGN_OR_RAISE(std::optional<std::shared_ptr<arrow::DataType>> pruned_value,
+                                   PruneDataType(read_map.item_type(), data_map.item_type()));
+            if (!pruned_key.has_value() || !pruned_value.has_value()) {
+                return std::optional<std::shared_ptr<arrow::DataType>>(std::nullopt);
+            }
+            std::shared_ptr<arrow::DataType> result_type = arrow::map(
+                pruned_key.value(), pruned_value.value(), data_map.key_field()->nullable());
```
lszskye: ARROW_EXPORT
std::shared_ptr<DataType> map(std::shared_ptr<DataType> key_type,
                              std::shared_ptr<DataType> item_type,
                              bool keys_sorted = false);

ARROW_EXPORT
std::shared_ptr<DataType> map(std::shared_ptr<DataType> key_type,
                              std::shared_ptr<Field> item_field,
                              bool keys_sorted = false);

the third param in `map` function is `keys_sorted`, but pass data_map.key_field()->nullable()

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// Map selected-keys support
+
+Result<std::vector<std::string>> NestedProjectionUtils::GetMapSelectedKeys(
+    const std::shared_ptr<arrow::Field>& field) {
+    std::vector<std::string> result;
+    if (!field || !field->HasMetadata() || !field->metadata()) {
+        return result;
+    }
```
lszskye, line 246: Under what circumstances would the field not exist(`!field`)? In such cases, shouldn't we avoid entering this function altogether and instead return early at the call site?

src/paimon/core/utils/nested_projection_utils.cpp
lszskye, line 360: assert or use `checked_pointer_cast`, here is the definition of `checked_pointer_cast`
std::shared_ptr<T> checked_pointer_cast(std::shared_ptr<U> r) noexcept {
#ifdef NDEBUG
  return std::static_pointer_cast<T>(std::move(r));
#else
  return std::dynamic_pointer_cast<T>(std::move(r));
#endif
}

src/paimon/core/utils/nested_projection_utils.cpp
lszskye, line 386: PAIMON_RETURN_NOT_OK_FROM_ARROW(map_builder->Reserve(row_count));

src/paimon/core/utils/nested_projection_utils.cpp
lucasfang: In the current main code path, !field is unlikely to happen, and your point is valid. But this check makes the API more robust.

src/paimon/core/utils/field_mapping.cpp
```diff
@@ -101,8 +102,22 @@ ExistFieldInfo FieldMappingBuilder::CreateExistFieldInfo(
...
+                continue;
+            }
+
+            DataField pruned_data_field(data_field.Id(),
+                                        data_field.ArrowField()->WithType(pruned_type.value()),
+                                        data_field.Description());
             exist_field_info.exist_read_schema.push_back(read_field);
-            exist_field_info.exist_data_schema.push_back(data_field);
+            exist_field_info.exist_data_schema.push_back(pruned_data_field);
             exist_field_info.idx_in_target_read_schema.push_back(target_idx);
```
lxy-9602, line 121: If a struct really has no child columns recalled, what does the corresponding field return now — null, [], an error, or something else?

src/paimon/core/utils/field_mapping.cpp
```diff
@@ -146,7 +161,11 @@ Result<std::vector<std::shared_ptr<CastExecutor>>> FieldMappingBuilder::CreateDa
         if (!read_fields[i].Type()->Equals(data_fields[i].Type())) {
             if (read_type == FieldType::MAP || read_type == FieldType::ARRAY ||
                 read_type == FieldType::STRUCT) {
-                return Status::Invalid("Only support column type evolution in atomic data type.");
+                // Nested types may differ due to nested column pruning (different
+                // number of sub-fields). No cast is needed — type pruning is
+                // handled by PruneDataType during field mapping construction.
+                cast_executors.push_back(nullptr);
+                continue;
```
lxy-9602, line 180: Would this cause data generated by Java that support nested-type schema evolution to no longer fail in C++ as before, but instead return incorrect results silently?

lxy-9602: We may need to validate this field by field to ensure the types are consistent. Also, I wanted to confirm: Java currently support schema evolution for child fields inside a struct, including renaming and type changes. If a child field is renamed, can it still be handled correctly now?

src/paimon/format/parquet/parquet_file_batch_reader.cpp
lxy-9602, line 181: I’m not quite following why `file_reader_wrapper.cpp` needs to be changed here. With the current logic, if `has_nested_field` is true, it naturally won’t go through the page index filter. Would it make sense to remove the changes in the other places instead?

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -202,6 +201,12 @@ Status ParquetFileBatchReader::SetReadSchema(
 
         read_data_type_ = arrow::struct_(read_schema->fields());
 
+        // Provide the read schema to FileReaderWrapper for page-filtered reading.
+        // This is needed because nested column pruning produces leaf column indices
+        // whose names don't correspond to top-level Arrow fields, so the wrapper
+        // cannot infer the correct schema from leaf column names alone.
+        reader_->SetReadSchemaForPageFilter(read_schema);
+
```
lxy-9602: Is that really necessary? Since `SetReadSchemaForPageFilter` shouldn't be invoked for page-filtered reads in the first place (nest fields recalled), perhaps we don't need this change.

src/paimon/core/io/field_mapping_reader.cpp
```diff
@@ -283,20 +314,33 @@ Result<std::shared_ptr<arrow::Array>> FieldMappingReader::GenerateNonExistArray(
...
-                                       const std::vector<DataField>& read_fields_of_data_array,
-                                       const std::vector<int32_t>& idx_in_target_schema,
-                                       arrow::ArrayVector* target_array,
-                                       std::vector<std::string>* target_field_names) {
+Status FieldMappingReader::MappingFields(const std::shared_ptr<arrow::Array>& data_array,
+                                         const std::vector<DataField>& read_fields_of_data_array,
+                                         const std::vector<int32_t>& idx_in_target_schema,
+                                         arrow::ArrayVector* target_array,
+                                         std::vector<std::string>* target_field_names) {
     auto* struct_array = arrow::internal::checked_cast<arrow::StructArray*>(data_array.get());
```
lxy-9602, line 397: Since `FieldMappingBuilder::PruneDataType` maps the requested nested field back to the data-file field by `paimon.id`, passing the old data-field name into the Parquet reader makes name-based leaf selection work for old files. However, after the file reader returns a StructArray with the old nested child names, `FieldMappingReader` only rewrites top-level field names. If Java supports nested sub-field rename, should we also recursively rebase nested StructArray field names/types from the data schema back to the read schema before returning the final batch?

src/paimon/core/utils/nested_projection_utils.h
```diff
@@ -0,0 +1,114 @@
...
+    NestedProjectionUtils() = delete;
+    ~NestedProjectionUtils() = delete;
+
+    static std::shared_ptr<arrow::Field> FindFieldByName(const arrow::FieldVector& fields,
+                                                         const std::string& name);
+
+    /// Extract the paimon field ID from an Arrow field's metadata ("paimon.id").
+    /// Returns -1 if the metadata key is not present.
+    static int32_t GetPaimonFieldId(const std::shared_ptr<arrow::Field>& field) {
+        if (!field || !field->HasMetadata() || !field->metadata()) {
```
lxy-9602: Please mv implementation of `GetPaimonFieldId` and `FindFieldByPaimonId` in `.cpp`

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,346 @@
...
+        }
+    }
+    return nullptr;
+}
+
+Result<bool> NestedProjectionUtils::HasNestedSubfieldProjectionType(
+    const std::shared_ptr<arrow::DataType>& file_type,
+    const std::shared_ptr<arrow::DataType>& read_type) {
+    if (file_type->id() != read_type->id()) {
+        return false;
```
lxy-9602: Is this function used in the format reader? In what scenarios would `file_type` and `read_type` be different? Should this return error status or just `false`?

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,346 @@
...
+            auto read_struct = std::static_pointer_cast<arrow::StructType>(read_type);
+            if (read_struct->num_fields() != file_struct->num_fields()) {
+                return true;
+            }
+            for (const auto& read_child : read_struct->fields()) {
+                auto file_child = FindFieldByName(file_struct->fields(), read_child->name());
+                if (!file_child) {
+                    return true;
+                }
+                PAIMON_ASSIGN_OR_RAISE(
```
lxy-9602, line 93: Also, in a case where the query requests a field that doesn’t exist, should this return `true` or a bad status?

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,346 @@
...
+    }
+}
+
+Result<bool> NestedProjectionUtils::HasNestedSubfieldProjection(
+    const std::shared_ptr<arrow::Schema>& file_schema,
+    const std::shared_ptr<arrow::Schema>& read_schema) {
+    for (const auto& read_field : read_schema->fields()) {
+        auto file_field = file_schema->GetFieldByName(read_field->name());
+        if (!file_field) {
+            continue;
```
lxy-9602: A similar issue here.

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,346 @@
...
+        int64_t start = map_array->value_offset(map_idx);
+        int64_t end = map_array->value_offset(map_idx + 1);
+
+        // Keep selected keys in the exact selected_keys order.
+        for (const auto& selected_key : selected_keys) {
+            for (int64_t entry_idx = start; entry_idx < end; ++entry_idx) {
+                std::string_view key_view;
+                if (string_keys) {
+                    if (string_keys->IsNull(entry_idx)) {
+                        continue;
```
lxy-9602: `FilterMapArrayBySelectedKeys` looks a bit too long. Would it be ok to extract the dictionary-related logic into a separate function?

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,346 @@
...
+                    }
+                    key_view = string_keys->GetView(entry_idx);
+                } else {
+                    if (dict_keys->IsNull(entry_idx)) {
+                        continue;
```
lxy-9602: In arrow map key can never be null

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -0,0 +1,346 @@
...
+                if (!data_child) {
+                    // Schema Evolution: field not present in data, skip.
+                    continue;
+                }
+                PAIMON_ASSIGN_OR_RAISE(
+                    std::optional<std::shared_ptr<arrow::DataType>> pruned_child_type,
+                    PruneDataType(read_child->type(), data_child->type()));
+                if (!pruned_child_type.has_value()) {
+                    // All sub-fields of this child were pruned away; skip it.
+                    continue;
```
lxy-9602, line 180: In pruning, if just one child field is missing, does the entire field get dropped? For example, with `struct<f1:struct<s1, s2:struct<x,y>>>`, if x is not found, does that mean the whole `f1` gets dropped?

src/paimon/core/utils/nested_projection_utils_test.cpp
```diff
@@ -0,0 +1,457 @@
...
+    arrow::StringDictionaryBuilder dict_builder(arrow::default_memory_pool());
+    for (int64_t i = 0; i < string_keys->length(); ++i) {
+        ASSERT_TRUE(dict_builder.Append(string_keys->GetView(i)).ok());
+    }
+    auto dict_keys_result = dict_builder.Finish();
+    ASSERT_TRUE(dict_keys_result.ok());
+    auto dict_keys = dict_keys_result.ValueUnsafe();
+
+    auto dict_map_type = arrow::map(dict_keys->type(), map->items()->type());
+    auto dict_map_array = std::make_shared<arrow::MapArray>(
```
lxy-9602, line 459: @lszskye Have you tested the case where the key is dictionary-encoded? It looks like this PR adds some extra handling specifically for that path.
@lucasfang Could you also help confirm whether the case where the value is dictionary-encoded works correctly?

test/inte/nested_column_pruning_inte_test.cpp
```diff
@@ -0,0 +1,1145 @@
...
+
+    arrow::EqualOptions equal_options = arrow::EqualOptions::Defaults();
+    bool is_equal = expected_chunked->Equals(read_result, equal_options.diff_sink(&std::cout));
+    if (!is_equal) {
+        std::cout << "[expected_type] " << expected_chunked->type()->ToString() << std::endl;
+        std::cout << "[actual_type]   " << read_result->type()->ToString() << std::endl;
+        std::cout << "[expected] " << expected_chunked->ToString() << std::endl;
+        std::cout << "[actual]   " << read_result->ToString() << std::endl;
+    }
+    ASSERT_TRUE(is_equal);
```
lxy-9602: Could we extract some common helper functions for the end-to-end tests? There’s quite a bit of duplicated logic across tests in the same file, such as the `if (!is_equal) output`, the reading logic, and the logic for filling in `_VALUE_KIND` from the projected schema.

test/inte/nested_column_pruning_inte_test.cpp
```diff
@@ -0,0 +1,1145 @@
...
+
+    ArrowSchema c_schema;
+    ASSERT_TRUE(arrow::ExportSchema(*projected_schema, &c_schema).ok());
+
+    ReadContextBuilder read_context_builder(table_path_);
+    read_context_builder.SetOptions(options).SetReadSchema(&c_schema);
+    ASSERT_OK_AND_ASSIGN(auto read_context, read_context_builder.Finish());
+    ASSERT_OK_AND_ASSIGN(auto table_read, TableRead::Create(std::move(read_context)));
+    auto batch_reader_result = table_read->CreateReader(data_splits);
+    if (!batch_reader_result.ok()) {
```
lxy-9602: So is projecting the child fields of a struct inside a list supposed to work or not? From the code, it seems ambiguous — it looks like it could be supported, but also possibly not.

test/inte/nested_column_pruning_inte_test.cpp
```diff
@@ -0,0 +1,1145 @@
...
+        dir_.reset();
+    }
+
+ protected:
+    std::string file_format_;
+    std::string test_dir_;
+    std::string table_path_;
+    std::unique_ptr<UniqueTestDirectory> dir_;
+};
+
```
lxy-9602, line 89: Could we add a few more integration coverage cases here?

1. Null nested values: e.g. rows where the projected struct/map itself is null, or a nested child is null, to ensure pruning and map selected-keys filtering preserve null semantics correctly.

2. Schema evolution error path: e.g. old data file schema and current read schema diverge for nested fields in a way that should fail, so we verify the error is reported during reader construction/read instead of silently skipping the field.  Since paimon-cpp does not seem to expose a schema-evolution API yet, one concrete way is to create the table and write data with the old schema first, then directly write a new schema-2 file, and read the table with the new schema.

3. Nested map selected-keys: currently selected-keys coverage appears to focus on top-level MAP fields. It would be useful to cover metadata on a map nested inside a STRUCT.

test/inte/nested_column_pruning_inte_test.cpp
```diff
@@ -0,0 +1,1145 @@
...
+    }
+
+    auto read_result_result = ReadResultCollector::CollectResult(batch_reader_result.value().get());
+    ASSERT_FALSE(read_result_result.ok());
+    auto message = read_result_result.status().ToString();
+    ASSERT_TRUE(message.find("partial projection inside list/map") != std::string::npos ||
+                message.find("type mismatch") != std::string::npos)
+        << "unexpected error: " << message;
+}
+
```
lszskye, line 1642: add a case for STRUCT<MAP<STRING, INT32>> , read with selected keys filter.

src/paimon/core/operation/internal_read_context.cpp
```diff
@@ -54,9 +54,10 @@ Result<std::shared_ptr<arrow::Field>> InternalReadContext::AlignReadFieldWithTab
             auto table_child =
                 NestedProjectionUtils::FindFieldByName(table_struct->fields(), read_child->name());
             if (!table_child) {
-                return Status::Invalid(
-                    fmt::format("Read schema nested field '{}' does not exist in table field '{}'",
-                                read_child->name(), table_field->name()));
+                // Keep missing children in aligned read schema. They will be
+                // materialized as nulls during field mapping.
+                rebased_children.push_back(read_child);
+                continue;
```
lxy-9602: Why doesn't it returns an error when a nested column is missing from the table schema? Shouldn't all columns in the read schema be present in the latest table schema?

src/paimon/core/utils/nested_projection_utils.cpp
lxy-9602, line 186: Under what scenario can `pruned_fields.empty()` happen? Does this correspond to the case where no child fields of the nested column are projected? If so, is the outer struct removed afterward? Throwing an error seems more appropriate here; otherwise users may get a result array, only to have `FindFieldByName` fail later.

src/paimon/core/utils/nested_projection_utils_test.cpp
```diff
@@ -108,12 +108,15 @@ TEST(NestedProjectionUtilsTest, PruneDataTypeStructPruneSubset) {
...
     auto data_type = arrow::struct_({MakeField("x", arrow::int32(), 1)});
     auto read_type = arrow::struct_({MakeField("y", arrow::int32(), 99)});
 
-    ASSERT_OK_AND_ASSIGN(auto result, NestedProjectionUtils::PruneDataType(read_type, data_type));
-    ASSERT_FALSE(result.has_value());
+    auto result = NestedProjectionUtils::PruneDataType(read_type, data_type);
+    ASSERT_FALSE(result.ok());
+    ASSERT_TRUE(result.status().ToString().find(
+                    "does not support schema evolution inside struct") != std::string::npos)
+        << result.status().ToString();
```
lxy-9602: ASSERT_NOK_WITH_MSG

src/paimon/core/utils/nested_projection_utils_test.cpp
```diff
@@ -145,12 +148,11 @@ TEST(NestedProjectionUtilsTest, PruneDataTypeListWithStructElement) {
...
-    ASSERT_TRUE(result.has_value());
-    auto list_type = std::dynamic_pointer_cast<arrow::ListType>(result.value());
-    ASSERT_NE(list_type, nullptr);
-    ASSERT_EQ(list_type->value_type()->num_fields(), 1);
-    ASSERT_EQ(list_type->value_type()->field(0)->name(), "a");
+    auto result = NestedProjectionUtils::PruneDataType(read_type, data_type);
+    ASSERT_FALSE(result.ok());
+    ASSERT_TRUE(result.status().ToString().find("partial projection inside list") !=
+                std::string::npos)
+        << result.status().ToString();
```
lxy-9602: ASSERT_NOK_WITH_MSG

test/inte/nested_column_pruning_inte_test.cpp
```diff
@@ -249,18 +257,89 @@ TEST_P(NestedColumnPruningInteTest, PruneSameNestedFieldNameFromDifferentStructC
...
     }
-    ASSERT_TRUE(is_equal);
+
+    auto read_result_result = ReadResultCollector::CollectResult(batch_reader_result.value().get());
+    ASSERT_FALSE(read_result_result.ok());
+    auto message = read_result_result.status().ToString();
+    ASSERT_TRUE(message.find("does not support schema evolution inside struct") !=
+                    std::string::npos ||
+                message.find("requires paimon.id for nested struct field") != std::string::npos)
+        << "unexpected error: " << message;
```
lxy-9602: I’d like to confirm whether the error is expected to be raised in `CreateReader` or in `NextBatch`. If we can confirm that, we probably shouldn’t assert both. Also, please check all related code for `ASSERT_NOK_WITH_MSG`.

test/inte/nested_column_pruning_inte_test.cpp
```diff
@@ -249,18 +257,89 @@ TEST_P(NestedColumnPruningInteTest, PruneSameNestedFieldNameFromDifferentStructC
...
 }
 
-// Test: Querying struct sub-fields with a non-existent nested field should fail fast.
+// Test: Querying struct sub-fields with partial non-existent nested fields keeps
+// existing nested fields and materializes missing ones as null.
 TEST_P(NestedColumnPruningInteTest, QueryStructSubFieldsWithNonExistentField) {
```
lxy-9602, line 401: The comment seems inconsistent with the code. Shouldn’t this always fail?

src/paimon/core/utils/field_mapping.cpp
lucasfang: NULL

lxy-9602 (approved):
+1

PR #356, opened 2026-06-10
feat: Optimize PK MOR read performance by reduce per-row object allocation and improve merge read performance.

### LINKED ISSUE #321: [Feature] optimize count(*) for PK table
### Search before asking

### Motivation

Currently, executing count(*) on a PK table in paimon-cpp requires reading all rows through the standard CreateReader path, which involves full deserialization of all columns. Even when only the row count is needed, the system must perform complete I/O and decoding for every column — this is highly inefficient for large tables.

### Solution

The core optimization idea is column pruning — only read PK-related fields needed for the merge, and skip the row-to-column conversion overhead entirely.

### Anything else?

_No response_

- [ ] I'm willing to submit a PR!

dalingmeng:
### Purpose

Linked issue: close #321 

Optimize PK MOR read performance by reducing per-row overhead in the merge-on-read path.

This change improves the key-value data file reader and merge functions to avoid unnecessary
object allocation and repeated checks during row iteration. It also adds a fast path when all rows
in a batch are selected, reducing bitmap lookup overhead in common scan scenarios.

### Tests

GenericRowTest.TestResetFields

### Generative AI tooling

Generated-by: Aone Copilot, ide-idealab/claude4.6-opus


lxy-9602 (commented):
+1

src/paimon/core/io/key_value_data_file_record_reader.h
```diff
@@ -55,7 +55,10 @@ class KeyValueDataFileRecordReader : public KeyValueRecordReader {
     class Iterator : public KeyValueRecordReader::Iterator {
      public:
         Iterator(KeyValueDataFileRecordReader* reader, int64_t previous_batch_first_row_number)
-            : previous_batch_first_row_number_(previous_batch_first_row_number), reader_(reader) {}
```
lucasfang, line 58: why adding all_selected_？

dalingmeng: `all_selected_` is used to avoid calling `selection_bitmap_.Cardinality()` every time in `HasNext()`.

`Cardinality()` is not just a simple field access. Internally, CRoaring iterates over roaring containers and sums each container's cardinality. So we cache the result once per iterator/batch to avoid repeated container traversal in the hot path.

lucasfang: maybe just cache selection_bitmap_.Cardinality()  for better readability? otherwise all_selected is vague

dalingmeng: Replaced `all_selected_` with `selection_cardinality_` to cache the cardinality value directly, as suggested

lxy-9602 (commented):
+1

lxy-9602 (approved):

PR #360, opened 2026-06-11
feat(shared-shredding): support nested field projection in ORC format reader

lszskye:
## Purpose
Linked issue: #342 
Support nested struct sub-field projection in ORC format reader. Previously, reading nested struct types required selecting all sub-fields; now users can read only specific sub-fields of a nested struct (e.g., read only `col1.sub1` and `col1.sub3` from `struct<col1:struct<sub1:int,sub2:double,sub3:string>>`).

Nested field predicate pushdown is **not** included in this change.


lxy-9602 (commented):
+1

lxy-9602 (approved):
+1

PR #361, opened 2026-06-11
fix: make byte hashing independent of char signedness

wgtmac:
### Purpose

Fix ARM bucket mismatches caused by platform-dependent `char` signedness.

This change makes byte hashing explicit:
- Hive `TINYINT` hashing now treats row bytes as signed Java bytes.
- Hive string/binary byte-array hashing now treats high-bit bytes as signed Java bytes.
- Murmur tail-byte hashing now reads tail bytes as `int8_t`, avoiding unsigned plain-`char` drift.

### Tests

Pass existing CI.

### API and Format

N/A

### Documentation

N/A

### Generative AI tooling

Generated-by: OpenAI Codex

### DIFF
--- a/src/paimon/common/utils/murmurhash_utils.h
+++ b/src/paimon/common/utils/murmurhash_utils.h
@@ -182,7 +182,9 @@ class MurmurHashUtils {
         int32_t length_aligned = length_in_bytes - length_in_bytes % 4;
         int32_t h1 = HashBytesByInt(segment, offset, length_aligned, seed);
         for (int32_t i = length_aligned; i < length_in_bytes; i++) {
-            int32_t k1 = MixK1(segment.Get(offset + i));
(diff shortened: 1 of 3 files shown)


zjw1111:
Thanks for the fix! Could you add a compatibility test that compares the results with Java, covering negative tinyint values, high-byte UTF-8/binary data, and the Murmur unaligned tail-byte case?

wgtmac:
> Thanks for the fix! Could you add a compatibility test that compares the results with Java, covering negative tinyint values, high-byte UTF-8/binary data, and the Murmur unaligned tail-byte case?

I would love to do this. But unfortunately I cannot build and run the test on my macOS laptop and I do not yet have access to the Java code. What I tried is to make sure it does not break existing CI and work as expected on our internal ARM environment.

zjw1111 (approved):
+1

PR #362, opened 2026-06-11
feat(schema): Shrink PK table to unsupported schema evolution

gripleaf:
### Purpose

When unrelated table-level properties are tweaked on a Paimon PK table, a new schema version is generated. The new and old schemas are read-compatible (field shape, ids and PK layout do not change), but paimon-cpp currently rejects any non-zero schema id on PK tables with `do not support schema evolution in pk table while scan process`, because full PK schema evolution is not implemented.

This PR narrows that restriction. It does NOT add general PK schema-evolution support; instead it lets the scan path accept multiple schema versions on a PK table when those versions differ only in option keys that the user has explicitly whitelisted as safe to ignore.

Key changes:

- New public option `scan.pk.schema-compatibility.ignored-options` in  `include/paimon/defs.h`, a comma-separated list of regex patterns  full-matched against option keys allowed to differ across schema versions.  Default is empty (strict, same behavior as before).
- `TableSchema::IsCompatibleForPKScan` performs a conservative blacklist  comparison: fields, field ids, `highestFieldId`, partition / primary /  bucket keys and `numBucket` must be identical, and every option key must  be equal unless it matches one of the user-supplied regex patterns.  Invalid regex patterns are silently dropped so a single malformed entry cannot block scans.
- `SchemaManager::IsSchemasCompatibleForPKScan` iterates every schema id of  the table, skipping self-compare, and checks compatibility against the  current `TableSchema`. `schema_cache_` is now protected by a `mutex`  because parallel scan paths may query it concurrently.
- `NewDataTableScan` invokes the compatibility check once at scan-creation  time. If all other schema versions are PK-scan-compatible with the latest  schema the scan proceeds; otherwise it still returns `NotImplemented`  with the original message, preserving the existing failure behavior.
- `KeyValueFileStoreScan::FilterByValueFilter` no longer hard-fails when a  manifest entry's `schema_id` differs from the current table schema id.  It resolves the data file's own schema via `SchemaManager::ReadSchema`  and builds the `SimpleStatsEvolution` against it. Cross-schema  compatibility is already verified once at scan-creation time, so this  hot path stays cheap.

### Tests

New unit tests:

- `src/paimon/core/schema/table_schema_test.cpp`
  - `TableSchemaTest.TestPrimaryKeyScanSchemaCompatibility`: covers the    strict default, opt-in via regex, field-shape changes always rejected,    wildcard `.*` accepting read-side option changes, and invalid regex patterns being silently ignored without altering correctness.
- `src/paimon/core/schema/schema_manager_test.cpp`
  - `SchemaManagerTest.TestAllSchemasCompatibleForPrimaryKeyScan`: verifies multi-version iteration, self-compare skipping, opt-in compatibility for property-only changes, and field-shape changes still rejected even with `.*` configured.
- `src/paimon/core/table/source/table_scan_test.cpp`
  - `TableScanTest.TestPkPropertyOnlySchemaChange`: end-to-end check that without the option a PK property-only schema change is rejected, and with `scan.pk.schema-compatibility.ignored-options` set the scan plan is created successfully.
  - `TableScanTest.TestPkSchemaChangeChecksAllSchemas`: ensures every schema version is checked, so a field-level change anywhere in history still fails the scan.

### API and Format

- Adds a new public option key constant `Options::SCAN_PK_SCHEMA_COMPATIBILITY_IGNORED_OPTIONS`  (literal `scan.pk.schema-compatibility.ignored-options`) in `include/paimon/defs.h`.
- No on-disk storage format or wire-protocol changes.

### Documentation

No standalone documentation is required; this is a narrow relaxation of an existing restriction rather than a new feature surface. The new option's semantics are documented inline as a doxygen comment on `Options::SCAN_PK_SCHEMA_COMPATIBILITY_IGNORED_OPTIONS`.

### Generative AI tooling

Generated-by: OpenAI Codex

### DIFF
--- a/include/paimon/defs.h
+++ b/include/paimon/defs.h
@@ -508,6 +508,14 @@ struct PAIMON_EXPORT Options {
     /// "lookup.cache-max-disk-size" - Max disk size for lookup cache, you can use this option
     /// to limit the use of local disks. Default value is unlimited (INT64_MAX).
     static const char LOOKUP_CACHE_MAX_DISK_SIZE[];
+
+    /// "scan.pk.schema-compatibility.ignored-options" - A comma-separated list of regular
+    /// expressions describing option keys that are allowed to differ across schema versions when
+    /// performing the primary-key scan compatibility check. This is NOT general schema-evolution
+    /// support: it only permits property-only changes that do not affect the read semantics of
+    /// PK tables (e.g. write-side knobs). All other options must still be identical across schema
+    /// versions, otherwise the scan will fail. No default value (no extra options ignored).
+    static const char SCAN_PK_SCHEMA_COMPATIBILITY_IGNORED_OPTIONS[];
 };
 
 static constexpr int64_t BATCH_WRITE_COMMIT_IDENTIFIER = std::numeric_limits<int64_t>::max();
--- a/src/paimon/common/defs.cpp
+++ b/src/paimon/common/defs.cpp
@@ -141,4 +141,6 @@ const char Options::LOOKUP_CACHE_HIGH_PRIO_POOL_RATIO[] = "lookup.cache.high-pri
 const char Options::BUCKET_FUNCTION_TYPE[] = "bucket-function.type";
 const char Options::LOOKUP_CACHE_FILE_RETENTION[] = "lookup.cache-file-retention";
 const char Options::LOOKUP_CACHE_MAX_DISK_SIZE[] = "lookup.cache-max-disk-size";
+const char Options::SCAN_PK_SCHEMA_COMPATIBILITY_IGNORED_OPTIONS[] =
+    "scan.pk.schema-compatibility.ignored-options";
 }  // namespace paimon
--- a/src/paimon/core/core_options.cpp
+++ b/src/paimon/core/core_options.cpp
@@ -381,6 +381,7 @@ struct CoreOptions::Impl {
     std::vector<std::string> blob_descriptor_fields;
     std::vector<std::string> blob_view_fields;
     std::vector<std::string> blob_external_storage_fields;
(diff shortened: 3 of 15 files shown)


include/paimon/defs.h
```diff
@@ -508,6 +508,14 @@ struct PAIMON_EXPORT Options {
...
     /// to limit the use of local disks. Default value is unlimited (INT64_MAX).
     static const char LOOKUP_CACHE_MAX_DISK_SIZE[];
+
+    /// "scan.pk.schema-compatibility.ignored-options" - A comma-separated list of regular
+    /// expressions describing option keys that are allowed to differ across schema versions when
+    /// performing the primary-key scan compatibility check. This is NOT general schema-evolution
+    /// support: it only permits property-only changes that do not affect the read semantics of
+    /// PK tables (e.g. write-side knobs). All other options must still be identical across schema
+    /// versions, otherwise the scan will fail. No default value (no extra options ignored).
+    static const char SCAN_PK_SCHEMA_COMPATIBILITY_IGNORED_OPTIONS[];
```
lxy-9602, line 518: I completely understand your use case, and it sounds very reasonable. While I think introducing an intermediate option that is not supported by Java would be a bit redundant at this stage. I’d prefer going straight to proper schema evolution support for primary-key tables.

For this part, we can align with the Java implementation by following `FilterByStats` in `append_only_file_store_scan.cpp`. With support from `SimpleStatsEvolution::EvolutionStats` util, I’d expect the scope of changes to be roughly comparable to this PR.

gripleaf: I have submitted a new PR #368 following your suggestion.

PR #366, opened 2026-06-13
feat: add global system tables framework under `sys` database

suxiaogang223:
## Summary

Add global system table support under the `sys` database, enabling catalog-level metadata queries.

Closes part of #141.

### Infrastructure
- Add `GetOptions()` to `Catalog` interface for catalog-level config access
- `FileSystemCatalog` stores and exposes catalog options
- New `GlobalSystemTableLoader` with independent registry for `sys` tables
- Extend `SystemTablePath` with `is_global` flag and `TryParsePath` to detect `sys/` paths
- `FileSystemCatalog` handles `sys` database in `ListTables`/`DatabaseExists`/`TableExists`/`LoadTableSchema`

### Global System Tables

| Table | Status | Columns |
|-------|--------|---------|
| `sys.catalog_options` | ✅ Complete | `key`, `value` |
| `sys.all_table_options` | ✅ Complete | `database_name`, `table_name`, `key`, `value` |
| `sys.tables` | ✅ Complete | `database_name`, `table_name`, `table_type`, `partitioned`, `primary_key`, `record_count`, `file_size_in_bytes`, `file_count`, `last_file_creation_time` |
| `sys.partitions` | ✅ Complete | `database_name`, `table_name`, `partition_name`, `record_count`, `file_size_in_bytes`, `file_count`, `last_update_time` |

### Integration Tests (4 new tests in SystemTableReadInteTest)

- `TestReadGlobalCatalogOptions` — verifies schema + content
- `TestReadGlobalAllTableOptions` — verifies table options across databases
- `TestReadGlobalTables` — verifies schema, table metadata
- `TestReadGlobalPartitions` — verifies empty result for unpartitioned tables

## Files Changed

```
 include/paimon/catalog/catalog.h                   |   5 +
 src/paimon/CMakeLists.txt                          |   1 +
 src/paimon/core/catalog/catalog.cpp                |   2 +-
 src/paimon/core/catalog/file_system_catalog.cpp    |  39 +-
 src/paimon/core/catalog/file_system_catalog.h      |   6 +-
 .../core/table/system/global_system_tables.cpp     | 370 ++++++++++++++
 .../core/table/system/global_system_tables.h       | 117 +++++
 src/paimon/core/table/system/system_table.cpp      |  28 +-
 src/paimon/core/table/system/system_table.h        |   2 +
 .../core/catalog/file_system_catalog_test.cpp      |  14 +-
 test/inte/read_inte_test.cpp                       | 237 +++++++++
 11 files changed, 810 insertions(+), 10 deletions(-)
```

## Verification

Fedora x86_64, GCC 16:
- Build: ✅
- paimon-core-test: 36/40 pass (4 failures pre-existing)
- paimon-read-inte-test (global tables): 4/4 ✅
- paimon-read-inte-test (system tables): same as before


src/paimon/core/table/system/system_table.cpp
```diff
@@ -204,6 +215,21 @@ Result<std::shared_ptr<SystemTable>> SystemTableLoader::LoadFromPath(
...
     const auto& parsed = system_table_path.value();
+
+    // Handle global system tables (under sys/ directory)
+    if (parsed.is_global) {
+        GlobalSystemTableContext context;
+        context.fs = fs;
+        // The warehouse is the grandparent of the sys/<table> path
+        context.warehouse = PathUtil::GetParentDirPath(PathUtil::GetParentDirPath(path));
+        context.catalog_options = dynamic_options;
+        // Note: context.catalog is intentionally left as nullptr here.
```
zjw1111, line 236: I think this path can crash for the catalog-dependent tables. The normal engine read path (`TableScan::NewScan` / `TableRead::NewRead`) resolves system tables through `LoadFromPath`, which lands here and never sets `context.catalog`. Since `GlobalSystemTableContext::catalog` has no in-class initializer, `context.catalog` is actually indeterminate here rather than `nullptr` as the comment says. Later, `BuildRows()` for `sys.tables` / `sys.partitions` / `sys.all_table_options` calls `context_.catalog->ListDatabases()`, dereferencing an uninitialized pointer -> UB/segfault. Only `sys.catalog_options` works on this path.

The integration tests don't catch this because `ReadGlobalSystemTable` calls `GlobalSystemTableLoader::Load` directly with a real catalog, bypassing `LoadFromPath` entirely.

Could you guard this explicitly, e.g. return a `Status::NotImplemented(...)` for tables that require catalog enumeration when `context.catalog == nullptr`, and add a test that exercises the path-based (no-catalog) read? That would turn the crash into a controlled error.

src/paimon/core/table/system/global_system_tables.h
```diff
@@ -0,0 +1,117 @@
...
+#include "paimon/core/table/system/in_memory_system_table.h"
+
+namespace paimon {
+class Catalog;
+class FileSystem;
+
+/// Context passed to global system table constructors, providing catalog-level
+/// access for enumerating databases, tables, and reading metadata.
+struct GlobalSystemTableContext {
+    Catalog* catalog;  // non-owning pointer
```
zjw1111: Would it be possible to add a default initializer here, i.e. `Catalog* catalog = nullptr;`? Right now `GlobalSystemTableContext context;` in `LoadFromPath` leaves this member indeterminate rather than null, which makes the null-check suggested in `system_table.cpp` reliable and matches the intent of the comment there.

src/paimon/core/table/system/system_table.cpp
```diff
@@ -179,6 +180,17 @@ Result<std::shared_ptr<SystemTable>> SystemTableLoader::Load(
 
 Result<std::optional<SystemTablePath>> SystemTableLoader::TryParsePath(const std::string& path) {
     std::string table_name = PathUtil::GetName(path);
+    std::string parent = PathUtil::GetParentDirPath(path);
+    std::string parent_name = PathUtil::GetName(parent);
+
+    // Detect global system table paths: <warehouse>/sys/<table_name>
+    if (parent_name == "sys") {
```
zjw1111: Minor: this hardcodes the literal `"sys"`, but the repo already exposes `Catalog::SYSTEM_DATABASE_NAME` (used by `FileSystemCatalog::IsSystemDatabase`). Could you reuse that constant here to keep the system-database name in one place?

zjw1111 (commented):
Thanks a lot for building out the global `sys` system tables — the loader/registry split and reuse of `InMemorySystemTable` make the design clean and easy to follow.

I left a few inline comments. The most important one is a potential crash on the normal engine read path for `sys.tables` / `sys.partitions` / `sys.all_table_options` (see the comment in `system_table.cpp`). A couple of smaller notes are below:

- `global_system_tables.cpp` (`CatalogOptionsSystemTable::BuildRows`, around line 264): this table stores fields via a non-owning `std::string_view` into `context_.catalog_options`, while the other tables copy through `BinaryString` (`StringValue`). It's safe today because the map is a stable member, but for consistency and to avoid a future dangling-view footgun, would it be possible to use `StringValue(...)` here too?
- `file_system_catalog.cpp` (line 282): the `const_cast<FileSystemCatalog*>(this)` looks unnecessary since only const `Catalog` methods are used later — declaring `GlobalSystemTableContext::catalog` as `const Catalog*` would let you drop it. Minor, feel free to skip.

Thanks again!

zjw1111 (commented):
One additional inline note on the `sys.tables` `primary_key` column — a schema/data-contract nit. Thanks!

src/paimon/core/table/system/global_system_tables.cpp
```diff
@@ -0,0 +1,510 @@
...
+    return kName;
+}
+
+Result<std::shared_ptr<arrow::Schema>> TablesSystemTable::ArrowSchema() const {
+    return arrow::schema({
+        arrow::field("database_name", arrow::utf8(), /*nullable=*/false),
+        arrow::field("table_name", arrow::utf8(), /*nullable=*/false),
+        arrow::field("table_type", arrow::utf8(), /*nullable=*/false),
+        arrow::field("partitioned", arrow::boolean(), /*nullable=*/false),
+        arrow::field("primary_key", arrow::utf8(), /*nullable=*/false),
```
zjw1111: One more small consistency note, thanks for bearing with me.

This field is declared non-nullable, but `BuildRows` writes `NullType()` for tables without a primary key:

```cpp
row.SetField(4, primary_keys_str.empty() ? VariantType(NullType())
                                         : VariantType(StringValue(primary_keys_str)));
```

So a no-PK table produces a null in a column the schema promises is non-null — a schema/data-contract mismatch. `TestReadGlobalTables` only covers a PK table, so this path isn't exercised. Would it be possible to either mark this field `nullable=true` (closer to how Java Paimon expresses it), or emit an empty string for no-PK tables? A no-PK test case would help lock this down too.

src/paimon/core/catalog/file_system_catalog.h
```diff
@@ -38,7 +38,8 @@ class Logger;
 
 class FileSystemCatalog : public Catalog {
  public:
-    FileSystemCatalog(const std::shared_ptr<FileSystem>& fs, const std::string& warehouse);
+    FileSystemCatalog(const std::shared_ptr<FileSystem>& fs, const std::string& warehouse,
+                      const std::map<std::string, std::string>& catalog_options = {});
```
lxy-9602, line 43: Could you please avoid default parameters in production code?

src/paimon/core/catalog/file_system_catalog_test.cpp
```diff
@@ -606,8 +608,15 @@ TEST(FileSystemCatalogTest, TestInvalidList) {
...
-                        "do not support listing tables for system database.");
+    ASSERT_OK_AND_ASSIGN(auto sys_tables, catalog.ListTables("sys"));
+    ASSERT_FALSE(sys_tables.empty());
+    // Verify expected global system table names are present
+    ASSERT_TRUE(std::find(sys_tables.begin(), sys_tables.end(), "catalog_options") !=
+                sys_tables.end());
+    ASSERT_TRUE(std::find(sys_tables.begin(), sys_tables.end(), "all_table_options") !=
+                sys_tables.end());
+    ASSERT_TRUE(std::find(sys_tables.begin(), sys_tables.end(), "tables") != sys_tables.end());
+    ASSERT_TRUE(std::find(sys_tables.begin(), sys_tables.end(), "partitions") != sys_tables.end());
```
lxy-9602, line 620: Please rename the test from `TestInvalidList` to something like `TestSystemList`.

src/paimon/core/catalog/file_system_catalog_test.cpp
```diff
@@ -606,8 +608,15 @@ TEST(FileSystemCatalogTest, TestInvalidList) {
     auto dir = UniqueTestDirectory::Create();
     ASSERT_TRUE(dir);
     FileSystemCatalog catalog(core_options.GetFileSystem(), dir->Str());
-    ASSERT_NOK_WITH_MSG(catalog.ListTables("sys"),
-                        "do not support listing tables for system database.");
+    ASSERT_OK_AND_ASSIGN(auto sys_tables, catalog.ListTables("sys"));
+    ASSERT_FALSE(sys_tables.empty());
+    // Verify expected global system table names are present
+    ASSERT_TRUE(std::find(sys_tables.begin(), sys_tables.end(), "catalog_options") !=
+                sys_tables.end());
```
lxy-9602, line 616: Java gates `sys.catalog_options` behind `catalog-options-table.enabled` and keeps it disabled by default, since it may expose catalog-level configuration. Should we mirror that behavior here instead of always listing/loading `catalog_options` under `sys`?

src/paimon/core/table/system/global_system_tables.cpp
```diff
@@ -0,0 +1,510 @@
...
+        arrow::field("record_count", arrow::int64(), /*nullable=*/true),
+        arrow::field("file_size_in_bytes", arrow::int64(), /*nullable=*/true),
+        arrow::field("file_count", arrow::int64(), /*nullable=*/true),
+        arrow::field("last_file_creation_time", arrow::timestamp(arrow::TimeUnit::MILLI),
+                     /*nullable=*/true),
+    });
```
lxy-9602, line 413: I noticed that the current C++ `sys.tables` schema and semantics differ from Java Paimon's `sys.tables` contract.

In Java, `AllTablesTable` exposes 14 columns:

database_name, table_name, table_type, partitioned, primary_key,
owner, created_at, created_by, updated_at, updated_by,
record_count, file_size_in_bytes, file_count, last_file_creation_time

In this PR, the C++ implementation exposes only 9 columns:

database_name, table_name, table_type, partitioned, primary_key,
record_count, file_size_in_bytes, file_count, last_file_creation_time

There are also a few semantic/type differences:

- `primary_key`: Java uses a boolean to indicate whether the table has primary keys, while C++ currently returns a string containing the primary key column names, or null when there is no primary key.
- `table_type`: Java reads this from the table option `CoreOptions.TYPE` with its default value, while C++ derives `MANAGED` / `EXTERNAL` from whether `data-file.external-paths` is set.
- `last_file_creation_time`: Java exposes this as a BIGINT epoch millis value, while C++ exposes it as an Arrow timestamp(ms).
- Java includes audit fields (`owner`, `created_at`, `created_by`, `updated_at`, `updated_by`), but C++ currently omits them.

Could we align the C++ schema and behavior with Java here? Otherwise clients that rely on the Java `sys.tables` contract may see different columns, different types, and different values when using the C++ reader.

src/paimon/core/table/system/global_system_tables.cpp
```diff
@@ -0,0 +1,510 @@
...
+    return arrow::schema({
+        arrow::field("database_name", arrow::utf8(), /*nullable=*/false),
+        arrow::field("table_name", arrow::utf8(), /*nullable=*/false),
+        arrow::field("partition_name", arrow::utf8(), /*nullable=*/true),
+        arrow::field("record_count", arrow::int64(), /*nullable=*/true),
+        arrow::field("file_size_in_bytes", arrow::int64(), /*nullable=*/true),
+        arrow::field("file_count", arrow::int64(), /*nullable=*/true),
+        arrow::field("last_update_time", arrow::timestamp(arrow::TimeUnit::MILLI),
+                     /*nullable=*/true),
+    });
```
lxy-9602, line 479: Could we align `sys.partitions` with Java's `AllPartitionsTable` contract?

Java exposes:
`database_name`, `table_name`, `partition_name`, `record_count`, `file_size_in_bytes`, `file_count`, `last_file_creation_time`, `done`

This PR exposes:
`database_name`, `table_name`, `partition_name`, `record_count`, `file_size_in_bytes`, `file_count`, `last_update_time`

So C++ is missing `done`, uses a different time column name/type, and formats partition names differently. Java uses `PartitionUtils.buildPartitionName(spec)`, e.g. `dt=2024-07-22/region=cn`, while C++ currently returns values like `{dt=2024-07-22,region=cn}`.


Could we align this table with Java's schema and semantics?

src/paimon/core/table/system/global_system_tables.cpp
```diff
@@ -0,0 +1,555 @@
...
+            if (!table_path_result.ok()) {
+                continue;
+            }
+            std::string table_path = table_path_result.value();
+
+            auto file_stats_result =
+                AggregateFileStats(context_.fs, table_path, data_schema->Options());
+            if (!file_stats_result.ok()) {
+                continue;
+            }
```
lxy-9602, line 512: I’m a bit concerned that the current validation may be too permissive. As it stands, corrupted schema or manifest files could lead to successful queries with incomplete results, where errors are silently converted into missing rows or nulls. In the Java side, only explicit `DatabaseNotExist`, `TableNotExist`, and some `NotImplemented` cases are ignored. At minimum, I think we should add a TODO and a function comment to explicitly note that this case is currently inconsistent with the Java behavior.

lxy-9602 (dismissed):
+1

zjw1111 (commented):
Thanks again for the updates. Two small consistency notes on the new `sys.tables` implementation versus Java Paimon.

src/paimon/core/table/system/global_system_tables.cpp
```diff
@@ -0,0 +1,566 @@
...
+
+    PAIMON_ASSIGN_OR_RAISE(std::vector<CatalogTableInfo> tables,
+                           LoadAllDataTables(*context_.catalog));
+    for (const CatalogTableInfo& table : tables) {
+        const std::shared_ptr<DataSchema>& data_schema = table.schema;
+        Identifier id(table.database_name, table.table_name);
+
+        const auto& opts = data_schema->Options();
+        auto table_type = opts.find("type");
+        const std::string table_type_str = table_type == opts.end() ? "TABLE" : table_type->second;
```
zjw1111: Thanks for aligning the `sys.tables` schema with Java — the 14-column layout looks great.

One small consistency issue on the default `table_type`: here it falls back to `"TABLE"` (uppercase), but Java's `AllTablesTable` uses `options.getOrDefault(TYPE.key(), TYPE.defaultValue().toString())`, and `TableType.TABLE.toString()` returns the lowercase `"table"` (see `TableType`: `TABLE("table", ...)`). So for a normal table without an explicit `type` option, the C++ reader emits `"TABLE"` while Java emits `"table"`, which clients relying on the Java contract would see as a mismatch.

Would it be possible to default to the lowercase `"table"`? Note that `TestReadGlobalTables` currently asserts `"TABLE"`, so that assertion would need updating as well.

src/paimon/core/table/system/global_system_tables.cpp
```diff
@@ -0,0 +1,566 @@
...
+            for (const auto& [key, value] : partition_values) {
+                if (!partition_key.empty()) {
+                    partition_key += "/";
+                }
+                partition_key += key + "=" + value;
+            }
+        }
+
+        auto& stats = result.by_partition[partition_key];
+        stats.record_count += file->row_count;
```
zjw1111, line 255: Thanks for populating the file statistics here.

I wonder if this diverges from Java's contract for a filesystem catalog. In Java, `CatalogUtils.toTableAndSnapshots` only loads a `TableSnapshot` when `catalog.supportsVersionManagement()` is true; `FileSystemCatalog` (via `AbstractCatalog`) returns false, so `snapshot` stays null and `AllTablesTable` emits null for `record_count` / `file_size_in_bytes` / `file_count` / `last_file_creation_time`. This PR instead reads the latest snapshot and manifests to compute them, so for a table with committed data the C++ reader returns non-null values where Java returns null.

In addition, the semantics of `record_count` itself differ: here it is the sum of `row_count` over merged ADD entries (i.e. live rows), whereas Java's `snapshot.recordCount()` is the accumulated `totalRecordCount`, so even when populated the two are not equivalent.

`TestReadGlobalTables` doesn't surface this because the tables have no committed data (empty snapshot → all-null branch). Could you clarify whether this is an intentional enhancement? If so, it would help to document the difference from the Java FS-catalog contract in a comment and add a "table with data" case to lock in the expected values.

lxy-9602 (approved):
+1

PR #368, opened 2026-06-15
feat(schema): support pk table schema evolution

gripleaf:
### Purpose

  Support schema evolution scan for primary-key tables.

  This change:
  - removes the scan-time rejection for evolved primary-key table schemas
  - aligns primary-key value stats filtering with append-only scan behavior
  - uses `SimpleStatsEvolution::EvolutionStats` to evolve historical data-file stats before predicate evaluation
  - reconstructs predicates against the data-file schema and conservatively keeps files when predicate fields are missing or require casts
  - adds coverage for renamed fields, newly added fields, and PK table scan planning with evolved schemas

### Tests

  Added and verified focused unit tests for:

  - Primary-key table scan planning with evolved schemas, covering that PK tables with schema changes can create scan plans instead of being rejected.
  - Primary-key value stats filtering across schema evolution, covering renamed-field stats alignment and conservative retention for old files missing newly added fields.
  - Predicate reconstruction behavior for evolved stats filtering, covering that fields missing from historical data schemas are excluded to avoid false-negative file pruning.

  ### API and Format

  No public API, storage format, or protocol change.

  ### Documentation

  No documentation change.

  ### Generative AI tooling


src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -70,11 +70,13 @@ Result<std::shared_ptr<Predicate>> FileStoreScan::ReconstructPredicateWithNonCas
...
-                excluded_field_names.insert(field_name);
-            }
+        if (data_iter == id_to_data_fields.end()) {
+            excluded_field_names.insert(field_name);
+            continue;
+        }
+        // Exclude fields requiring casting to avoid false negatives in stats filtering.
+        if (!data_iter->second.second.Type()->Equals(table_iter->second.Type())) {
+            excluded_field_names.insert(field_name);
         }
```
lxy-9602, line 79: One concern about excluding predicate fields that are missing from the historical data schema: this differs from Java's scan semantics for both append and primary-key tables.

In Java, `AppendOnlyFileStoreScan` and `KeyValueFileStoreScan` both call `filterUnsafeFilter(..., keepNewFieldFilter = true)`. With that flag, if an old data schema does not contain a newly added field, `devolveFilters` keeps the predicate instead of dropping it.

This works together with stats evolution: old files can represent the newly added field as all NULL (`nullCount = rowCount`). With that behavior:

- `new_field IS NULL` can match old files.
- `new_field > 30` / `new_field = 3` can prune old files.
- This is correct because rows written before the column was added logically read the new column as NULL.

The current change drops such predicates in `ReconstructPredicateWithNonCastedFields`, which makes C++ more conservative and weakens stats pruning compared with Java, and it also changes the existing append-table behavior. Is this divergence intentional for C++/PK scans, or should missing-field predicates be kept like Java while still excluding cast-required predicates?

src/paimon/core/table/source/table_scan_test.cpp
```diff
@@ -47,14 +47,16 @@ TEST(TableScanTest, TestNonExistTable) {
...
     std::string path =
         paimon::test::GetDataDir() + "/orc/pk_table_with_alter_table.db/pk_table_with_alter_table/";
     ScanContextBuilder builder(path);
     builder.AddOption(Options::FILE_FORMAT, "orc");
     ASSERT_OK_AND_ASSIGN(auto context, builder.Finish());
-    ASSERT_NOK_WITH_MSG(TableScan::Create(std::move(context)), "do not support schema evolution");
+    ASSERT_OK_AND_ASSIGN(auto table_scan, TableScan::Create(std::move(context)));
+    ASSERT_OK_AND_ASSIGN(auto plan, table_scan->CreatePlan());
+    ASSERT_TRUE(plan->SnapshotId());
+    ASSERT_FALSE(plan->Splits().empty());
```
lxy-9602, line 59: Could we add a inte test using `pk_table_with_alter_table` to verify that primary-key stats filtering still works correctly after a PK field rename (maybe in `scan_and_read_inte_test.cpp`)?

In this table, schema-0 has primary key `key2`, while schema-1 renames it to `key_2`:

- schema-0: `primaryKeys = [key0, key1, key2]`
- schema-1: `primaryKeys = [key0, key1, key_2]`
- `key0` and `key1` are partition keys, so the trimmed PK used by `key_stats` is effectively `key2` / `key_2`.

It would be useful to add a scan predicate on the renamed PK field, for example `key_2 > 500 ...` or `key_2 < 32`.

lxy-9602 (approved):
+1

PR #371, opened 2026-06-17
fix(parquet): fallback nested field reading to RowGroup reading (walkaround)

zhf999:
### Purpose

- Fix a correctness issue when predicate pushdown enables page-level filtering on Parquet files containing nested columns.
- When read schema contains nested fields (struct/list/map), page-level predicate filtering currently uses `BuildPageFilteredSchema` mapping that does not correctly align nested Parquet leaf columns, causing nested column reads to fail.
- Add a fallback path: when nested columns exist in read schema, disable page-level index filtering and fall back to row-group level reading, which preserves correctness.

### Changes

- In `src/paimon/format/parquet/parquet_file_batch_reader.cpp`:
  - Detect nested columns in file schema via `ArrowSchemaValidator::IsNestedType`.
  - Skip page-level index filter initialization/build when nested field exists.
  - This makes reader use row-group filtered flow for nested schemas, avoiding the failing nested leaf-index mapping path.

### Tests

- Added/updated unit tests in `src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp` (no public API test harness changes).
- Validated expected behavior on:
  - nested struct full read with predicate
  - nested struct only-read with predicate on non-selected top-level field
  - nested list read with predicate
  - nested map read with predicate
  - adjacent nested columns coexistence

### API and Format

- No API changes.
- No storage format/protocol changes.

### Documentation

- No user-facing feature/documentation updates required.

### Generative AI tooling

- No


src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -129,6 +130,11 @@ Status ParquetFileBatchReader::SetReadSchema(
 
         PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Schema> file_schema, reader_->GetSchema());
         std::unordered_map<std::string, std::vector<int32_t>> field_index_map;
+        bool has_nested_field = false;
+        for (const auto& field : read_schema->fields()) {
+            has_nested_field =
+                has_nested_field || ArrowSchemaValidator::IsNestedType(field->type());
```
lxy-9602: It looks like has_nested_field could break early here.

src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp
```diff
@@ -873,5 +873,279 @@ TEST_F(PageFilteredRowGroupReaderTest, ComputePageRangesWithDictionaryEncoding)
...
+/// Predicate on "id": id >= 50.
+TEST_F(PageFilteredRowGroupReaderTest, NestedStructColumnOnlyReadNestedField) {
+    std::string file_name = dir_->Str() + "/nested_struct_only_nested.parquet";
+    auto data = MakeNestedStructData(100);
+    WriteTestFile(file_name, data, /*write_batch_size=*/10, /*max_row_group_length=*/50);
+
+    auto field_x = arrow::field("x", arrow::int32());
+    auto field_y = arrow::field("y", arrow::int32());
+    // Read only the nested "info" column
+    auto read_schema = arrow::schema({arrow::field("info", arrow::struct_({field_x, field_y}))});
```
lxy-9602: Could `arrow::field("info", arrow::struct_({field_x, field_y}))` be reused as field_info here? Also, it seems we cannot apply a predicate on id without also projecting id.

src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp
```diff
@@ -873,5 +873,279 @@ TEST_F(PageFilteredRowGroupReaderTest, ComputePageRangesWithDictionaryEncoding)
...
+    // Should get rows 50-99 = 50 rows
+    ASSERT_TRUE(result);
+    ASSERT_EQ(50, result->length());
+
+    // Build expected: only the "info" field from rows 50-99
+    auto sliced = std::dynamic_pointer_cast<arrow::StructArray>(data->Slice(50, 50));
+    ASSERT_TRUE(sliced);
+    // Extract only the "info" column (field index 1) and wrap as struct with single field
+    auto expected =
+        arrow::StructArray::Make({sliced->field(1)},
```
lxy-9602: I’m not sure I fully understand this test. Is the intention to show that filtering on nested fields can only be pushed down to the row-group level? If so, with the current predicate choice, it seems the recall would be the same whether filtering happens at the page level or only at the row-group level.

src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp
```diff
@@ -873,5 +873,279 @@ TEST_F(PageFilteredRowGroupReaderTest, ComputePageRangesWithDictionaryEncoding)
...
+
+/// Test: page-level filtering on a file with a list column.
+///
+/// Schema: { id: int32, tags: list<item: int32> }
+/// 100 rows, 10 per page, 1 row group.
+/// Predicate: id >= 50 → pages 0-4 skipped, pages 5-9 read → 50 rows expected.
+TEST_F(PageFilteredRowGroupReaderTest, NestedListColumnPageFilter) {
+    std::string file_name = dir_->Str() + "/nested_list_filter.parquet";
+    auto data = MakeListColumnData(100);
+    WriteTestFile(file_name, data, /*write_batch_size=*/10, /*max_row_group_length=*/50);
```
lxy-9602, line 1042: Why nested can use page?

src/paimon/format/parquet/parquet_file_batch_reader.cpp
zhf999: Done, the loop breaks early now.

src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp
zhf999: Done, the field `info` variable has been reused and `id` has been added to projection field.

src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp
zhf999: Done, the predicate has been changed to `id>=70` so that the query result differs on rowgroup-level filtering and page-level filtering.

src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp
zhf999: Done, the comments have been changed.

lxy-9602 (approved):
+1

PR #373, opened 2026-06-17
feat(parquet): support parquet metadata cache

gripleaf:
### Purpose

  Support caching raw Parquet metadata footer bytes to reduce repeated footer reads and metadata deserialization when opening the same Parquet files in one process.

  This change:
  - Adds `CacheKind::PARQUET_METADATA`.
  - Adds `CacheKey::ForParquetMeta(file_uri)` for footer cache keys.
  - Injects cache into reader builders through `WithCache`.
  - Caches raw Parquet footer bytes in `ParquetReaderBuilder`.
  - Passes cached `parquet::FileMetaData` into `ParquetFileBatchReader::Create`.
  - Reuses a shared counting cache test utility for manifest and Parquet metadata cache tests.

  ### Tests

  - ParquetFileBatchReaderTest.TestParquetMetadataRawBytesCacheReusesFooter
  - ParquetFileBatchReaderTest.TestParquetMetadataCacheBypassesWhenGetUriFails
  - WriteAndReadInteTest.TestAppendWithParquetMetadataCache

  ### API and Format

  This change affects public API in `include`:
  - Adds `CacheKind::PARQUET_METADATA`.
  - Adds `CacheKey::ForParquetMeta`.
  - Adds `ReaderBuilder::WithCache`.

  This change also updates the internal Parquet reader creation path to accept an optional `parquet::FileMetaData`.

  No storage format or protocol change.

  ### Documentation

  Yes. Added `docs/source/user_guide/parquet_metadata_cache.rst` and linked it from the user guide.

  ### Generative AI tooling


include/paimon/cache/cache.h
```diff
@@ -32,6 +32,7 @@ class CacheValue;
 enum class CacheKind {
     DEFAULT,
     MANIFEST,
+    PARQUET_METADATA,
 };
```
lxy-9602, line 36: `Parquet` is just one format plugin among others such as `ORC, Avro, Lance, and Blob`, so it probably should not be named explicitly in a header like CacheKind. Something more generic like `DATA_FILE_FOOTER` would be preferable. Also, I would avoid introducing an extra helper like `ForParquetMeta` in the header below; using `ForKind` directly should be sufficient.

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -71,7 +71,8 @@ ParquetFileBatchReader::ParquetFileBatchReader(
 Result<std::unique_ptr<ParquetFileBatchReader>> ParquetFileBatchReader::Create(
     std::shared_ptr<arrow::io::RandomAccessFile>&& input_stream,
     const std::shared_ptr<arrow::MemoryPool>& pool,
-    const std::map<std::string, std::string>& options, int32_t batch_size) {
+    const std::map<std::string, std::string>& options, int32_t batch_size,
+    std::shared_ptr<::parquet::FileMetaData> file_metadata) {
     try {
```
lxy-9602, line 77: Could you please move pool to be the last parameter? Previously, options and batch_size were also not placed very well :(

src/paimon/testing/utils/manifest_cache_test_utils.h
```diff
@@ -16,82 +16,18 @@
...
-#include "paimon/cache/cache.h"
-#include "paimon/common/io/cache/lru_cache.h"
-#include "paimon/result.h"
+#include "paimon/testing/utils/counting_cache_test_utils.h"
 
 namespace paimon::test {
 
-class CountingManifestRoutingCache : public Cache {
+class CountingManifestRoutingCache : public CountingRoutingCache {
  public:
```
lxy-9602: If `CountingManifestRoutingCache` is only a few lines, could we consider merging it into `CountingRoutingCache`?

src/paimon/format/parquet/parquet_reader_builder.h
```diff
@@ -41,24 +51,111 @@ class ParquetReaderBuilder : public ReaderBuilder {
...
 
  private:
+    Result<MemorySegment> SerializeParquetMetadataFooter(
+        const std::shared_ptr<arrow::io::RandomAccessFile>& input_stream,
+        const ::parquet::ReaderProperties& reader_properties,
+        const std::shared_ptr<arrow::MemoryPool>& arrow_pool) const {
+        constexpr int64_t kParquetFooterSize = 8;
+
+        std::shared_ptr<::parquet::FileMetaData> metadata =
+            ::parquet::ParquetFileReader::Open(input_stream, reader_properties)->metadata();
```
zjw1111, line 96: `ParquetFileReader::Open()` (here and at line 120) and `WriteFileMetaData()` (line 100) can throw `parquet::ParquetException`. These calls sit inside `Build()`, which has no `try-catch`.

In the existing codebase every public method that touches the Parquet C++ API wraps its body with `try { … } PAIMON_PARQUET_CATCH_AND_RETURN_STATUS(…)` (e.g. `ParquetFileBatchReader::Create`, `FileReaderWrapper` methods). Because the new caching helpers are invoked *before* `ParquetFileBatchReader::Create`, an exception here escapes the established catch boundary and propagates uncaught up through `AbstractSplitRead` (also no `try-catch`), which could terminate the process on a corrupted or unexpected Parquet file.

Would it be possible to wrap `Build()` in the same `try / PAIMON_PARQUET_CATCH_AND_RETURN_STATUS("ParquetReaderBuilder::Build")` pattern, or add individual `try-catch` blocks inside `SerializeParquetMetadataFooter` and `ParseParquetMetadataFooter`?

lxy-9602 (approved):
+1

lucasfang (approved):
+1

PR #376, opened 2026-06-17
feat(shredding): support shared shredding file reader

lszskye:
### Purpose
Linked issue: #342 
This PR adds the read-side `SharedShreddingFileReader` for Shared Shredding MAP columns, and extends the related utilities and tests.

Main changes:

- Add `SharedShreddingFileReader`
  - Converts the physical schema
    `STRUCT<__field_mapping, __col_i..., __overflow>`
    back to the logical `MAP<STRING, T>` schema.
  - Translates logical MAP read requests into physical struct child-column reads in `SetReadSchema`.
  - Rebuilds logical MAP arrays from physical shared-shredding struct arrays in `NextBatchWithBitmap`.
  - Supports:
    - reading the full MAP
    - reading a selected-key sub-MAP
    - falling back to `__overflow` when needed
  - Validates invalid physical data:
    - `__field_mapping` must not be NULL for non-NULL physical struct rows
    - `__field_mapping` elements must not be NULL

- Extend `MapSharedShreddingUtils`
  - Add `BuildSpecificPhysicalStructType`, used to build a physical struct schema containing only selected physical columns.
  - Add `IsOverflowField`, used to determine whether a logical key may appear in `__overflow`.
  - Keep the existing `KeyValueMetadata` utility signatures and adapt call sites instead.

### Tests
  - `SharedShreddingFileReaderTest`
  - `MapSharedShreddingUtilsTest`


src/paimon/common/data/shredding/map_shared_shredding_utils.h
```diff
@@ -80,6 +88,13 @@ class MapSharedShreddingUtils {
         const std::shared_ptr<arrow::Schema>& logical_schema,
         const std::map<std::string, int32_t>& field_to_num_columns);
 
+    /// Builds a projected physical Struct type for one shredding MAP column.
+    /// The struct always contains __field_mapping and only the requested physical columns.
+    static std::shared_ptr<arrow::DataType> BuildSpecificPhysicalStructType(
+        const std::shared_ptr<arrow::DataType>& value_type,
+        const std::set<int32_t>& physical_col_ids, bool value_nullable,
+        bool include_overflow = false);
```
lxy-9602: Please rm default param for `include_overflow`.

lxy-9602: Also add comments for input and output param like `value_type`.

src/paimon/common/data/shredding/map_shared_shredding_utils.cpp
```diff
@@ -75,6 +90,25 @@ Result<std::shared_ptr<MapSharedShreddingContext>> MapSharedShreddingUtils::Crea
...
 
 // ---- Schema conversion ----
+std::shared_ptr<arrow::DataType> MapSharedShreddingUtils::BuildSpecificPhysicalStructType(
+    const std::shared_ptr<arrow::DataType>& value_type, const std::set<int32_t>& physical_col_ids,
+    bool value_nullable, bool include_overflow) {
+    std::vector<int32_t> sorted_cols(physical_col_ids.begin(), physical_col_ids.end());
+    arrow::FieldVector struct_fields;
+    struct_fields.reserve(sorted_cols.size() + 2);
+    struct_fields.push_back(
+        arrow::field(MapSharedShreddingDefine::kFieldMapping, arrow::list(arrow::int32()), false));
```
lxy-9602: nullable `false`?

src/paimon/common/data/shredding/map_shared_shredding_utils.cpp
```diff
@@ -75,6 +90,25 @@ Result<std::shared_ptr<MapSharedShreddingContext>> MapSharedShreddingUtils::Crea
...
+    for (const auto& col : sorted_cols) {
+        struct_fields.push_back(arrow::field(MapSharedShreddingDefine::PhysicalColumnName(col),
+                                             value_type, value_nullable));
+    }
+    if (include_overflow) {
+        struct_fields.push_back(arrow::field(
+            MapSharedShreddingDefine::kOverflow,
+            arrow::map(arrow::int32(), arrow::field("value", value_type, value_nullable)), true));
+    }
```
lxy-9602, line 124: Please refactor and reuse `BuildPhysicalStructType` as inner helper.

src/paimon/common/data/shredding/map_shared_shredding_utils.cpp
```diff
@@ -75,6 +90,25 @@ Result<std::shared_ptr<MapSharedShreddingContext>> MapSharedShreddingUtils::Crea
 }
 
 // ---- Schema conversion ----
+std::shared_ptr<arrow::DataType> MapSharedShreddingUtils::BuildSpecificPhysicalStructType(
+    const std::shared_ptr<arrow::DataType>& value_type, const std::set<int32_t>& physical_col_ids,
+    bool value_nullable, bool include_overflow) {
+    std::vector<int32_t> sorted_cols(physical_col_ids.begin(), physical_col_ids.end());
+    arrow::FieldVector struct_fields;
```
lxy-9602, line 112: why copy `physical_col_ids` to `sorted_cols`? Seems unnecessary?

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+                                      arrow::Concatenate(item_slices, pool));
+    return items;
+}
+
+}  // namespace
+
+Result<std::unique_ptr<SharedShreddingFileReader>> SharedShreddingFileReader::Create(
+    std::unique_ptr<FileBatchReader>&& reader, const CoreOptions& options, const TableSchema&,
+    const std::shared_ptr<MemoryPool>& pool) {
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<::ArrowSchema> file_schema, reader->GetFileSchema());
```
lxy-9602, line 57: `TableSchema` is empty? Consider rm this param?

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+        if (MapSharedShreddingUtils::HasShreddingMetadata(metadata)) {
+            PAIMON_ASSIGN_OR_RAISE(
+                MapSharedShreddingFieldMeta meta,
+                MapSharedShreddingUtils::DeserializeMetadata(
+                    metadata, MapSharedShreddingDefine::kDefaultDictCompression));
+            shared_shredding_name_to_meta[field->name()] = std::move(meta);
+        }
+    }
+    return std::unique_ptr<SharedShreddingFileReader>(new SharedShreddingFileReader(
+        std::move(reader), options, shared_shredding_name_to_meta, pool));
```
lxy-9602: I’m wondering why we prepare `shared_shredding_name_to_meta` eagerly during `Create`, instead of building it from `read_schema` and only keeping the shared-shredding fields that are actually needed for the read. That way, the three maps could potentially be merged into a single map with an inner struct.

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+    }
+    if (!value_type) {
+        return Status::Invalid(fmt::format("cannot infer shared-shredding value type for field {}",
+                                           physical_field->name()));
+    }
+    return arrow::field(
+        physical_field->name(),
+        arrow::map(arrow::utf8(), arrow::field("value", value_type, value_nullable)),
+        physical_field->nullable());
+}
```
lxy-9602, line 135: Is the metadata on the map field intentionally ignored here?

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+                "cannot cast physical shredding field {} to StructArray", physical_field->name()));
+        }
+
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Array> logical_map_array,
+                               RebuildLogicalMapArray(physical_field, physical_struct_array));
+        resolved_arrays[field_idx] = logical_map_array;
+        resolved_fields[field_idx] =
+            arrow::field(physical_field->name(), logical_map_array->type(),
+                         physical_field->nullable(), physical_field->metadata());
+    }
```
lxy-9602, line 267: Why we have `metadata` here but ignore in above func?

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+    PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::StructArray> new_struct_array,
+                                      arrow::StructArray::Make(resolved_arrays, resolved_fields));
+    auto new_c_array = std::make_unique<ArrowArray>();
+    auto new_c_schema = std::make_unique<ArrowSchema>();
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(
+        arrow::ExportArray(*new_struct_array, new_c_array.get(), new_c_schema.get()));
+    batch = std::make_pair(std::move(new_c_array), std::move(new_c_schema));
+    return batch_with_bitmap;
```
lxy-9602, line 276: Is there any case for dict array in map & out of map?

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+}
+
+Result<std::shared_ptr<arrow::Array>> SharedShreddingFileReader::RebuildLogicalMapArray(
+    const std::shared_ptr<arrow::Field>& physical_field,
+    const std::shared_ptr<arrow::StructArray>& physical_struct_array) const {
+    std::string shredding_field_name = physical_field->name();
```
lxy-9602, line 282: split this func, too long.

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+    arrow::StringBuilder keys_builder(arrow_pool_.get());
+    arrow::ArrayVector item_slices;
+    int32_t item_count = 0;
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(offsets_builder.Append(0));
+    bool has_null = physical_struct_array->null_count() > 0;
+    std::vector<uint8_t> row_validities;
+    if (has_null) {
+        row_validities.reserve(physical_struct_array->length());
+    }
+
```
lxy-9602, line 327: Consider using `map_builder` here instead of preparing the `offsets` and `null bitmap manually`. Please refer to `map_shared_shredding_batch_converter.cpp` or `row_to_arrow_array_converter.h`.

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+    arrow::Int32Builder offsets_builder(arrow_pool_.get());
+    arrow::StringBuilder keys_builder(arrow_pool_.get());
+    arrow::ArrayVector item_slices;
+    int32_t item_count = 0;
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(offsets_builder.Append(0));
+    bool has_null = physical_struct_array->null_count() > 0;
+    std::vector<uint8_t> row_validities;
+    if (has_null) {
```
lxy-9602: Also please do reserve for map array.

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+            }
+
+            int32_t overflow_offset = overflow_array->value_offset(row);
+            int32_t overflow_length = overflow_array->value_length(row);
+            for (int32_t pos = 0; pos < overflow_length; ++pos) {
+                int32_t overflow_index = overflow_offset + pos;
+                if (!overflow_keys->IsNull(overflow_index) &&
+                    overflow_keys->Value(overflow_index) == selected_field_id) {
+                    PAIMON_RETURN_NOT_OK_FROM_ARROW(keys_builder.Append(selected_key));
+                    item_slices.push_back(overflow_items->Slice(overflow_index, 1));
```
lxy-9602: Try `AppendArraySlice` for builder rather than concact.

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+    const std::map<std::string, MapSharedShreddingFieldMeta>& shared_shredding_name_to_meta,
+    const std::shared_ptr<MemoryPool>& pool)
+    : arrow_pool_(GetArrowPool(pool)),
+      reader_(std::move(reader)),
+      options_(options),
+      shared_shredding_name_to_meta_(shared_shredding_name_to_meta) {}
+
+Result<std::unique_ptr<::ArrowSchema>> SharedShreddingFileReader::GetFileSchema() const {
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<::ArrowSchema> physical_schema,
+                           reader_->GetFileSchema());
```
lxy-9602, line 87: One thing we may need to be careful about at the framework level later: a user might first configure `tags = shared-shredding`, and then later change the options so that it becomes `tags = default`. If the `SharedShreddingFileReader` layer remains unchanged, we will need to handle reader routing very carefully in that case. So my understanding is that we should not determine which columns are shared-shredding based on `options`; instead, it should be determined from metadata.

src/paimon/common/data/shredding/shared_shredding_file_reader_test.cpp
```diff
@@ -0,0 +1,461 @@
...
+}
+
+TEST_F(SharedShreddingFileReaderTest, TestSpecialSelectedKeysWithDuplicatedEmptyKey) {
+    for (const std::string& selected_keys : {",", ",,"}) {
+        auto reader = CreateReader();
+        auto read_schema = ExportSchema(ReadSchema(selected_keys));
+        ASSERT_NOK_WITH_MSG(reader->SetReadSchema(read_schema.get(), /*predicate=*/nullptr,
+                                                  /*selection_bitmap=*/std::nullopt),
+                            "duplicate key  in paimon.map.selected-keys for field tags");
+    }
```
lxy-9602, line 346: Consider using `key []` here to emphasize the key.

src/paimon/common/data/shredding/shared_shredding_file_reader_test.cpp
```diff
@@ -0,0 +1,461 @@
...
+    auto reader = CreateReader(physical_array, physical_schema);
+    auto read_schema = ExportSchema(ReadSchema("b"));
+    ASSERT_OK(reader->SetReadSchema(read_schema.get(), /*predicate=*/nullptr,
+                                    /*selection_bitmap=*/std::nullopt));
+    ASSERT_NOK_WITH_MSG(ReadResultCollector::CollectResult(reader.get()),
+                        "__field_mapping element cannot be null");
+}
+
+TEST_F(SharedShreddingFileReaderTest, TestReadsRealFormatFile) {
+    // TODO(lisizhuo.lsz): support other format
```
lxy-9602, line 518: consider use `TETS_P`

lxy-9602: Also please add tests for nested map value.

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+    shared_shredding_name_to_selected_keys_.clear();
+    shared_shredding_name_to_map_type_.clear();
+    arrow::FieldVector resolved_fields = logical_read_schema->fields();
+    for (const auto& name : shared_shredding_names) {
+        const auto& field = logical_read_schema->GetFieldByName(name);
+        if (!field) {
+            return Status::Invalid(
+                fmt::format("cannot find shared-shredding field {} in read schema", field->name()));
+        }
```
zjw1111, line 164: Nit: when `field` is null, `field->name()` will dereference a null pointer and crash. Could you use `name` (the loop variable) here instead?

```suggestion
                fmt::format("cannot find shared-shredding field {} in read schema", name));
```

src/paimon/common/data/shredding/map_shared_shredding_utils.cpp
```diff
@@ -35,8 +35,23 @@
 #include "rapidjson/writer.h"
 
 namespace paimon {
-// ---- Column detection ----
+Result<std::vector<int32_t>> MapSharedShreddingUtils::GetPhysicalColumnIndices(
+    const MapSharedShreddingFieldMeta& meta, const std::string& name) {
+    auto name_iter = meta.name_to_id.find(name);
+    if (name_iter == meta.name_to_id.end()) {
+        return Status::Invalid(
+            fmt::format("cannot find field {} in map shared sharedding meta", name));
```
zjw1111: Minor typo: `sharedding` → `shredding`. The same typo also appears at lines 48–49 and in `IsOverflowField` (line 442).

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -0,0 +1,484 @@
...
+                fmt::format("cannot find shared-shredding metadata for field {}", field->name()));
+        }
+        PAIMON_ASSIGN_OR_RAISE(std::optional<std::vector<std::string>> selected_keys_opt,
+                               GetSelectedKeys(field));
+        std::vector<std::string> selected_keys;
+        if (!selected_keys_opt) {
+            // select all keys
+            selected_keys.reserve(meta_iter->second.name_to_id.size());
+            for (const auto& [name, _] : meta_iter->second.name_to_id) {
+                selected_keys.push_back(name);
```
zjw1111: This `name` (a MAP key like `"a"`, `"b"`) shadows the outer loop variable `name` (the shredding column name like `"tags"`). It works today because the code below uses `field->name()` instead of `name`, but it would be easy to introduce a bug here during future edits. Would it be possible to rename this to something like `key_name` or `map_key`?

src/paimon/common/data/shredding/map_shared_shredding_utils.cpp
```diff
@@ -400,6 +434,16 @@ bool MapSharedShreddingUtils::HasShreddingMetadata(
...
 }
 
+Result<bool> MapSharedShreddingUtils::IsOverflowField(const MapSharedShreddingFieldMeta& meta,
+                                                      const std::string& name) {
+    auto name_iter = meta.name_to_id.find(name);
+    if (name_iter == meta.name_to_id.end()) {
+        return Status::Invalid(
+            fmt::format("cannot find field {} in map shared sharedding meta", name));
+    }
+    return meta.overflow_field_set.count(name_iter->second) > 0 ? true : false;
```
zjw1111: Ultra-minor: the ternary is redundant since `count() > 0` already yields a `bool`.

```suggestion
    return meta.overflow_field_set.count(name_iter->second) > 0;
```

src/paimon/common/data/shredding/shared_shredding_file_reader.h
```diff
@@ -0,0 +1,80 @@
...
+
+    Result<std::shared_ptr<arrow::Array>> RebuildLogicalMapArray(
+        const std::shared_ptr<arrow::Field>& physical_field,
+        const std::shared_ptr<arrow::StructArray>& physical_struct_array) const;
+
+    static Result<std::shared_ptr<arrow::Field>> ToLogicalMapField(
+        const std::shared_ptr<arrow::Field>& physical_field);
+
+ private:
+    std::shared_ptr<arrow::MemoryPool> arrow_pool_;
```
zjw1111, line 77: Nit: there's already a `private:` section on line 59 — could you merge the two into one?

zjw1111 (commented):
Thanks for adding the read-side `SharedShreddingFileReader` — the implementation is well-structured and the test coverage is solid. I left a few inline comments; one is about a potential null-pointer dereference in the error path, the rest are minor.

src/paimon/common/data/shredding/shared_shredding_file_reader_test.cpp
lszskye: > consider use `TETS_P`

will use `TETS_P` once zhanyu pr(supporting for query nested type for parquet) come in

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
lszskye: (1) From the FileBatchReader API contract, SetReadSchema() may be called multiple times(although we currently have no such usage). Therefore, it seems more appropriate to load immutable, file-level metadata such as shared_shredding_name_to_meta_ once during reader creation.
(2) The three maps have different lifetimes. shared_shredding_name_to_meta_ contains immutable file-level information, while the shared_shredding_name_to_selected_keys_ and shared_shredding_name_to_map_type_ are rebuilt for each SetReadSchema() call. Combining them into a single per-read struct would mix file-level state with read-specific state.

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
lszskye: I agree

lxy-9602:
For follow-up commits related to shredding, please include the issue link so others can trace the context more easily. Also, please rewriter pr description and use the paimon-cpp PR template.

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
```diff
@@ -160,22 +142,25 @@ Status SharedShreddingFileReader::SetReadSchema(
...
-
     shared_shredding_name_to_selected_keys_.clear();
     shared_shredding_name_to_map_type_.clear();
     arrow::FieldVector resolved_fields = logical_read_schema->fields();
     for (const auto& name : shared_shredding_names) {
         const auto& field = logical_read_schema->GetFieldByName(name);
         if (!field) {
             return Status::Invalid(
-                fmt::format("cannot find shared-shredding field {} in read schema", field->name()));
+                fmt::format("cannot find shared-shredding field in read schema"));
```
lxy-9602, line 163: fmt::format("cannot find shared-shredding field {} in read schema", name));

lxy-9602 (approved):
+1

src/paimon/common/data/shredding/shared_shredding_file_reader.cpp
lxy-9602: Next pr please.

PR #377, opened 2026-06-17
feat(shredding): improve shared-shredding adaptive width restore & refactor write

lxy-9602:
### Purpose

Linked issue: #342 
This PR improves shared-shredding MAP write behavior around adaptive width selection and writer initialization.

Main changes:

- Adjust adaptive K selection to avoid one very wide MAP row forcing all following files to use max width.
- Restore shared-shredding K from recent data files when creating new writers.
- Support restore across partial-write/data-evolution files with multiple shared-shredding MAP columns.
- Fall back to configured max-columns when a shared-shredding MAP column has no restorable metadata.
- Refactor data file writer creation into dedicated writer factory classes for append, key-value, blob, and shared-shredding writers.

### Tests
Added/updated staged tests include:

- `MapSharedShreddingContextTest.ComputeAdaptiveWidthUsesPercentileUnlessMaxIsClose`
- `MapSharedShreddingUtilsTest.BuildPhysicalStructTypePreservesFieldMetadata`
- `AppendOnlyWriterShreddingTest.TestSharedShreddingMapAllNullThenAllEmptyFiles`
- `AppendOnlyWriterShreddingTest.TestSharedShreddingMapAllEmptyFirstFile`
- `AppendOnlyWriterShreddingTest.TestSharedShreddingMapNullOverflow`
- `AppendOnlyFileStoreWriteTest.TestSharedShreddingMapRestoreInitializesNextWriter`
- `AppendOnlyFileStoreWriteTest.TestSharedShreddingRestoreIgnoresAvroFileWithoutMetadata`
- `AppendOnlyFileStoreWriteTest.TestSharedShreddingRestoreMultipleMapColumns`
- `AppendOnlyFileStoreWriteTest.TestSharedShreddingRestoreUsesDefaultForMissingMap`
- `KeyValueFileStoreWriteTest.TestSharedShreddingMapRestoreInitializesNextWriter`
- `MergeTreeWriterTest.TestSharedShreddingMapRestoreInitializesNextWriter`
- `PostponeBucketWriterTest.TestSharedShreddingMap`

### Generative AI tooling
Generated-by: Codex (GPT-5)


src/paimon/core/io/map_shared_shredding_core_utils.cpp
```diff
@@ -0,0 +1,141 @@
...
+        }
+        if (candidate_fields.empty()) {
+            continue;
+        }
+
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Schema> file_schema,
+                               ReadFileSchema(file, path_factory, options, pool));
+        for (const auto& field_name : candidate_fields) {
+            std::shared_ptr<arrow::Field> field = file_schema->GetFieldByName(field_name);
+            if (!field) {
```
zjw1111, line 101: In a data-evolution scenario (e.g., a new MAP column is added to the schema after some files have already been written), old data files legitimately do not contain the newly added field. When an old file has no `write_cols` recorded (`file->write_cols` is empty), all `pending_fields` become candidates — including fields that don't exist in the old file's schema. This causes the function to return `Status::Invalid` and fail writer creation.

Could you change this to `continue` instead of returning an error? That would be consistent with how the code already handles the "field exists but has no metadata" case a few lines below (lines 107-109), and would make the restore logic more resilient to mixed-schema files.

zjw1111 (commented):
Thanks for adding the restore-from-recent-files logic — it's a great improvement over the previous "always start cold" approach.

lszskye (approved):
+1

PR #380, opened 2026-06-23
fix: close input/output streams when remote lookup file read/write failed

lxy-9602:
### Purpose

No Linked issue.

Fix several review findings around stream cleanup safety, test assertions, and minor style issues.
- Add `ScopeGuard` cleanup for `RemoteLookupFileManager::CopyFromInputToOutput` to ensures input/output streams are closed on early returns from read/write/flush errors.
- Add a null check after `dynamic_cast` in `append_compaction_inte_test.cpp`
- Clean up style issues


zjw1111 (approved):
LGTM. Thanks for fixing the stream leak in `CopyFromInputToOutput` — the ScopeGuard approach is clean and correctly handles all early-return paths. The supplementary null checks and format-string fix are solid as well.

PR #383, opened 2026-06-26
fix: handle non-contiguous RowRanges when resolving global row IDs

zhf999:
### Purpose

- Fix a correctness bug where upper layers assumed returned batch rows are continuous and derived file row IDs as `previous_batch_start + offset`.
- Introduce unified usage of `GetPreviousBatchFileRowId(batch_row_id)` to resolve the file row ID for a row index inside the current batch.
- In `PrefetchFileBatchReaderImpl`, cache the actual file row IDs for each returned batch and keep row-id mapping aligned when a batch is sliced by `read_range`.
- For Parquet, add target row-range union and per-batch row mapping logic:
  - Merge filtered target ranges into a file target row set.
  - Build `batch_row_id -> file_row_id` mapping in `NextBatch()`.
  - Keep `GetPreviousBatchFileRowId()` correct under non-contiguous rows caused by predicate + bitmap filtering.
- Update upper-layer consumers to query per-row file IDs directly (deletion vectors, bitmap index filtering, `_ROW_ID` field conversion, KeyValue iteration positions).

### Tests

- Updated and adapted reader tests to the new interface and semantics:
  - `src/paimon/format/avro/avro_file_batch_reader_test.cpp`
  - `src/paimon/format/blob/blob_file_batch_reader_test.cpp`
  - `src/paimon/format/lance/lance_format_reader_writer_test.cpp`
  - `src/paimon/format/orc/orc_file_batch_reader_test.cpp`
  - `src/paimon/format/parquet/parquet_file_batch_reader_test.cpp`
  - `src/paimon/common/reader/prefetch_file_batch_reader_impl_test.cpp`
- Added/extended coverage in Parquet with `TestRowMapping` to validate file row mapping across non-contiguous ranges.
- Remaining changes are interface migration plus consistency updates of call sites and assertions.

### API and Format

- Public reader API change: `FileBatchReader` and implementations now use `GetPreviousBatchFileRowId(uint64_t batch_row_id)`.
- Semantic change: returns the file row ID for the `batch_row_id` inside the current batch (instead of deriving by batch start + offset under contiguous assumptions).
- Storage format and on-disk protocol are unchanged.

### Documentation
No.

### Generative AI tooling

gpt-5.3-codex


include/paimon/reader/file_batch_reader.h
```diff
@@ -46,8 +46,8 @@ class PAIMON_EXPORT FileBatchReader : public BatchReader {
     using BatchReader::NextBatch;
     using BatchReader::NextBatchWithBitmap;
 
-    /// Get the row number of the first row in the previously read batch.
-    virtual Result<uint64_t> GetPreviousBatchFirstRowNumber() const = 0;
+    /// Get the global row number of the row in the previously read batch.
+    virtual Result<uint64_t> GetPreviousBatchGlobalRowId(uint64_t batch_row_id) const = 0;
```
lxy-9602, line 57: `GetPreviousBatchGlobalRowId` -> `GetPreviousBatchFileRowId`

src/paimon/common/file_index/bitmap/apply_bitmap_index_batch_reader.h
```diff
@@ -95,11 +95,11 @@ class ApplyBitmapIndexBatchReader : public FileBatchReader {
...
-        PAIMON_ASSIGN_OR_RAISE(int32_t start_pos, reader_->GetPreviousBatchFirstRowNumber());
-        int32_t length = batch_size;
-        for (auto iter = bitmap_.EqualOrLarger(start_pos);
-             iter != bitmap_.End() && *iter < start_pos + length; ++iter) {
-            is_valid.Add(*iter - start_pos);
+        for (int32_t i = 0; i < batch_size; ++i) {
+            PAIMON_ASSIGN_OR_RAISE(uint64_t global_row_id, reader_->GetPreviousBatchGlobalRowId(i));
+            if (bitmap_.Contains(global_row_id)) {
+                is_valid.Add(i);
+            }
```
lxy-9602, line 111: This looks inefficient. Contains starts scanning from the beginning on every call, which can be quite costly for large bitmaps. In this case, since the iterator only moves forward, it should be enough to adjust the filter logic to take advantage of that.

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -571,8 +598,19 @@ Result<std::unique_ptr<::ArrowSchema>> PrefetchFileBatchReaderImpl::GetFileSchem
     return readers_[0]->GetFileSchema();
 }
 
-Result<uint64_t> PrefetchFileBatchReaderImpl::GetPreviousBatchFirstRowNumber() const {
-    return previous_batch_first_row_num_;
+Result<uint64_t> PrefetchFileBatchReaderImpl::GetPreviousBatchGlobalRowId(
+    uint64_t batch_row_id) const {
```
lxy-9602, line 606: Why can’t we just return `previous_batch_first_row_num_` + `batch_row_id` directly?

src/paimon/common/reader/prefetch_file_batch_reader_impl_test.cpp
```diff
@@ -636,12 +633,13 @@ TEST_F(PrefetchFileBatchReaderImplTest, TestPartialReaderSuccessRead) {
...
-    ASSERT_EQ(std::numeric_limits<uint64_t>::max(),
-              reader->GetPreviousBatchFirstRowNumber().value());
+    ASSERT_NOK(reader->GetPreviousBatchGlobalRowId(0));
     ASSERT_OK_AND_ASSIGN(auto batch_with_bitmap, reader->NextBatchWithBitmap());
     auto& [batch, bitmap] = batch_with_bitmap;
     ASSERT_EQ(batch.first->length, bitmap.Cardinality());
-    ASSERT_EQ(reader->GetPreviousBatchFirstRowNumber().value(), 0);
+    uint64_t global_row_id = 0;
+    ASSERT_OK_AND_ASSIGN(global_row_id, reader->GetPreviousBatchGlobalRowId(0));
+    ASSERT_EQ(global_row_id, 0);
```
lxy-9602: Using `.value()` directly should also work here.

src/paimon/common/reader/prefetch_file_batch_reader_impl_test.cpp
```diff
@@ -734,17 +728,16 @@ TEST_F(PrefetchFileBatchReaderImplTest, TestCallNextBatchAfterReadingEof) {
...
             /*initialize_read_ranges=*/true, /*prefetch_cache_mode=*/PrefetchCacheMode::ALWAYS,
             CacheConfig(), GetDefaultPool()));
-    ASSERT_EQ(std::numeric_limits<uint64_t>::max(),
-              reader->GetPreviousBatchFirstRowNumber().value());
+    ASSERT_NOK(reader->GetPreviousBatchGlobalRowId(0));
     ASSERT_OK_AND_ASSIGN(auto result_array,
                          ReadResultCollector::CollectResult(
                              reader.get(), /*max simulated data processing time*/ 100));
-    ASSERT_EQ(reader->GetPreviousBatchFirstRowNumber().value(), 10);
     auto expected_array = std::make_shared<arrow::ChunkedArray>(data_array);
```
lxy-9602, line 754: If we drop the final validation, this scenario would no longer be exercised by the test. Could we mimic `ReadResultCollector::CollectResult` inside this test instead? We don’t need to materialize the full result — the goal is just to validate the last row of the final batch.

src/paimon/common/reader/prefetch_file_batch_reader_impl_test.cpp
```diff
@@ -923,4 +914,56 @@ TEST_F(PrefetchFileBatchReaderImplTest, TestPrefetchWithBitmap) {
...
+    ASSERT_OK_AND_ASSIGN(global_row_id, reader->GetPreviousBatchGlobalRowId(5));
+    ASSERT_EQ(global_row_id, 25);
+
+    ASSERT_OK_AND_ASSIGN(batch,
+                         paimon::test::ReadResultCollector::CollectResultOneBatch(reader.get()));
+    ASSERT_OK_AND_ASSIGN(global_row_id, reader->GetPreviousBatchGlobalRowId(0));
+    ASSERT_EQ(global_row_id, 70);
+    ASSERT_OK_AND_ASSIGN(global_row_id, reader->GetPreviousBatchGlobalRowId(5));
+    ASSERT_EQ(global_row_id, 75);
+
```
lxy-9602, line 965: Could we just add a for loop to test `GetPreviousBatchGlobalRowId` for every element from the first to the last?

src/paimon/core/deletionvectors/apply_deletion_vector_batch_reader.h
```diff
@@ -96,9 +96,15 @@ class ApplyDeletionVectorBatchReader : public FileBatchReader {
 
  private:
     Result<RoaringBitmap32> Filter(int32_t batch_size) const {
-        PAIMON_ASSIGN_OR_RAISE(uint64_t previous_batch_first_row_number,
-                               reader_->GetPreviousBatchFirstRowNumber());
-        return deletion_vector_->IsValid(previous_batch_first_row_number, batch_size);
+        RoaringBitmap32 is_valid;
```
lxy-9602, line 99: Is `DeletionVector::IsValid(previous_batch_first_row_number, batch_size)` unused at this point? If it’s no longer needed, we should probably remove it.

src/paimon/core/io/complete_row_tracking_fields_reader.cpp
```diff
@@ -86,15 +86,14 @@ CompleteRowTrackingFieldsBatchReader::NextBatchWithBitmap() {
...
+        auto row_id_convert_func = [this](int32_t idx_in_array) -> Result<int64_t> {
             if (first_row_id_ == std::nullopt) {
                 return Status::Invalid(
                     "unexpected: read _ROW_ID special field, but first row id is null in meta");
             }
-            return first_row_id_.value() + previous_batch_first_row_number + idx_in_array;
+            PAIMON_ASSIGN_OR_RAISE(uint64_t global_row_id,
+                                   reader_->GetPreviousBatchGlobalRowId(idx_in_array));
+            return first_row_id_.value() + global_row_id;
         };
```
lxy-9602, line 97: The name global row id doesn’t seem quite appropriate in this scenario. It usually implies a row ID at the table level, while here `global_row_id` actually refers to a row ID within the file.

src/paimon/testing/mock/mock_file_batch_reader.h
```diff
@@ -156,8 +156,8 @@ class MockFileBatchReader : public PrefetchFileBatchReader {
         return metrics;
     }
 
-    Result<uint64_t> GetPreviousBatchFirstRowNumber() const override {
-        return ToReaderRowNumber(previous_batch_first_row_num_);
+    Result<uint64_t> GetPreviousBatchGlobalRowId(uint64_t batch_row_id) const override {
+        return previous_batch_first_row_num_ + batch_row_id;
```
lxy-9602, line 163: If `previous_batch_first_row_num_ `can be -1, then the current implementation of `GetPreviousBatchGlobalRowId` would be incorrect, which means the refactoring is not behaviorally equivalent.

src/paimon/testing/utils/read_result_collector.h
```diff
@@ -131,6 +131,53 @@ class ReadResultCollector {
         return chunk_array;
     }
 
+    static Result<std::shared_ptr<arrow::ChunkedArray>> CollectResultOneBatch(
+        BatchReader* batch_reader) {
+        return CollectResultOneBatch(batch_reader, /*max simulated data processing time*/ 0);
```
lxy-9602: `/*max_data_processing_time_in_us=*/`

src/paimon/testing/utils/read_result_collector.h
```diff
@@ -131,6 +131,53 @@ class ReadResultCollector {
...
+    }
+
+    static Result<std::shared_ptr<arrow::ChunkedArray>> CollectResultOneBatch(
+        BatchReader* batch_reader, int64_t max_data_processing_time_in_us) {
+        int64_t seed = DateTimeUtils::GetCurrentUTCTimeUs();
```
lxy-9602, line 116: Can `CollectResultOneBatch` just return an Arrow array directly? It doesn’t look like we need a `ChunkedArray` here.

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -526,4 +533,53 @@ Result<std::vector<int32_t>> ParquetFileBatchReader::ComputeNestedColumnIndices(
     return indices;
 }
 
+Result<RowRanges> ParquetFileBatchReader::GetAllTargetRowRanges(
+    const std::vector<TargetRowGroup>& target_row_groups) {
+    row_mapping_.clear();
```
lxy-9602, line 536: Please rename this to something like `UpdateAllTargetRowRanges`. Methods prefixed with `Get` should ideally avoid modifying member variables. Also, update doesn’t need to return anything here — it can just update `all_row_ranges_` directly.

include/paimon/reader/file_batch_reader.h
zhf999: Done.

src/paimon/common/file_index/bitmap/apply_bitmap_index_batch_reader.h
zhf999: Done, now we use an iterator to get the final result.

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
zhf999: The `PrefetchFileBatchReaderImpl` may hold `ParquetFileBatchReader`, which may return contenation of two discontinuous batch. Should we fallback `PrefetchFileBatchReaderImpl::GetPreviousBatchGlobalRowId` to simply return `previous_batch_first_row_num_ `+ `batch_row_id ` like `LanceFileBatchReader` or `BlobFileBatchReader`?

src/paimon/common/reader/prefetch_file_batch_reader_impl_test.cpp
zhf999: `MockFormatReaderBuilder` returns a batch with a random size, we can't assert the file row IDs for any batch. We can't call `GetPreviousBatchGlobalRowId` after reading EOF, either. We can only call `ASSERT_NOK(reader->GetPreviousBatchGlobalRowId(0));` here.

src/paimon/common/reader/prefetch_file_batch_reader_impl_test.cpp
zhf999: Now we use `.value()` directly to validate the result.

src/paimon/common/reader/prefetch_file_batch_reader_impl_test.cpp
zhf999: Done.

src/paimon/core/deletionvectors/apply_deletion_vector_batch_reader.h
zhf999: It has been removed.

src/paimon/core/io/complete_row_tracking_fields_reader.cpp
zhf999: The  variable name  has been changed to `file_row_id`.

src/paimon/testing/utils/read_result_collector.h
zhf999: `CollectResultOneBatch` is designed to align with `CollectResult`. Should we implement `CollectResultOneBatch` with a different return value?

src/paimon/format/parquet/parquet_file_batch_reader.cpp
zhf999: Done.

src/paimon/format/orc/orc_file_batch_reader_test.cpp
```diff
@@ -493,11 +493,10 @@ TEST_P(OrcFileBatchReaderTest, TestNextBatchSimple) {
     for (auto batch_size : {1, 2, 3, 5, 8, 10}) {
         auto orc_batch_reader =
             PrepareOrcFileBatchReader(file_name, &read_schema, batch_size, natural_read_size);
-        ASSERT_EQ(std::numeric_limits<uint64_t>::max(),
-                  orc_batch_reader->GetPreviousBatchFirstRowNumber().value());
+        ASSERT_EQ(orc_batch_reader->GetPreviousBatchFileRowId(0).value(), -1);
```
zjw1111: why -1 here? Status::Invalid?

src/paimon/testing/utils/read_result_collector.h
```diff
@@ -131,6 +131,53 @@ class ReadResultCollector {
         return chunk_array;
     }
 
+    static Result<std::shared_ptr<arrow::ChunkedArray>> CollectResultOneBatch(
+        BatchReader* batch_reader) {
+        return CollectResultOneBatch(batch_reader, /*max_simulated_data_processing_time*/ 0);
+    }
+
+    static Result<std::shared_ptr<arrow::ChunkedArray>> CollectResultOneBatch(
```
zjw1111, line 114: It's very similar with `CollectResult`, can you refactor to extract the common parts?

src/paimon/format/parquet/parquet_file_batch_reader.h
```diff
@@ -94,9 +96,18 @@ class ParquetFileBatchReader : public PrefetchFileBatchReader {
     Result<std::vector<std::pair<uint64_t, uint64_t>>> GenReadRanges(
         bool* need_prefetch) const override;
 
-    Result<uint64_t> GetPreviousBatchFirstRowNumber() const override {
-        assert(reader_);
-        return reader_->GetPreviousBatchFirstRowNumber();
+    Result<uint64_t> GetPreviousBatchFileRowId(uint64_t batch_row_id) const override {
```
zjw1111, line 99: change the interface in FileReaderWrapper together

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -213,11 +214,15 @@ Status ParquetFileBatchReader::SetReadSchema(
...
             } else {
-                target_row_groups.emplace_back(/*rg_index=*/rg_id,
-                                               /*is_partially_matched=*/false,
-                                               /*ranges=*/RowRanges());
+                target_row_groups.emplace_back(
+                    /*rg_index=*/rg_id,
+                    /*is_partially_matched=*/false,
+                    /*ranges=*/
+                    RowRanges(Range(0, reader_->GetAllRowGroupRanges()[rg_id].second -
+                                           reader_->GetAllRowGroupRanges()[rg_id].first - 1)));
```
zjw1111, line 220: format it

```suggestion
                target_row_groups.emplace_back(
                    /*rg_index=*/rg_id, /*is_partially_matched=*/false, /*ranges=*/
                    RowRanges(Range(0, reader_->GetAllRowGroupRanges()[rg_id].second -
                                           reader_->GetAllRowGroupRanges()[rg_id].first - 1)));
```

src/paimon/common/reader/prefetch_file_batch_reader_impl_test.cpp
```diff
@@ -607,12 +605,11 @@ TEST_F(PrefetchFileBatchReaderImplTest, TestReadWithLargeBatchSize) {
...
             /*initialize_read_ranges=*/true, /*prefetch_cache_mode=*/PrefetchCacheMode::ALWAYS,
             CacheConfig(), GetDefaultPool()));
-    ASSERT_EQ(std::numeric_limits<uint64_t>::max(),
-              reader->GetPreviousBatchFirstRowNumber().value());
+    ASSERT_NOK(reader->GetPreviousBatchFileRowId(0));
     ASSERT_OK_AND_ASSIGN(auto result_array,
                          ReadResultCollector::CollectResult(
                              reader.get(), /*max simulated data processing time*/ 100));
-    ASSERT_EQ(reader->GetPreviousBatchFirstRowNumber().value(), 101);
+    ASSERT_NOK(reader->GetPreviousBatchFileRowId(0));
```
zjw1111: Now that the interface has been modified, these calls will always trigger ASSERT_NOK, so there's no point in testing them anymore, right? There seem to be similar issues in other test files as well.

include/paimon/reader/file_batch_reader.h
```diff
@@ -46,8 +46,8 @@ class PAIMON_EXPORT FileBatchReader : public BatchReader {
     using BatchReader::NextBatch;
     using BatchReader::NextBatchWithBitmap;
 
-    /// Get the row number of the first row in the previously read batch.
-    virtual Result<uint64_t> GetPreviousBatchFirstRowNumber() const = 0;
+    /// Get the global row number of the row in the previously read batch.
+    virtual Result<uint64_t> GetPreviousBatchFileRowId(uint64_t batch_row_id) const = 0;
```
zjw1111, line 56: Are there any explicit semantic constraints on this interface before reading begins and after EOF is reached? As per previous discussions, it returns Status::Invalid before reading starts. However, after reaching EOF, the behavior currently varies wildly—some continue to accumulate, while others return errors. Should we impose some constraints on this? cc @lxy-9602

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -409,42 +423,54 @@ Status PrefetchFileBatchReaderImpl::EnsureReaderPosition(
...
+                                   readers_[reader_idx]->GetPreviousBatchFileRowId(i));
+            global_row_ids.push_back(global_row_id);
+        }
+        if (global_row_ids.empty()) {
+            ReaderUtils::ReleaseReadBatch(std::move(read_batch));
+            return Status::OK();
+        }
+        auto [slice_begin, slice_end] = ComputeBatchSliceByReadRange(global_row_ids, read_range);
+        if (slice_begin >= slice_end) {
+            readers_pos_[reader_idx]->store(read_range.second);
```
zjw1111: This is not a correctness issue, but when a batch is fully outside the current read range, advancing `readers_pos_` only to `read_range.second` instead of the first returned row id may cause extra range retries, so please pay attention to the potential performance impact.

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -40,6 +40,19 @@ class Schema;
 
 namespace paimon {
 
+namespace {
+
+std::pair<int64_t, int64_t> ComputeBatchSliceByReadRange(
```
zjw1111, line 45: Could we make `ComputeBatchSliceByReadRange` only return the end offset, since slicing an Arrow array with a non-zero offset may cause issues when converting it to a RecordBatch?
or `assert(slice_begin == 0)`, or return `Status::Invalid`, or only return the end offset, all OK, it's up to you.

src/paimon/format/lance/lance_file_batch_reader.cpp
```diff
@@ -130,7 +130,7 @@ Result<BatchReader::ReadBatch> LanceFileBatchReader::NextBatch() {
     if (is_eof) {
         return BatchReader::MakeEofBatch();
```
zjw1111, line 132: It looks like `previous_batch_row_count_` is not reset to 0 on the EOF path. The other four format readers (Avro, ORC, Blob, Parquet) all reset it here, so `GetPreviousBatchFileRowId()` returns `Status::Invalid` after EOF. Without the reset, the Lance reader still returns a seemingly valid but semantically incorrect row ID (pointing past the end of the file).

The corresponding test assertion at `lance_format_reader_writer_test.cpp:1065` (`ASSERT_EQ(6, ...)`) also appears to carry over the old expectation — would it be possible to add `previous_batch_row_count_ = 0;` before this return and update the test to `ASSERT_NOK(...)` for consistency?

src/paimon/format/parquet/parquet_file_batch_reader.h
```diff
@@ -173,6 +188,8 @@ class ParquetFileBatchReader : public PrefetchFileBatchReader {
         const std::shared_ptr<arrow::Schema>& read_schema,
         const std::shared_ptr<arrow::Schema>& file_schema);
 
+    Status UpdateAllTargetRowranges(const std::vector<TargetRowGroup>& target_row_groups);
```
zjw1111: Tiny nit: per the project's PascalCase convention, this would read better as `UpdateAllTargetRowRanges` (capital 'R' in "Ranges").

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -571,8 +606,19 @@ Result<std::unique_ptr<::ArrowSchema>> PrefetchFileBatchReaderImpl::GetFileSchem
     return readers_[0]->GetFileSchema();
 }
 
-Result<uint64_t> PrefetchFileBatchReaderImpl::GetPreviousBatchFirstRowNumber() const {
-    return previous_batch_first_row_num_;
+Result<uint64_t> PrefetchFileBatchReaderImpl::GetPreviousBatchFileRowId(
+    uint64_t batch_row_id) const {
+    if (current_batch_global_row_ids_.size() == 0) {
```
zjw1111: Nit: Google C++ Style prefers `empty()` over `size() == 0` for readability. Same pattern appears in `parquet_file_batch_reader.h` (`row_mapping_.size() == 0`).

zjw1111 (commented):
Thanks for the well-structured fix! The approach of replacing `GetPreviousBatchFirstRowNumber()` with per-row `GetPreviousBatchFileRowId(batch_row_id)` is a clean solution to the non-contiguous row problem under page-level predicate pushdown. The Parquet `GenerateRowMapping` logic correctly handles range gaps, and the test coverage (`TestRowMapping*`) is thorough.

A few items below for your consideration.

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -409,28 +423,48 @@ Status PrefetchFileBatchReaderImpl::EnsureReaderPosition(
...
             PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Array> src_array,
                                               arrow::ImportArray(c_array.get(), c_schema.get()));
-            int32_t target_length = read_range.second - first_row_number;
-            auto array = src_array->Slice(/*offset=*/0, target_length);
+            auto array = src_array->Slice(0, slice_end);
             PAIMON_RETURN_NOT_OK_FROM_ARROW(
                 arrow::ExportArray(*array, c_array.get(), c_schema.get()));
-            bitmap.RemoveRange(target_length, src_array->length());
+            RoaringBitmap32 sliced_bitmap;
+            for (auto iter = bitmap.Begin(); iter != bitmap.End() && *iter < slice_end; ++iter) {
```
zjw1111: maybe can directly use `RemoveRange`

zjw1111 (dismissed):
LGTM

zjw1111 (commented):
Two minor non-blocking nits — feel free to address in a follow-up commit.

include/paimon/reader/file_batch_reader.h
```diff
@@ -46,8 +46,8 @@ class PAIMON_EXPORT FileBatchReader : public BatchReader {
     using BatchReader::NextBatch;
     using BatchReader::NextBatchWithBitmap;
 
-    /// Get the row number of the first row in the previously read batch.
-    virtual Result<uint64_t> GetPreviousBatchFirstRowNumber() const = 0;
+    /// Get the global row number of the row in the previously read batch.
```
zjw1111: Nit (non-blocking): now that all readers consistently return `Status::Invalid` before the first read and after EOF, it would be nice to document that contract here. Something like:

```cpp
/// Get the file-level row ID for a given batch-relative row index
/// in the previously read batch.
///
/// @param batch_row_id  Zero-based index within the current batch.
/// @return The corresponding file-level row ID, or Status::Invalid
///         if no batch has been read yet, the last batch was EOF,
///         or batch_row_id is out of range.
```

This would save future readers from having to inspect every implementation to learn the expected behavior.

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -409,28 +423,44 @@ Status PrefetchFileBatchReaderImpl::EnsureReaderPosition(
...
+        auto [slice_begin, slice_end] = ComputeBatchSliceByReadRange(global_row_ids, read_range);
+        // slice_begin should always be 0, records before read_range.first have been consumed or
+        // filtered out.
+        if (slice_begin != 0) {
+            return Status::Invalid(fmt::format("Slice begin is {}, which is not 0.", slice_begin));
+        }
 
-        if (first_row_number >= read_range.second) {
+        if (0 == slice_end) {
             // fully out of range, data before first_row_number has been filtered out
```
zjw1111: Nit (non-blocking): a couple of comments in this block still reference the old variable name `first_row_number`, which no longer exists after the refactor. For example this line says "data before first_row_number has been filtered out". Could you update these comments to match the new logic (e.g., reference `global_row_ids` or `slice_end` instead) in a follow-up?

src/paimon/testing/utils/read_result_collector.h
```diff
@@ -131,6 +102,35 @@ class ReadResultCollector {
         return chunk_array;
     }
 
+    static Result<std::shared_ptr<arrow::ChunkedArray>> CollectResult(BatchReader* batch_reader) {
+        return CollectResult(batch_reader, /*max simulated data processing time*/ 0);
+    }
```
lxy-9602, line 107: `/*max simulated data processing time*/` -> `/*max_simulated_data_processing_time=*/`

src/paimon/testing/utils/read_result_collector.h
```diff
@@ -131,6 +102,35 @@ class ReadResultCollector {
...
+
+    static Result<std::shared_ptr<arrow::ChunkedArray>> CollectResultOneBatch(
+        BatchReader* batch_reader) {
+        return CollectResultOneBatch(batch_reader, /*max_simulated_data_processing_time*/ 0);
+    }
```
lxy-9602, line 112: `/*max_simulated_data_processing_time*/`-> `/*max_simulated_data_processing_time=*/`

lxy-9602 (approved):
+1

zjw1111 (approved):

PR #386, opened 2026-06-29
feat(manifest): support snapshot mainifest cache

gripleaf:
### Purpose

  In our production workload, a Paimon table may have about 60k buckets. A batch of data is imported roughly every 15 minutes, and the interval may become longer. During query planning, one scan read and decoded about 4.89 million manifest entries with 16 threads, which took around 30 seconds, while only a small subset of entries was finally kept after pruning.

  This patch introduces a snapshot-level live manifest cache to reduce repeated manifest read and decode cost. The cache stores merged live manifest entries for snapshots. When scanning a newer snapshot, paimon-cpp looks up the latest cached snapshot not greater than the target snapshot. If it is the same snapshot, the scan can use it directly; otherwise paimon-
  cpp incrementally builds the target snapshot by reading intermediate delta manifests.

  The implementation reuses the existing byte-oriented `Cache` interface:
  - add `CacheKind::SNAPSHOT_LIVE_MANIFEST`;
  - use `table_path#branch` as the logical cache key;
  - store multiple snapshots in one serialized binary cache value, bounded by `scan.manifest-entry-cache.max-snapshots`;
  - enable the optimization when a cache is provided through `ScanContextBuilder::WithCache()` and `scan.manifest-entry-cache.max-snapshots` is greater than 0.

  This trades one in-memory serialize/deserialize path for avoiding remote manifest reads plus manifest decoding. In the expected case, live entries are much fewer than all historical manifest entries; even in a conservative case, the cache is useful as long as deserializing cached live entries is cheaper than reading manifest files from remote storage and
  decoding all required manifest entries again.

  ### Tests

  - `cmake --build build --target paimon-core-test -j2`
  - `./build/release/paimon-core-test --gtest_filter='FileStoreScanTest.TestSnapshotLiveManifestCache:FileStoreScanTest.TestSnapshotLiveManifestCacheUsesCacheKey:TableScanTest.*:CoreOptionsTest.TestDefaultValue:CoreOptionsTest.TestFromMap:CoreOptionsTest.TestInvalidCase'`
  - `git -c filter.lfs.process= -c filter.lfs.clean=cat -c filter.lfs.required=false diff --check`

  ### API and Format

  This change extends the public cache kind enum with `CacheKind::SNAPSHOT_LIVE_MANIFEST`.

  It removes the separate `scan.manifest-entry-cache.enabled` option. The cache is controlled by `scan.manifest-entry-cache.max-snapshots`: values greater than 0 enable the cache path when a cache is provided through `WithCache()`, and 0 disables it.

  It does not change table storage format, file format, or network protocol. The serialized snapshot live manifest bundle is an in-memory cache value only and can be rebuilt from manifest files if absent or evicted.

  ### Documentation

  Yes. Added/updated user guide documentation for snapshot live manifest cache under `docs/source/user_guide/manifest_entry_cache.rst`.

  ### Generative AI tooling

  Generated-by: Codex (GPT-5)

### DIFF
--- a/src/paimon/core/manifest/snapshot_live_manifest_entries.cpp
+++ b/src/paimon/core/manifest/snapshot_live_manifest_entries.cpp
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2026-present Alibaba Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "paimon/core/manifest/snapshot_live_manifest_entries.h"
+
+#include <algorithm>
+#include <cstring>
+#include <utility>
+
+#include "paimon/common/io/memory_segment_output_stream.h"
+#include "paimon/core/manifest/manifest_entry_serializer.h"
+#include "paimon/io/byte_array_input_stream.h"
+#include "paimon/io/data_input_stream.h"
+#include "paimon/memory/bytes.h"
+#include "paimon/memory/memory_pool.h"
+#include "paimon/memory/memory_segment.h"
+
+namespace paimon {
+namespace {
+
+constexpr int32_t kMagic = 0x534d4543;  // SMEC
+
+size_t NormalizeMaxSnapshots(int32_t max_snapshots) {
+    return static_cast<size_t>(std::max(0, max_snapshots));
+}
+
+std::shared_ptr<Bytes> ToBytes(const MemorySegmentOutputStream& out,
+                               const std::shared_ptr<MemoryPool>& pool) {
+    auto bytes = Bytes::AllocateBytes(static_cast<size_t>(out.CurrentSize()), pool.get());
+    int64_t offset = 0;
+    for (const auto& segment : out.Segments()) {
+        int64_t copy_size =
+            std::min<int64_t>(segment.Size(), static_cast<int64_t>(bytes->size()) - offset);
+        if (copy_size <= 0) {
+            break;
+        }
+        std::memcpy(bytes->data() + offset, segment.Data(), static_cast<size_t>(copy_size));
+        offset += copy_size;
+    }
+    return bytes;
+}
+
+}  // namespace
+
+SnapshotLiveManifestEntries::SnapshotLiveManifestEntries(int32_t max_snapshots)
+    : max_snapshots_(max_snapshots) {}
+
+std::optional<SnapshotLiveManifestEntries::Entry> SnapshotLiveManifestEntries::LatestBeforeOrEqual(
+    int64_t snapshot_id) const {
+    auto iter = entries_by_snapshot_.upper_bound(snapshot_id);
+    if (iter == entries_by_snapshot_.begin()) {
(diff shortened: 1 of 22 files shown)


src/paimon/core/manifest/snapshot_live_manifest_entries.cpp
```diff
@@ -0,0 +1,144 @@
...
+    MemorySegmentOutputStream out(MemorySegmentOutputStream::DEFAULT_SEGMENT_SIZE, pool);
+    out.WriteValue<int32_t>(kMagic);
+    out.WriteValue<int32_t>(kVersion);
+    out.WriteValue<int32_t>(static_cast<int32_t>(entries_by_snapshot_.size()));
+
+    ManifestEntrySerializer serializer(pool);
+    for (const auto& [snapshot_id, entries] : entries_by_snapshot_) {
+        out.WriteValue<int64_t>(snapshot_id);
+        PAIMON_RETURN_NOT_OK(serializer.SerializeList(*entries, &out));
+    }
```
lxy-9602, line 94: I’d like to understand the need for versioning in the serialization/deserialization of `SnapshotLiveManifestEntries`. Is the expectation that it may be persisted or transferred across module boundaries? My understanding is probably not, since both parsing and toBytes seem to be handled entirely within `SnapshotLiveManifestEntries` itself.

My concern is that once we introduce a version here, any format change in a nested field of `ManifestEntry`—for example `DataFileMeta`—could require bumping the outer `SnapshotLiveManifestEntries` version as well to preserve cross-language compatibility. Right now, we only carry this kind of version information for protocols like split and commit message, so I want to confirm whether the version here is actually necessary.

src/paimon/core/operation/append_only_file_store_scan_test.cpp
```diff
@@ -175,4 +177,74 @@ TEST(AppendOnlyFileStoreScanTest, TestScanDurationMetric) {
...
+    context_builder.AddOption(Options::FILE_FORMAT, "orc")
+        .AddOption(Options::MANIFEST_FORMAT, "orc")
+        .AddOption(Options::SCAN_MANIFEST_ENTRY_CACHE_MAX_SNAPSHOTS, "8")
+        .WithCache(cache);
+    if (bucket) {
+        context_builder.SetBucketFilter(bucket.value());
+    }
+    auto scan_context_result = context_builder.Finish();
+    EXPECT_TRUE(scan_context_result.ok()) << scan_context_result.status().ToString();
+    auto table_scan_result = TableScan::Create(std::move(scan_context_result).value());
```
lxy-9602: Consider use `EXPECT_OK_AND_ASSIGN`.

test/inte/scan_inte_test.cpp
```diff
@@ -284,6 +284,107 @@ TEST(ScanInteManifestCacheTest, TestRepeatedScanReusesManifestCache) {
     ASSERT_EQ(supplier_calls_after_first_scan, manifest_cache->SupplierCallCount());
 }
 
+Result<std::vector<std::shared_ptr<DataSplitImpl>>> RunBucketSnapshotScan(
```
lxy-9602, line 290: To ensure correctness across orthogonal scan scenarios, please make the relevant scan tests parameterized.

In particular, convert the scan-inte/scan-and-read-inte/data-evolution tests that cover bucket pruning, predicate filtering, PK scans, data evolution scans, row-range scans, etc. to TEST_P, and run each case with snapshot live manifest cache both disabled and enabled.

This would make sure the new snapshot manifest cache path remains semantically equivalent to the existing non-cache path across the existing scan/filter matrix.

src/paimon/core/manifest/snapshot_live_manifest_entries.cpp
gripleaf: Yes, you are right. SnapshotLiveManifestEntries is an in-memory state, and there is no multi-version issue during its lifecycle. I will remove the version control logic.

lszskye (approved):
+1

PR #387, opened 2026-06-29
feat: support to read chain data split

### LINKED ISSUE #385: [Bug] Chain Table read failure
### Search before asking

### Paimon-cpp version

latest

### Minimal reproduce step

Read a Chain Table with paimon-cpp

### What doesn't meet your expectations?

It throws runtime error:
`invalid data split byte stream, remaining {} bytes after deserializing`

### Anything else?

https://paimon.apache.org/docs/master/primary-key-table/chain-table/

Weixin-Xu:
### Purpose

Linked issue: close #385

Support deserializing, serializing, and reading community-format ChainSplit.

This change adds ChainSplit support for the current Java SplitSerializer wire format, including:
- recognizing the SplitSerializer magic, serializer version, and CHAIN_SPLIT type
- deserializing ChainSplit fields: logical partition, data files, file bucket path mapping, and file branch mapping
- serializing ChainSplit back through SplitSerializer without dropping chain metadata
- resolving ChainSplit data file paths through per-file bucket path mapping
- failing fast when a non-external ChainSplit file is missing bucket path mapping
- preserving existing DataSplit, FallbackDataSplit, and IndexedSplit behavior

### Tests

- Added `ChainSplitTest`
  - deserialize community-format ChainSplit bytes
  - deserialize ChainSplit bytes with multiple data files
  - verify per-file bucket path resolution
  - verify external file path is preserved
  - verify missing bucket path mapping returns an error
  - verify ChainSplit serialization preserves bucket and branch mappings
  - verify malformed ChainSplit bytes return an error

### API and Format

This change does not modify table storage format. It updates split protocol support to handle the community Java ChainSplit format.

The read path consumes `fileBucketPathMapping` for per-file bucket path resolution. It also deserializes and preserves `fileBranchMapping` for format compatibility; branch-aware read/schema selection is intentionally deferred to a follow-up change.

### Documentation

No user-facing documentation update.
### Generative AI tooling

Generated-by: OpenAI Codex


src/paimon/core/operation/abstract_split_read.h
```diff
@@ -76,6 +76,9 @@ class AbstractSplitRead : public SplitRead {
     Result<std::unique_ptr<BatchReader>> ApplyPredicateFilterIfNeeded(
         std::unique_ptr<BatchReader>&& reader, const std::shared_ptr<Predicate>& predicate) const;
 
+    Result<std::shared_ptr<DataFilePathFactory>> CreateDataFilePathFactory(
```
zjw1111, line 80: Minor: `CreateDataFilePathFactory` is currently in the `public` section, but it seems to only be called by subclasses internally. Would it be possible to move it into the `protected` section just below?

src/paimon/core/table/source/chain_split_impl.h
```diff
@@ -0,0 +1,70 @@
...
+        return all_snapshot_split_;
+    }
+
+    const BinaryRow& ReadPartition() const {
+        return read_partition_;
+    }
+
+    const std::unordered_map<std::string, std::string>& FileBucketPathMapping() const {
+        return file_bucket_path_mapping_;
+    }
```
zjw1111, line 57: I noticed that `file_branch_mapping_` is deserialized and stored, but not consumed by any read path in this PR (e.g. `ChainDataFilePathFactory` only uses `file_bucket_path_mapping_`). I assume this is reserved for future use (e.g. selecting the correct schema per branch, as the Java side does). Could you add a brief note in the PR description mentioning this is intentionally deferred?

zjw1111 (commented):
Thanks for adding ChainDataSplit support! The overall design looks good. A few minor suggestions below.

src/paimon/core/operation/raw_file_split_read.cpp
```diff
@@ -66,17 +66,26 @@ Result<std::unique_ptr<BatchReader>> RawFileSplitRead::CreateReader(
...
-    DeletionVector::Factory dv_factory) {
+    DeletionVector::Factory dv_factory,
+    std::shared_ptr<DataFilePathFactory> data_file_path_factory) {
     const auto& predicate = context_->GetPredicate();
-    PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<DataFilePathFactory> data_file_path_factory,
-                           path_factory_->CreateDataFilePathFactory(partition, bucket));
+    if (!data_file_path_factory) {
+        PAIMON_ASSIGN_OR_RAISE(data_file_path_factory,
+                               path_factory_->CreateDataFilePathFactory(partition, bucket));
+    }
```
lxy-9602: This design feels a bit too implicit. Why can’t each caller just pass in the correct `data_file_path_factory` directly?

lxy-9602: Also, let’s avoid using default arguments in production code.

src/paimon/core/table/source/split.cpp
```diff
@@ -224,13 +326,23 @@ Result<std::shared_ptr<Split>> Split::Deserialize(const char* buffer, size_t len
         PAIMON_ASSIGN_OR_RAISE(int64_t stream_length, in.Length());
         if (pos == stream_length) {
             return data_split;
+        } else if (data_split->BucketPath() == ChainDataSplitImpl::VIRTUAL_BUCKET_PATH) {
+            auto chain_split = ReadChainDataSplitTail(data_split, &in, pool);
+            if (!chain_split.ok()) {
+                return Status::Invalid(fmt::format("invalid ChainDataSplit byte stream: {}",
+                                                   chain_split.status().ToString()));
+            }
+            return std::static_pointer_cast<Split>(chain_split.value());
```
lxy-9602: Could you please use `PAIMON_ASSIGN_OR_RAISE` rather than `if (!chain_split.ok())`.

src/paimon/core/table/source/split.cpp
```diff
@@ -129,6 +193,44 @@ Result<std::shared_ptr<DataSplitImpl>> ReadDataSplitWithoutMagicNumber(
...
+        PAIMON_ASSIGN_OR_RAISE(std::string key, in->ReadString());
+        PAIMON_ASSIGN_OR_RAISE(std::string value, in->ReadString());
+        result.emplace(std::move(key), std::move(value));
+    }
+    return result;
+}
+
+Result<std::shared_ptr<ChainDataSplitImpl>> ReadChainDataSplitTail(
+    const std::shared_ptr<DataSplitImpl>& base_split, DataInputStream* in,
+    const std::shared_ptr<MemoryPool>& pool) {
```
lxy-9602: Could you move the output parameter (`in`) in to the end of the parameter list?

src/paimon/core/table/source/split.cpp
```diff
@@ -37,6 +40,65 @@
 namespace paimon {
 struct DataFileMeta;
 namespace {
+Result<std::vector<std::shared_ptr<DataFileMeta>>> ReadVersion7DataFileMetaList(
+    DataInputStream* in, const std::shared_ptr<MemoryPool>& pool) {
```
lxy-9602, line 55: I found that the chain table split is currently not compatible with the Java implementation. Java uses `ChainSplitHeader + logicalPartition + files + bucketMap + branchMap`, while C++ uses `DataSplit + ChainTail`. The split format must be kept fully consistent with Java.

src/paimon/core/operation/abstract_split_read.h
Weixin-Xu: Fixed in the latest revision. `CreateDataFilePathFactory` is now protected since it is only used by `AbstractSplitRead` subclasses.

src/paimon/core/table/source/chain_split_impl.h
Weixin-Xu: Added a note in the API and Format section to clarify that `file_branch_mapping` is preserved for ChainDataSplit metadata compatibility, while branch-aware read/schema selection is deferred. The current read path only consumes `file_bucket_path_mapping`.

src/paimon/core/operation/raw_file_split_read.cpp
Weixin-Xu: Fixed. The callers now construct the appropriate `DataFilePathFactory` explicitly before delegating:
- split-based reads use `CreateDataFilePathFactory(data_split)`, so ChainDataSplit can get the chain-aware factory.
- partition/bucket reads create the normal factory with `path_factory_->CreateDataFilePathFactory(partition, bucket)`.

The private helper now requires a `DataFilePathFactory` argument directly, so there is no `nullptr` fallback and no default argument.

src/paimon/core/table/source/split.cpp
Weixin-Xu: Fixed. Moved `DataInputStream* in` to the end of `ReadChainDataSplitTail`'s parameter list and updated both call sites.

src/paimon/core/table/source/split.cpp
Weixin-Xu: Fixed. Replaced the manual `chain_split.ok()` checks with `PAIMON_ASSIGN_OR_RAISE` in both ChainDataSplit tail parsing paths.

src/paimon/core/table/source/split.cpp
Weixin-Xu: Yes, I was working from an outdated commit. I’ll realign the implementation with the current community Java format, then submit an updated patch after further changes and verification.

src/paimon/core/operation/abstract_split_read.cpp
```diff
@@ -116,6 +118,20 @@ Result<std::unique_ptr<BatchReader>> AbstractSplitRead::ApplyPredicateFilterIfNe
...
+    }
+
+    PAIMON_ASSIGN_OR_RAISE(
+        std::shared_ptr<DataFilePathFactory> base_factory,
+        path_factory_->CreateDataFilePathFactory(data_split->Partition(), data_split->Bucket()));
+    return base_factory;
+}
+
 Result<std::unique_ptr<ReaderBuilder>> AbstractSplitRead::PrepareReaderBuilder(
     const std::string& format_identifier) const {
```
lxy-9602, line 183: I think there is still a semantic gap with the Java ChainSplit read path.

The PR now deserializes and preserves `fileBranchMapping`, and uses `fileBucketPathMapping` to resolve each file path. However, Java also consumes `fileBranchMapping` when reading each file: `MergeFileSplitRead#createChainReader` builds a `ChainReadContext`, and `ChainKeyValueFileReaderFactory#getDataSchema` uses `chainReadContext.fileBranchMapping().get(fileMeta.fileName())` to choose the branch-specific `SchemaManager` before loading `fileMeta.schemaId()`.

In the C++ path, `AbstractSplitRead::CreateFieldMappingReader` still loads schemas from the current table/current branch only via `context_->GetTableSchema()` or `schema_manager_->ReadSchema(file_meta->schema_id)`. For a ChainSplit containing files from both snapshot and delta branches, the same `schema_id` may refer to different schemas across branches, or may not exist in the current branch at all. In that case C++ can read with the wrong schema or fail, while Java reads with the schema from the file’s own branch.

Could we either make ChainSplit reads branch-aware when resolving `DataFileMeta::schema_id`, using `ChainSplitImpl::FileBranchMapping()`, or explicitly fail fast for ChainSplits whose files belong to a non-current branch until branch-aware schema selection is supported?

src/paimon/core/io/chain_split_file_path_factory.h
```diff
@@ -0,0 +1,52 @@
...
+    static Result<std::shared_ptr<ChainSplitFilePathFactory>> Create(
+        const std::vector<std::shared_ptr<DataFileMeta>>& data_files,
+        std::unordered_map<std::string, std::string> file_bucket_path_mapping,
+        std::unordered_map<std::string, std::string> file_branch_mapping);
+
+    explicit ChainSplitFilePathFactory(
+        std::unordered_map<std::string, std::string> file_bucket_path_mapping);
+    ChainSplitFilePathFactory(std::unordered_map<std::string, std::string> file_bucket_path_mapping,
+                              std::unordered_map<std::string, std::string> file_branch_mapping);
+
```
lxy-9602, line 41: If the intended construction path for `ChainSplitFilePathFactory` in production code is through `Create`, could we make the other constructors private as well, to avoid external misuse?

src/paimon/core/io/chain_split_file_path_factory.cpp
```diff
@@ -0,0 +1,88 @@
...
+    if (file_meta->external_path) {
+        return file_meta->external_path.value();
+    }
+
+    return PathUtil::JoinPath(file_bucket_path_mapping_.at(file_meta->file_name),
+                              file_meta->file_name);
+}
+
+std::string ChainSplitFilePathFactory::ToAlignedPath(
+    const std::string& file_name, const std::shared_ptr<DataFileMeta>& aligned) const {
```
lxy-9602, line 71: There is already a Java class that serves a similar purpose. Could we keep the naming consistent and use `ChainReadDataFilePathFactory`, so we can avoid future inconsistencies and make maintenance easier?

src/paimon/core/operation/abstract_split_read.cpp
```diff
@@ -116,6 +120,65 @@ Result<std::unique_ptr<BatchReader>> AbstractSplitRead::ApplyPredicateFilterIfNe
...
+    auto it = branch_schema_managers_.find(normalized_branch);
+    if (it != branch_schema_managers_.end()) {
+        return it->second.get();
+    }
+
+    auto schema_manager = std::make_unique<SchemaManager>(options_.GetFileSystem(),
+                                                          context_->GetPath(), normalized_branch);
+    auto [inserted_it, _] =
+        branch_schema_managers_.emplace(normalized_branch, std::move(schema_manager));
+    return inserted_it->second.get();
```
lxy-9602, line 155: Returning a `const SchemaManager*` seems a little unusual here. Have you considered using a smart pointer to `SchemaManager` instead?

src/paimon/core/operation/abstract_split_read.cpp
```diff
@@ -116,6 +120,65 @@ Result<std::unique_ptr<BatchReader>> AbstractSplitRead::ApplyPredicateFilterIfNe
...
+        std::dynamic_pointer_cast<ChainSplitFilePathFactory>(data_file_path_factory);
+    if (chain_path_factory) {
+        std::optional<std::string> branch = chain_path_factory->BranchForFile(file_meta->file_name);
+        if (!branch) {
+            return Status::Invalid(
+                fmt::format("branch is missing for ChainSplit file {}", file_meta->file_name));
+        }
+        PAIMON_ASSIGN_OR_RAISE(schema_manager, GetSchemaManagerForBranch(branch.value()));
+        current_branch = schema_manager == schema_manager_.get();
+    }
```
lxy-9602, line 174: `current_branch = schema_manager == schema_manager_.get()` relies on raw pointer identity, which feels fairly error-prone and easy to misuse. Could we switch to comparing branch names instead?

src/paimon/core/table/source/chain_split_impl.h
```diff
@@ -0,0 +1,70 @@
...
+                   std::unordered_map<std::string, std::string>&& file_branch_mapping)
+        : DataSplitImpl(base_split->Partition(), base_split->Bucket(), base_split->BucketPath(),
+                        std::vector<std::shared_ptr<DataFileMeta>>(base_split->DataFiles())),
+          all_snapshot_split_(all_snapshot_split),
+          read_partition_(read_partition),
+          file_bucket_path_mapping_(std::move(file_bucket_path_mapping)),
+          file_branch_mapping_(std::move(file_branch_mapping)) {
+        CopyMetadataFrom(*base_split);
+    }
+
```
lxy-9602, line 46: `ChainSplitImpl` inheriting from `DataSplitImpl` looks convenient for reusing existing read paths, but it weakens the type semantics compared with Java, where `ChainSplit` only implements `Split`. The inherited DataSplit APIs expose fake bucket/bucketPath semantics and may produce wrong behavior if generic DataSplit code calls methods like `GetFileList()`. Please either introduce a common file-list/read-scope abstraction used by both DataSplitImpl and ChainSplitImpl, or at least override/disable inherited methods whose semantics are invalid for ChainSplit and document that ChainSplit is intentionally accepted by DataSplit read paths.

src/paimon/core/table/source/chain_split_impl.h
```diff
@@ -0,0 +1,70 @@
...
+    bool AllSnapshotSplit() const {
+        return all_snapshot_split_;
+    }
+
+    const BinaryRow& ReadPartition() const {
+        return read_partition_;
+    }
```
lxy-9602, line 53: Is `ReadPartition` intended to be equivalent to `ChainSplit.logicalPartition` on the Java side? If they represent the same concept, could we align the function name to keep things consistent?

src/paimon/core/table/source/split.cpp
```diff
@@ -37,6 +44,72 @@
...
+constexpr int32_t kSplitSerializerDataSplit = 1;
+constexpr int32_t kSplitSerializerIndexedSplit = 3;
+constexpr int32_t kSplitSerializerChainSplit = 4;
+constexpr int32_t kSplitSerializerFallbackDataSplit = 6;
+
+Result<std::vector<std::shared_ptr<DataFileMeta>>> ReadVersion7DataFileMetaList(
+    DataInputStream* in, const std::shared_ptr<MemoryPool>& pool) {
+    PAIMON_ASSIGN_OR_RAISE(int32_t size, in->ReadValue<int32_t>());
+    if (size < 0) {
+        return Status::Invalid(fmt::format("invalid data file meta list size: {}", size));
```
lxy-9602, line 58: Could you move the output parameter `DataInputStream*` to the end to keep the code style consistent?

src/paimon/core/table/source/split.cpp
```diff
@@ -37,6 +44,72 @@
...
+    }
+
+    DataFileMetaFirstRowIdLegacySerializer legacy_serializer(pool);
+    DataFileMetaSerializer current_serializer(pool);
+    std::vector<std::shared_ptr<DataFileMeta>> result;
```
lxy-9602, line 63: It looks like there are currently two split serialization paths on the Java side: one based on `SplitSerializer`, and another based on `ChainSplit.serialize` / `DataSplit.serialize`. From what I see, this PR implements the `SplitSerializer` variant, while the existing Flink/Spark java code still relies on `ChainSplit.serialize / DataSplit.serialize`. That makes me wonder whether this change could lead to compatibility issues.

src/paimon/core/table/source/split.cpp
lxy-9602: Also, I’m not entirely clear on the rationale for introducing `ReadVersion7DataFileMetaList `as a separate function. Is there a reason we can’t continue to use `DataSplitImpl::GetFileMetaSerializer` to get the appropriate versioned serializer here?

src/paimon/core/table/source/split.cpp
```diff
@@ -37,6 +44,72 @@
...
+
+Result<std::vector<std::shared_ptr<DataFileMeta>>> ReadDataFileMetaList(
+    int32_t version, const ObjectSerializer<std::shared_ptr<DataFileMeta>>* data_file_serializer,
+    DataInputStream* in, const std::shared_ptr<MemoryPool>& pool) {
+    if (version == 7) {
+        return ReadVersion7DataFileMetaList(in, pool);
+    }
+    return data_file_serializer->DeserializeList(in);
+}
+
```
lxy-9602, line 112: Thanks a lot for adding support for chain split. Since this change affects the communication protocol, we need to be a bit more careful with the review and merge process. So far, I’ve reviewed the some parts of the PR. Could you please address the comments first? After that, we can continue with the next round of review.

lxy-9602:
Thank you for your contribution!

This repository has now been sealed. We have contributed all existing functionality to [[apache/paimon-cpp](https://github.com/apache/paimon-cpp)](https://github.com/apache/paimon-cpp), which is now the official C++ implementation of Apache Paimon.

Therefore, this pull request will not be merged into this repository. The functionality and APIs of the official repository are consistent with this repository, so you are welcome to submit this change to [[apache/paimon-cpp](https://github.com/apache/paimon-cpp)](https://github.com/apache/paimon-cpp). Please also direct all future contributions there.

Thank you again for your work and understanding.

PR #392, opened 2026-06-30
feat(types): support CHAR/VARCHAR/BINARY/VARBINARY in data type json parser

SteNicholas:
### Purpose

Linked issue: #197 

`CHAR`, `VARCHAR`, `BINARY`, and `VARBINARY` were already registered as keywords
(in the `Keyword` enum and `Keywords()` map), but `ParseTypeByKeyword` had no
`case` for them, so they fell through to the `default` branch. Deserializing any
schema that contained them failed with:

```
deserialize failed, possibly type incompatible: parse data type failed, error msg: Invalid: Unsupported type: VARCHAR
```

This change maps them to the corresponding Arrow types, consistent with how
`STRING` and `BYTES` are handled:

- `CHAR` / `VARCHAR` -> `arrow::utf8()`
- `BINARY` / `VARBINARY` -> `arrow::binary()`

It reuses the existing `ParseStringType<T>()` helper. Arrow has no fixed-length
char/binary type, so the optional length parameter (e.g. the `(10)` in
`VARCHAR(10)`) is parsed but not retained on the Arrow type. The declared length
is validated to be within `[1, INT_MAX]` (both inclusive), consistent with Java
Paimon; out-of-range lengths such as `VARCHAR(0)` or `VARCHAR(2147483648)` are
rejected.

### Tests

`DataTypeJsonParserTest.ParseTypeAtomicTypeSuccess` is extended to cover each new
type with and without a length argument (`CHAR`, `CHAR(10)`, `VARCHAR`,
`VARCHAR(10)`, `BINARY`, `BINARY(10)`, `VARBINARY`, `VARBINARY(10)`), the
inclusive length boundaries (`CHAR(1)`, `VARCHAR(2147483647)`), and invalid
lengths (`VARCHAR(0)`, `VARBINARY(0)`, `VARCHAR(2147483648)`, plus the existing
`VARCHAR(test)`).

`WriteAndReadInteTest.TestCharVarcharBinaryVarbinaryTypes` adds an end-to-end
write/read case (parquet/orc): it creates a table, then evolves to a hand-written
schema whose columns declare `CHAR(10)`/`VARCHAR(20)`/`BINARY(10)`/`VARBINARY(20)`
(the Arrow-based serializer always renders string/binary as `STRING`/`BYTES`, so
the schema is written by hand), and writes then reads back rows — including a
`NULL` row — through the full Paimon write/commit/scan/read flow.

### API and Format

No public API (`include/`) or storage format/protocol change. This only broadens
schema deserialization to accept type strings that previously errored.

### Documentation

Yes. `docs/source/user_guide/data_types.rst` is updated to mark
`CHAR`/`VARCHAR` -> `Utf8` and `BINARY`/`VARBINARY` -> `Binary` as supported, with
a note that the declared length is not enforced.

### Generative AI tooling

Generated-by: Claude Code (Claude Opus 4.8)

### DIFF
--- a/src/paimon/common/types/data_type_json_parser_test.cpp
+++ b/src/paimon/common/types/data_type_json_parser_test.cpp
@@ -111,6 +111,16 @@ TEST(DataTypeJsonParserTest, ParseTypeAtomicTypeSuccess) {
         {"TIMESTAMP_LTZ(9)", arrow::timestamp(arrow::TimeUnit::NANO, timezone)},
         {"BYTES", arrow::binary()},
         {"STRING", arrow::utf8()},
+        {"CHAR", arrow::utf8()},
+        {"CHAR(10)", arrow::utf8()},
+        {"VARCHAR", arrow::utf8()},
+        {"VARCHAR(10)", arrow::utf8()},
+        {"BINARY", arrow::binary()},
+        {"BINARY(10)", arrow::binary()},
+        {"VARBINARY", arrow::binary()},
+        {"VARBINARY(10)", arrow::binary()},
+        {"CHAR(1)", arrow::utf8()},
+        {"VARCHAR(2147483647)", arrow::utf8()},
     };
 
     for (const auto& test_case : test_cases) {
@@ -130,6 +140,12 @@ TEST(DataTypeJsonParserTest, ParseTypeAtomicTypeSuccess) {
         rapidjson::Document invalid_doc;
         rapidjson::Value value("VARCHAR(test)", invalid_doc.GetAllocator());
         ASSERT_NOK(DataTypeJsonParser::ParseType("field_name", value));
+    }
+    for (const char* invalid_type : {"VARCHAR(0)", "VARBINARY(0)", "VARCHAR(2147483648)"}) {
+        rapidjson::Document invalid_doc;
+        rapidjson::Value value(invalid_type, invalid_doc.GetAllocator());
+        ASSERT_NOK_WITH_MSG(DataTypeJsonParser::ParseType("field_name", value),
+                            "length must be between 1 and 2147483647");
     }
     {
         rapidjson::Document invalid_doc;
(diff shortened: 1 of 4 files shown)


src/paimon/common/types/data_type_json_parser_test.cpp
```diff
@@ -111,6 +111,14 @@ TEST(DataTypeJsonParserTest, ParseTypeAtomicTypeSuccess) {
...
         {"STRING", arrow::utf8()},
+        {"CHAR", arrow::utf8()},
+        {"CHAR(10)", arrow::utf8()},
+        {"VARCHAR", arrow::utf8()},
+        {"VARCHAR(10)", arrow::utf8()},
+        {"BINARY", arrow::binary()},
+        {"BINARY(10)", arrow::binary()},
+        {"VARBINARY", arrow::binary()},
+        {"VARBINARY(10)", arrow::binary()},
     };
```
lxy-9602, line 124: In Java Paimon, there is a validation that `length >= 1 && length <= INT_MAX`; could you add the same check here?
Also, please add an end-to-end write/read case to confirm that the full Paimon read/write flow works as expected. The schema for table creation here may need to be written by hand.

SteNicholas: @lxy-9602, both points are addressed in the latest revision:

1. **Length validation** — `ParseStringLength` now parses the length as 64-bit and validates it is within `[1, INT_MAX]` (both inclusive), consistent with Java Paimon. Out-of-range lengths like `VARCHAR(0)` or `VARCHAR(2147483648)` are rejected with a clear error. `ParseTypeAtomicTypeSuccess` adds the inclusive boundaries (`CHAR(1)`, `VARCHAR(2147483647)`) as success cases and `VARCHAR(0)`/`VARBINARY(0)`/`VARCHAR(2147483648)` as failures.

2. **End-to-end write/read** — added `WriteAndReadInteTest.TestCharVarcharBinaryVarbinaryTypes`. As you noted, the table schema is written by hand (the Arrow-based serializer always renders string/binary as `STRING`/`BYTES`): the table is created, then evolved to a schema declaring `CHAR(10)`/`VARCHAR(20)`/`BINARY(10)`/`VARBINARY(20)`, and rows (including a `NULL` row) are written and read back through the full Paimon write/commit/scan/read flow.

lxy-9602 (dismissed):
+1

lxy-9602 (approved):
+1

lszskye (approved):
+1

PR #393, opened 2026-07-01
feat(blob): Remove blob external storage mode

### LINKED ISSUE #283: [Feature] Support multiple blob fields and BlobView/descriptor-inline mode in paimon-cpp
### Search before asking

### Motivation

Paimon Java already provides a rich set of Blob capabilities beyond basic single-field read/write:

1. **Multiple Blob Fields per Table** — Java's `MultipleBlobFileWriter` creates an independent blob file writer for each blob column in a row. When a table schema contains several BLOB columns (e.g., `image BLOB, video BLOB, embedding BLOB`), each column is written to its own blob file, and the normal data file stores only the non-blob columns. During reads, `DedicatedFormatRollingFileWriter` coordinates multiple blob file readers to merge results back into a complete row. In paimon-cpp, both `BlobFormatWriter` and `BlobFileBatchReader` enforce `num_fields == 1`, meaning only a single blob column is allowed per table.

2. **BlobView Mode (cross-table blob reference)** — Java supports `BlobView`, where a blob field does not store actual data but holds a `BlobViewStruct` that references a blob in another upstream table (identified by `Identifier` + `fieldId` + `rowId`). At read time, `BlobViewLookup` batch-preloads the `BlobDescriptor`s from the upstream table using row-range parallelism, and `BlobViewResolver` lazily resolves each `BlobView` into a readable `Blob`. This avoids duplicating large blob data across derived tables. paimon-cpp has no equivalent of `BlobView`, `BlobViewStruct`, or `BlobViewLookup`.

3. **Blob Descriptor-Inline Mode (`blob-descriptor-field`)** — Java allows certain blob fields to be designated as "descriptor fields" via the `blob-descriptor-field` table option. These blob fields are serialized as compact `BlobDescriptor` bytes (URI + offset + length) and stored inline within the normal data file (e.g., Parquet/ORC), rather than being split out to a separate `.blob` file. This is useful when the blob data is already externally managed (e.g., on OSS/S3) and only a lightweight reference needs to be persisted. `BlobType.fieldsInBlobFile()` / `fieldsNotInBlobFile()` partitions the schema accordingly. paimon-cpp does not implement this partitioning logic.

These gaps prevent paimon-cpp users from:
- Defining tables with more than one BLOB column.
- Leveraging cross-table blob references for data lineage / deduplication use cases.
- Using lightweight descriptor-inline storage for externally managed blob data.

### Solution

- Phase 1: Multiple Blob Fields Support
- Phase 2: Blob Descriptor-Inline Mode
- Phase 3: BlobView Mode

### Anything else?

_No response_

lxy-9602:
### Purpose

Linked issue: close #283 

Remove the `blob-external-storage-field` and `blob-external-storage-path` functionality, following apache/paimon#8378.

This change:
- Removes the external-storage blob table options from public option definitions and `CoreOptions`.
- Removes external-storage blob schema validation.
- Deletes `ExternalStorageBlobWriter` and its write-path integration from `AppendOnlyWriter`.
- Simplifies `BlobFileContext` to only classify descriptor, view, inline, and regular blob-file fields.
- Simplifies blob writer plumbing by removing the now-unused `WriteConsumer` and blob path creator hooks.
- Keeps existing descriptor/blob-view/raw blob behavior unchanged.
- Keeps historical test data, but removes the deleted options from fixture schema files.

### API and Format
This removes two public table options from the C++ API:
- `blob-external-storage-field`
- `blob-external-storage-path`

No storage format change is introduced. Existing `BlobDescriptor` data whose URI points outside the main table path can still be read as long as the referenced files exist.

### Generative AI tooling


test/test_data/orc/blob_desc_field.db/blob_desc_field/schema/schema-0
```diff
@@ -28,14 +28,12 @@
   "options" : {
```
lszskye, line 28: please change the name of this db, remove "external"

test/inte/blob_table_inte_test.cpp
lszskye, line 2991: please query `external` in test/inte/blob_table_inte_test.cpp

lszskye (approved):
+1

PR #394, opened 2026-07-01
feat(schema): scan context support set table schema

gripleaf:
### Purpose

  Add `ScanContextBuilder::SetTableSchema` so scan operations can reuse a pre-loaded table schema instead of always loading the latest schema from table metadata.

  For main branch data table scans, `TableScan` now parses the provided schema JSON via `TableSchema::CreateFromJson(...)` and skips `SchemaManager::Latest()`. If no schema is provided, or the scan targets a non-main branch, the existing latest-schema loading behavior is preserved.

  ### Tests

  - `cmake --build build --target paimon-core-test`
  - `ctest -R paimon-core-test --output-on-failure`

  ### API and Format

  This change adds a public API in `include/paimon/scan_context.h`:

  - `ScanContextBuilder& SetTableSchema(const std::string& table_schema)`

  No storage format or protocol changes.

  ### Documentation

  This introduces a new scan-side optimization API. Inline API documentation is added in `scan_context.h`.

  ### Generative AI tooling


include/paimon/scan_context.h
```diff
@@ -88,6 +89,10 @@ class PAIMON_EXPORT ScanContext {
         return specific_file_system_;
     }
 
+    const std::optional<std::string>& GetSpecificTableSchema() {
+        return table_schema_;
+    }
```
lxy-9602, line 94: Could you add the `const` qualifier to this function?

lxy-9602: Also, could you make `GetSpecificTableSchema` in `read_context.h` a const method as well?

src/paimon/core/table/source/table_scan.cpp
```diff
@@ -194,13 +194,20 @@ Result<std::unique_ptr<TableScan>> NewDataTableScan(const std::shared_ptr<ScanCo
...
+        PAIMON_ASSIGN_OR_RAISE(table_schema,
+                               TableSchema::CreateFromJson(specific_table_schema.value()));
+    } else {
+        SchemaManager schema_manager(tmp_options.GetFileSystem(), context->GetPath(), branch);
+        PAIMON_ASSIGN_OR_RAISE(std::optional<std::shared_ptr<TableSchema>> latest_table_schema,
+                               schema_manager.Latest());
+        if (latest_table_schema == std::nullopt) {
+            return Status::Invalid("not found latest schema");
+        }
+        table_schema = latest_table_schema.value();
```
lxy-9602, line 209: Could you add an end-to-end test in `scan_inte_test.cpp`, similar to the one in `read_inte_test.cpp`, to validate the result of scan with `specific_table_schema` configured?

lxy-9602 (dismissed):
+1

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -162,11 +163,15 @@ Status PrefetchFileBatchReaderImpl::SetReadSchema(
...
-    return RefreshReadRanges();
+    return RefreshReadRangesAfterCleanUp();
 }
 
 Status PrefetchFileBatchReaderImpl::RefreshReadRanges() {
     PAIMON_RETURN_NOT_OK(CleanUp());
+    return RefreshReadRangesAfterCleanUp();
+}
+
+Status PrefetchFileBatchReaderImpl::RefreshReadRangesAfterCleanUp() {
```
lszskye, line 174: Why split out the RefreshReadRangesAfterCleanUp() function?

gripleaf: I split out RefreshReadRangesAfterCleanUp() because SetReadSchema() needs a different ordering now:

  CleanUp();
  reader->SetReadSchema(...);
  RefreshReadRangesAfterCleanUp();

  The original RefreshReadRanges() always calls CleanUp() internally. If SetReadSchema() called it directly after stopping the prefetch thread, we would either reset schemas before cleanup, which caused the ORC crash, or call CleanUp() twice.

  So the split keeps the behavior clear:

  RefreshReadRanges()
    -> CleanUp()
    -> RefreshReadRangesAfterCleanUp()

  SetReadSchema()
    -> CleanUp()
    -> reset reader schemas
    -> RefreshReadRangesAfterCleanUp()

  This lets us reuse the range-refresh logic while guaranteeing that schema reset happens only after the background prefetch thread has fully stopped.

lszskye (approved):
+1

PR #397, opened 2026-07-02
feat(commit): align functionality with java paimon commit

lucasfang:
### Purpose

Linked issue: N/A

This PR aligns the C++ commit path with Java Paimon commit semantics, with focus on correctness and conflict handling in PK/overwrite/index scenarios.

Main changes:
- Align commit conflict detection behavior with Java, including changed-entry collection flow and conflict-check triggers.
- Add and integrate dedicated commit helpers for:
  - Manifest entry change aggregation
  - Conflict detection
  - Row-id and write-column conflict checking
  - Retry waiter backoff behavior
  - Sequence snapshot property merge logic
- Align overwrite and append behavior in index-related paths:
  - Correct commit kind selection in append/overwrite edge cases
  - Proper handling of index-file commits in overwrite/filter-and-overwrite flows
- Align row-tracking commit behavior with Java:
  - Snapshot sequence assignment
  - Row-id assignment for normal/blob/vector-related files
- Include follow-up fixes from review feedback for boundary/correctness cases.

### Tests

Added test cases in this PR:
- TEST(BucketModeTest, TestResolveBucketMode)
- TEST(CommitChangesProviderTest, TestProvideReturnsGivenEntries)
- TEST(CommitChangesProviderTest, TestProvideUsesCopiedInputs)
- TEST(CommitContextTest, TestDefaultValue)
- TEST(CommitContextTest, TestSetContent)
- TEST(CommitContextTest, TestSetOptionsOverridesAddedOptions)
- TEST(CommitMetricsTest, TestReportCommitFromStats)
- TEST(CommitStatsTest, TestCalcChangedPartitionsAndBuckets)
- TEST(CommitStatsTest, TestFailedAppendSnapshot)
- TEST(CommitStatsTest, TestSucceedAllSnapshot)
- TEST(CompactedChangelogPathResolverTest, IsCompactedChangelogPath)
- TEST(CompactedChangelogPathResolverTest, KeepInvalidCompactedPath)
- TEST(CompactedChangelogPathResolverTest, KeepNonCompactedPath)
- TEST(CompactedChangelogPathResolverTest, KeepRealCompactedChangelogPath)
- TEST(CompactedChangelogPathResolverTest, ResolveFakeCompactedChangelogPath)
- TEST(CompactedChangelogPathResolverTest, ResolveFileWithoutExtension)
- TEST(CompactedChangelogPathResolverTest, ResolveNonCompactedChangelogPath)
- TEST(CompactedChangelogPathResolverTest, ResolveWithDifferentFormats)
- TEST(CoreOptionsTest, TestDynamicPartitionOverwriteOption)
- TEST_F(CleanInteTest, DISABLED_TestDropPartitionAndExpireSnapshotWithIOException)
- TEST_F(CommitScannerTest, TestReadAllEntriesFromChangedPartitionsBuildsScanFilterPartitions)
- TEST_F(CommitScannerTest, TestReadAllEntriesFromChangedPartitionsEmptyFastExit)
- TEST_F(CommitScannerTest, TestReadAllEntriesFromPartitionsRequiresSupplier)
- TEST_F(ConflictDetectionTest, TestBucketKeepSame)
- TEST_F(ConflictDetectionTest, TestBucketKeepSameCacheEviction)
- TEST_F(ConflictDetectionTest, TestBucketKeepSameHelpers)
- TEST_F(ConflictDetectionTest, TestCheckLsmKeyRangeConflict)
- TEST_F(ConflictDetectionTest, TestCollectUncheckedBucketPartitionsMismatch)
- TEST_F(ConflictDetectionTest, TestDedicatedStorageRowIdRangeParityWithRelease14)
- TEST_F(ConflictDetectionTest, TestDeletionVectorsAllowedWithResolvedDynamicBucketMode)
- TEST_F(ConflictDetectionTest, TestDeletionVectorsNotSupportedWithBucketUnawareMode)
- TEST_F(ConflictDetectionTest, TestDeletionVectorsNotSupportedWithResolvedBucketUnawareModeFromMinusOne)
- TEST_F(ConflictDetectionTest, TestFileDeletionConflicts)
- TEST_F(ConflictDetectionTest, TestGlobalIndexRowIdExistenceConflicts)
- TEST_F(FileStoreCommitImplTest, FilterAndCommitSkipCompactBeforeFileCheck)
- TEST_F(FileStoreCommitImplTest, FilterAndCommitWithCompactedChangelogFakePath)
- TEST_F(FileStoreCommitImplTest, TestAppendDiscardDuplicateFiles)
- TEST_F(FileStoreCommitImplTest, TestCommitWithCompactDeletedIndexFiles)
- TEST_F(FileStoreCommitImplTest, TestCommitWithCompactIndexFiles)
- TEST_F(FileStoreCommitImplTest, TestCommitWithDeletedIndexFiles)
- TEST_F(FileStoreCommitImplTest, TestCommitWithGlobalIndexFilesChecksConflicts)
- TEST_F(FileStoreCommitImplTest, TestCommitWithIndexFiles)
- TEST_F(FileStoreCommitImplTest, TestFilterAndOverwriteWithCompactIndexFiles)
- TEST_F(FileStoreCommitImplTest, TestFilterCommittedRejectsDuplicateIdentifiers)
- TEST_F(FileStoreCommitImplTest, TestFixedBucketPKTableCommitAllowed)
- TEST_F(FileStoreCommitImplTest, TestOverwriteWithCompactIndexFiles)
- TEST_F(FileStoreCommitImplTest, TestSnapshotSequenceMaxPropertyMergedOnCommit)
- TEST_F(FileStoreCommitImplTest, TestTryOverwriteWithProperties)
- TEST(FileStoreCommitTest, TestAppendDvIndexShouldUseOverwriteCommitKind)
- TEST_F(ManifestEntryChangesTest, TestChangedPartitionsIncludesDvAndGlobalIndex)
- TEST_F(ManifestEntryChangesTest, TestCollectAndSummary)
- TEST_F(ManifestEntryChangesTest, TestCollectInvalidCommitMessageType)
- TEST_F(ManifestEntryChangesTest, TestHasGlobalIndexFileAdditions)
- TEST_F(RowIdColumnConflictCheckerTest, TestAllowsDisjointWriteColumns)
- TEST_F(RowIdColumnConflictCheckerTest, TestDetectsSameWriteColumns)
- TEST_F(RowIdColumnConflictCheckerTest, TestIgnoreUnknownNonSystemWriteColumn)
- TEST_F(RowIdColumnConflictCheckerTest, TestMergesOverlappedDeltaRangesAndWriteColumns)
- TEST_F(RowIdColumnConflictCheckerTest, TestScansAllOverlappedRangesAfterBinarySearch)
- TEST_F(RowIdColumnConflictCheckerTest, TestTreatsNullWriteColumnsAsFullSchemaWrite)
- TEST_F(RowIdColumnConflictCheckerTest, TestUsesFieldIdAcrossRename)
- TEST_F(RowTrackingCommitUtilsTest, TestAssignRowTracking)
- TEST_F(RowTrackingCommitUtilsTest, TestAssignRowTrackingDoesNotMutateInputEntries)
- TEST_F(RowTrackingCommitUtilsTest, TestAssignRowTrackingReassignsOnRetryWithAdvancedRowId)
- TEST_F(RowTrackingCommitUtilsTest, TestAssignRowTrackingStampsSequence)
- TEST_F(RowTrackingCommitUtilsTest, TestAssignRowTrackingStampsSequenceRangeStartingAtZero)
- TEST_F(RowTrackingCommitUtilsTest, TestAssignRowTrackingWithoutFileSource)
- TEST(OverwriteChangesProviderTest, TestProvidePropagatesScanErrors)
- TEST(OverwriteChangesProviderTest, TestProvideWithLatestSnapshotAddsDeletesAndAppendsAllChanges)
- TEST(OverwriteChangesProviderTest, TestProvideWithoutLatestSnapshotUsesChangesOnly)
- TEST_P(DataEvolutionTableTest, TestCommitConflictOnOverlappedRowIdAndWriteColumns)
- TEST(RetryWaiterTest, TestRetryWaitWithLargeRetryCountIsClamped)
- TEST(RetryWaiterTest, TestRetryWaitWithZeroBoundsReturnsQuickly)
- TEST(SpecialFieldsTest, TestIsSystemField)

### API and Format

- No new end-user API introduced.
- Commit-related internal behavior is aligned with Java semantics.
- Snapshot/Tag metadata handling is adjusted for parity in commit path.
- No new data-file format is introduced in this PR.

### Documentation

- No dedicated user-facing documentation update is required for this internal commit-path alignment change.

### Generative AI tooling


src/paimon/core/operation/commit/row_id_column_conflict_checker.cpp
```diff
@@ -0,0 +1,224 @@
...
+            continue;
+        }
+
+        if (!file.write_cols.has_value()) {
+            return true;
+        }
+
+        for (const auto& write_col : file.write_cols.value()) {
+            auto field_id_result = ResolveFieldId(file, write_col);
+            if (!field_id_result.ok()) {
```
zjw1111: `ConflictsWith` swallows `ResolveFieldId` errors as `return false` (i.e. no conflict). Java's `fieldId()` throws on an unknown non-system write column (`RowIdColumnConflictCheckerTest.testFailsOnUnknownNonSystemWriteColumn`), and the constructor path (`AddWriteFieldIds`) already propagates the same error via `PAIMON_ASSIGN_OR_RAISE`, so this is inconsistent. Please make `ConflictsWith` return `Result<bool>` and propagate the error to the caller in `conflict_detection.cpp` — otherwise a schema mismatch can silently pass conflict detection.

src/paimon/core/operation/commit/row_id_column_conflict_checker.cpp
```diff
@@ -0,0 +1,224 @@
...
+Result<std::optional<int32_t>> RowIdColumnConflictChecker::ResolveFieldId(
+    const DataFileMeta& file, const std::string& write_col) const {
+    const std::map<std::string, int32_t>* field_id_by_name = nullptr;
+    PAIMON_RETURN_NOT_OK(FieldIdByName(file.schema_id, &field_id_by_name));
+    auto it = field_id_by_name->find(write_col);
+    if (it != field_id_by_name->end()) {
+        return std::optional<int32_t>(it->second);
+    }
+
+    if (SpecialFields::IsSpecialFieldName(write_col)) {
```
zjw1111: The set used via `IsSpecialFieldName` (`_SEQUENCE_NUMBER, _VALUE_KIND, rowkind, _ROW_ID, _INDEX_SCORE`) diverges from Java `SpecialFields.isSystemField` (`_KEY_` prefix + `_SEQUENCE_NUMBER, _VALUE_KIND, _LEVEL, rowkind, _ROW_ID`). As a result `_LEVEL`/`_KEY_*` are treated as user columns (error instead of skip) and `_INDEX_SCORE` is skipped (instead of error). Please use an `isSystemField`-equivalent check here rather than the shared `IsSpecialFieldName`.

src/paimon/core/operation/commit/conflict_detection.cpp
```diff
@@ -0,0 +1,537 @@
...
+}
+
+Status ConflictDetection::CheckConflicts(
+    const Snapshot& latest_snapshot, const std::vector<ManifestEntry>& base_entries,
+    const std::vector<ManifestEntry>& delta_entries,
+    const std::vector<IndexManifestEntry>& delta_index_entries,
+    const std::optional<std::shared_ptr<RowIdColumnConflictChecker>>&
+        row_id_column_conflict_checker,
+    const Snapshot::CommitKind& commit_kind) const {
+    if (options_.DeletionVectorsEnabled() &&
```
zjw1111, line 109: This uses `GetBucket() == UNAWARE_BUCKET` (== 0), but an unaware append table defaults to `bucket = -1` (`ResolveBucketMode` maps `-1 && no PK` to `BUCKET_UNAWARE`). Java guards on `bucketMode.equals(BUCKET_UNAWARE)`. With DV enabled on a `bucket=-1` unaware table this guard is skipped and the code falls through into bucket-aware conflict logic. Please check against the resolved bucket mode instead of the raw bucket number.

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -711,28 +1019,20 @@ Result<bool> FileStoreCommitImpl::TryCommitOnce(
...
+            ? std::nullopt
+            : std::optional<std::map<std::string, std::string>>(snapshot_properties),
         next_row_id_start);
 
     Result<bool> commit_result = CommitSnapshotImpl(new_snapshot, delta_statistics);
     if (!commit_result.ok()) {
-        // commit exception, not sure about the situation and should not clean up the files.
-        PAIMON_LOG_WARN(logger_, "You need call FilterAndCommit to retry commit for exception. %s",
+        // commit exception is uncertain; retry after checking whether this commit already exists.
+        PAIMON_LOG_WARN(logger_, "Retry commit for exception. %s",
```
zjw1111, line 1087: When `CommitSnapshotImpl` returns a non-OK status, the error is dropped here and the loop retries; on timeout only the generic conflict message is thrown (and it prints the configured max-retries, not the actual retry count). Java carries the underlying exception in `RetryCommitResult` and rethrows it in the timeout message. Please preserve the last hard-error status and surface it on abort, and distinguish conflict-retry from hard errors so the root cause isn't lost.

src/paimon/core/operation/commit/row_tracking_commit_utils.cpp
```diff
@@ -0,0 +1,133 @@
...
+    PAIMON_ASSIGN_OR_RAISE(
+        int64_t next_row_id_start,
+        AssignRowTrackingMeta(first_row_id_start, snapshot_assigned, &row_id_assigned));
+    return RowTrackingAssigned{next_row_id_start, std::move(row_id_assigned)};
+}
+
+void RowTrackingCommitUtils::AssignSnapshotId(int64_t snapshot_id,
+                                              const std::vector<ManifestEntry>& delta_files,
+                                              std::vector<ManifestEntry>* snapshot_assigned) {
+    for (const auto& entry : delta_files) {
```
zjw1111, line 62: `ManifestEntry assigned_entry = entry;` is a shallow copy sharing the same `DataFileMeta`, so `AssignSequenceNumber`/`AssignFirstRowId` mutate the caller's `delta_files` in place. Java's `assign*` return new immutable objects. Current callers overwrite `delta_files` right after, so it's latent, but any other holder of the same `shared_ptr<DataFileMeta>` gets silently polluted. Please clone the `DataFileMeta` before assigning (or have `assign*` return new objects).

src/paimon/core/operation/commit/commit_scanner.cpp
```diff
@@ -0,0 +1,203 @@
...
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<FileStoreScan> scan, scan_supplier_(scan_filter));
+    PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<FileStoreScan::RawPlan> plan,
+                           scan->WithSnapshot(snapshot)->WithKind(ScanMode::ALL)->CreatePlan());
+    return plan->Files();
+}
+
+Result<std::vector<ManifestEntry>> CommitScanner::ReadAllEntriesFromPartitions(
+    const Snapshot& snapshot,
+    const std::vector<std::map<std::string, std::string>>& partitions) const {
+    auto scan_filter = std::make_shared<ScanFilter>(/*predicate=*/nullptr, partitions,
```
zjw1111, line 140: The overwrite scan passes no bucket filter. Java `OverwriteChangesProvider.newScan()` applies `bucket -> bucket >= 0` when `numBucket != POSTPONE_BUCKET` to exclude postpone-bucket files; without it, pending postpone-bucket files get DELETE entries on a normal overwrite. `ScanFilter::bucket_filter_` (`optional<int32_t>`) also can't express `bucket>=0` today. Please add a way to exclude negative buckets for the overwrite scan.

src/paimon/core/operation/commit/overwrite_changes_provider.cpp
```diff
@@ -0,0 +1,76 @@
...
+                           manifest_scan_(latest_snapshot.value()));
+    std::unordered_set<FileEntry::Identifier> existing_identifiers;
+    existing_identifiers.reserve(entries.size());
+    for (const auto& entry : entries) {
+        existing_identifiers.insert(entry.CreateIdentifier());
+        delta_files->emplace_back(FileKind::Delete(), entry.Partition(), entry.Bucket(),
+                                  entry.TotalBuckets(), entry.File());
+    }
+
+    for (const auto& change : changes_) {
```
zjw1111: This dedup (skip an ADD whose identifier already exists) diverges from Java `buildResult()`, which just `addAll(newChanges)`. On an identifier collision the DELETE(X) is kept while ADD(X) is dropped, so after `FileEntry::MergeEntries` X is net-deleted (data loss). Unlikely to trigger given unique file names, but please align with Java (drop the dedup) or skip both the delete and the add together.

src/paimon/core/operation/commit/commit_changes_provider.h
```diff
@@ -0,0 +1,58 @@
...
+                                const std::vector<ManifestEntry>& changelog_files,
+                                const std::vector<IndexManifestEntry>& index_entries);
+
+    Status Provide(const std::optional<Snapshot>& latest_snapshot,
+                   std::vector<ManifestEntry>* delta_files,
+                   std::vector<ManifestEntry>* changelog_files,
+                   std::vector<IndexManifestEntry>* index_entries) const override;
+
+ private:
+    const std::vector<ManifestEntry>& delta_files_;
```
zjw1111: These providers hold `delta_files`/`changelog_files`/`index_entries` as `const&` members while being handed out as a `shared_ptr` (same for `OverwriteChangesProvider`, which also captures `this` by raw pointer). Current construction sites keep the referenced vectors alive during the synchronous `TryCommit`, so it's safe today, but it's a UAF footgun and differs from Java's value/GC capture. Please hold by value + `std::move`.

zjw1111 (commented):
Reviewed the commit conflict-detection alignment against Java Paimon. The port is largely faithful (ManifestEntryChanges collection order, RetryWaiter backoff math, SequenceSnapshotProperties ADD-only max, row-tracking assignment all match Java). A few correctness gaps and divergences are worth fixing before merge — see inline.

Highest priority: `RowIdColumnConflictChecker::ConflictsWith` swallows field-resolution errors and can let a schema mismatch pass conflict detection.

Also, a few items not inlined:
- Test coverage: `RowIdColumnConflictChecker` has no C++ test at all (Java has 7 cases, incl. `testUsesFieldIdAcrossRename` / `testFailsOnUnknownNonSystemWriteColumn` which cover two of the issues below); `CheckRowIdExistence` / `CheckRowIdRangeConflicts` / `SequenceSnapshotProperties` also lack direct tests. Please add these.
- Style (`docs/code-style.md`): plain `int` in `row_id_column_conflict_checker.cpp` binary search (`low/high/mid/index`) and `std::unordered_set<int>` in `file_store_commit_impl.cpp` NumChangedBuckets — use fixed-width types.
- Intentional partial ports (please confirm and add a note): DV+UNAWARE entry enrichment, `pkClusteringOverride` short-circuit in CheckKeyRange, partition-expire handling in CheckDeleteInEntries, and `dropStats` on overwrite delete entries are all omitted.
- The PR description is still the unfilled template — please fill in the change summary, linked issue, and tests.

src/paimon/common/table/special_fields.h
```diff
@@ -63,6 +64,15 @@ struct SpecialFields {
         return data_field;
     }
 
+    static bool IsSystemField(const std::string& field_name) {
+        if (StringUtils::StartsWith(field_name, KEY_FIELD_PREFIX)) {
+            return true;
+        }
+        return field_name == SequenceNumber().Name() || field_name == ValueKind().Name() ||
+               field_name == RowKind().Name() || field_name == RowId().Name() ||
+               field_name == IndexScore().Name();
```
zjw1111, line 73: use IsSpecialFieldName?

src/paimon/core/catalog/commit_table_request_test.cpp
```diff
@@ -37,7 +37,7 @@ TEST(CommitTableRequestTest, TestSimple) {
         /*changelog_manifest_list_size=*/std::nullopt, /*index_manifest=*/std::nullopt,
         /*commit_user=*/"commit_user_1", /*commit_identifier=*/9223372036854775807,
         /*commit_kind=*/Snapshot::CommitKind::Append(), /*time_millis=*/1758097357597,
-        /*log_offsets=*/std::map<int32_t, int64_t>(), /*total_record_count=*/5,
+        /*total_record_count=*/5,
```
zjw1111, line 40: format it

src/paimon/core/catalog/renaming_snapshot_commit_test.cpp
```diff
@@ -46,7 +46,7 @@ TEST(RenamingSnapshotCommitTest, TestSimple) {
         /*changelog_manifest_list_size=*/std::nullopt, /*index_manifest=*/std::nullopt,
         /*commit_user=*/"commit_user_1", /*commit_identifier=*/9223372036854775807,
         /*commit_kind=*/Snapshot::CommitKind::Append(), /*time_millis=*/1758097357597,
-        /*log_offsets=*/std::map<int32_t, int64_t>(), /*total_record_count=*/5,
+        /*total_record_count=*/5,
```
zjw1111, line 49: format it

src/paimon/core/core_options.h
```diff
@@ -48,6 +48,15 @@ class Cache;
 
 class PAIMON_EXPORT CoreOptions {
  public:
+    // Specifies how to initialize the next sequence number for primary key table writers.
```
zjw1111: use `///`

zjw1111 (commented):
Cross-language review against Java Paimon found four additional correctness and boundary-condition gaps; see the inline comments.

The existing unresolved threads about overwrite ADD/DELETE deduplication and preserving the underlying commit error also still apply. I would not approve this revision until the P1 data-integrity issues are addressed.

src/paimon/core/operation/commit/row_tracking_commit_utils.cpp
```diff
@@ -0,0 +1,144 @@
...
+        ManifestEntry assigned_entry = CloneEntryWithClonedFileMeta(entry);
```
zjw1111, line 63: `ManifestEntry assigned_entry = entry` only copies the shared `DataFileMeta`, so this assignment mutates the fixed provider's original input. If the atomic commit loses a CAS race, the next attempt receives metadata already stamped with the old snapshot ID and first-row ID; because those fields are now nonzero/non-null, it skips reassignment against the advanced latest snapshot and can commit overlapping row IDs or an old sequence. Java's `entry.assignSequenceNumber` / `assignFirstRowId` return new immutable entries. Please deep-copy the metadata (or make these APIs return new objects) and add a retry test where the first attempt fails and `nextRowId` advances concurrently.

src/paimon/core/operation/commit/conflict_detection.cpp
```diff
@@ -0,0 +1,540 @@
...
+        if (!overlap_group_end) {
+            group_max_to = std::max(group_max_to, entries_with_ranges[i].first.to);
+            continue;
+        }
+
+        std::optional<std::pair<int64_t, int64_t>> expected_range;
+        for (size_t j = group_start; j < i; ++j) {
+            const ManifestEntry& entry = entries_with_ranges[j].second;
+            if (IsDedicatedStorageFile(entry.FileName())) {
+                continue;
```
zjw1111: Skipping dedicated-storage files here omits Java's second half of the range check. Java verifies that every blob/vector row-ID range is fully contained by one data-file range; with this implementation, data `[0,9]` plus blob `[5,14]` returns OK. That can commit a dangling or cross-data-file mapping. Please split data and dedicated ranges and add the equivalent containment check.

src/paimon/core/operation/commit/conflict_detection.cpp
```diff
@@ -0,0 +1,540 @@
...
+        return Status::OK();
+    }
+
+    if (row_id_check_from_snapshot_.value() > latest_snapshot.Id()) {
+        return Status::OK();
+    }
+
+    PAIMON_ASSIGN_OR_RAISE(Snapshot check_snapshot,
+                           snapshot_manager_->LoadSnapshot(row_id_check_from_snapshot_.value()));
+    if (!check_snapshot.NextRowId()) {
```
zjw1111, line 481: This should fail closed rather than silently disable history conflict detection. Java requires `nextRowId` to be present for `rowIdCheckFromSnapshot` and throws otherwise. Returning OK here lets migrated/older snapshots bypass the row-ID column conflict scan and may admit overlapping writes. Please return an invalid status that identifies the snapshot.

src/paimon/core/operation/commit/retry_waiter.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace paimon {
+
+RetryWaiter::RetryWaiter(int64_t min_retry_wait_ms, int64_t max_retry_wait_ms)
+    : min_retry_wait_ms_(std::max<int64_t>(0, min_retry_wait_ms)),
+      max_retry_wait_ms_(std::max<int64_t>(0, max_retry_wait_ms)) {}
+
+void RetryWaiter::RetryWait(int32_t retry_count) const {
+    int32_t non_negative_retry_count = std::max<int32_t>(0, retry_count);
+    double exponential = std::pow(2.0, static_cast<double>(non_negative_retry_count));
```
zjw1111, line 34: Clamp before converting to `int64_t`. With a sufficiently large configured `commit.max-retries`, `min_retry_wait_ms_ * pow(2, retry_count)` exceeds the integer range before `std::min` is applied; the conversion is not safe and can produce a non-positive delay, turning retries into a hot loop. Java applies `Math.min(..., maxRetryWait)` before the cast. A saturating calculation would preserve the configured maximum wait.

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -711,28 +1074,20 @@ Result<bool> FileStoreCommitImpl::TryCommitOnce(
...
```
zjw1111, line 1087: On a non-OK `CommitSnapshotImpl` the underlying status is dropped and the loop retries; on timeout only the generic `Commit failed after {} millis with {} retries` is thrown (and it prints the configured timeout, not the elapsed time). Java carries the underlying exception in `RetryCommitResult` and rethrows it. Please preserve the last hard-error status and surface it on abort, and distinguish conflict-retry from hard errors so the root cause isn't lost.

src/paimon/core/operation/commit/conflict_detection.cpp
```diff
@@ -0,0 +1,614 @@
...
+    std::vector<Range> data_ranges;
+    data_ranges.reserve(data_files.size());
+    for (const ManifestEntry& data_file : data_files) {
+        int64_t data_range_from = data_file.File()->first_row_id.value();
+        int64_t data_range_to = data_range_from + data_file.File()->row_count - 1;
+        data_ranges.emplace_back(data_range_from, data_range_to);
+    }
+
+    PAIMON_ASSIGN_OR_RAISE(RowRangeIndex data_file_row_range_index,
+                           RowRangeIndex::Create(data_ranges, /*merge_adjacent=*/false));
```
zjw1111, line 438: `RowRangeIndex::Create` rejects empty input (`Status::Invalid("Ranges cannot be empty")`), but master's `RowRangeIndex.create` only rejects null and accepts an empty list (empty index). Here `data_files` can be empty while `dedicated_files` is non-empty; master then reports a proper `is not covered by one data file range` conflict, whereas this path surfaces the internal `Ranges cannot be empty` error instead. Please make `RowRangeIndex::Create` accept empty ranges (return an empty index) to match master, then drop the workaround empty-guard in `CheckRowIdExistence`.

src/paimon/core/operation/metrics/commit_stats.h
```diff
@@ -0,0 +1,202 @@
...
+        for (const auto& change : changes) {
+            for (const auto& entry : change) {
+                changed_partitions.insert(entry.Partition());
+            }
+        }
+        return static_cast<int64_t>(changed_partitions.size());
+    }
+
+    static int64_t NumChangedBuckets(const std::vector<std::vector<ManifestEntry>>& changes) {
+        std::unordered_map<BinaryRow, std::set<int32_t>> changed_partition_buckets;
```
zjw1111, line 151: `std::set` is used here but the header only includes `<unordered_set>`, not `<set>`. This compiles via transitive includes today but is fragile. Please add `#include <set>`.

include/paimon/utils/row_range_index.h
```diff
@@ -31,14 +31,21 @@ class PAIMON_EXPORT RowRangeIndex {
...
+                                        bool merge_adjacent = true);
 
     /// Returns the sorted, non-overlapping ranges held by this index.
     const std::vector<Range>& Ranges() const;
 
     /// Returns true if any range in this index intersects with the interval [start, end].
     bool Intersects(int64_t start, int64_t end) const;
 
+    /// Returns true if one range in this index fully contains `range`.
+    bool Contains(const Range& range) const;
```
zjw1111, line 44: `Contains` / `ContainsExactly` and the new `merge_adjacent` parameter are new public API with no unit tests (master has `RowRangeIndexTest` covering these). Please add coverage, especially `merge_adjacent=false` and the `Contains` vs `ContainsExactly` boundaries.

zjw1111 (commented):
Re-reviewed against the latest Java master (f7ec93b3c). The dedicated row-id range check, the RowRangeIndex-based existence / global-index checks, `ReadIncrementalEntries`, and the 3-branch sequence assignment now all align with master. No blocking data-correctness issues. Remaining items are inline.

One non-inline note: `SpecialFields::IsSystemField` omits `_LEVEL` (master's `SYSTEM_FIELD_NAMES` includes it). But `_LEVEL` only lives in the KeyValue/merge-tree schema and never reaches `write_cols` on the append/data-evolution path that `RowIdColumnConflictChecker` inspects, so it is not reachable here — optional.

zjw1111 (approved):
+1. Row-id / RowRangeIndex alignment with master and the added test coverage look good; CI is green. The retry error-preservation comment in file_store_commit_impl.cpp is non-blocking (diagnostics only) — please follow up separately.

zjw1111:
Post-merge follow-up (cleanup + open item):

1. Dead flag `enable-pk-commit-in-inte-test`: this PR removed the PK-commit gate in `FileStoreCommit::Create` that checked for this option, but the flag is still passed in tests and is now a no-op. Please drop it from:
   - `src/paimon/testing/utils/test_helper.h` (`new_options["enable-pk-commit-in-inte-test"]`)
   - `test/inte/write_inte_test.cpp`, `test/inte/pk_compaction_inte_test.cpp`
   - `src/paimon/core/operation/key_value_file_store_write_test.cpp`

2. Still open (non-blocking, raised inline before merge): retry-path error preservation in `file_store_commit_impl.cpp`. On a non-OK `CommitSnapshotImpl` the underlying status is dropped and the loop retries; on timeout the message reports the configured timeout and retry count, not the root-cause error. Please preserve the last hard-error status and surface it on abort.

zjw1111:
Additional non-blocking cleanup (from internal AI review, all verified accurate — style/perf only, no correctness impact):

1. Avoidable deep copy: `GetNewFilesIncrement()` / `GetCompactIncrement()` return `const &`, but the results are received by value (copies the inner `vector<shared_ptr<...>>`). Use `const auto&` at:
   - `src/paimon/core/operation/commit/manifest_entry_changes.cpp:38, 57`
   - `src/paimon/core/operation/file_store_commit_impl.cpp:260, 268` (same pattern, not flagged by the tool)

2. `src/paimon/core/operation/commit/retry_waiter.cpp`: `RetryWait` constructs a fresh `std::random_device` + `std::mt19937` on every call. Make the generator `thread_local` to reuse it across retries.

3. `src/paimon/core/operation/commit/row_id_column_conflict_checker_test.cpp:143`: test name `TestIgnoreUnknownNonSystemWriteColumn` is misleading — the assertion expects `ConflictsWith` to return an error, not to ignore the column (and it drops the third-person `s` used by sibling tests). Rename to e.g. `TestFailsOnUnknownNonSystemWriteColumn`.

src/paimon/core/schema/schema_validation.cpp
```diff
@@ -104,14 +104,10 @@ Status SchemaValidation::ValidateTableSchema(const TableSchema& schema) {
     // TODO(yonghao.fyh): check changelog num retain
     // TODO(yonghao.fyh): support file format validate data fields
     for (const auto& field_name : field_names) {
-        if (SpecialFields::IsSpecialFieldName(field_name)) {
+        if (SpecialFields::IsSystemField(field_name)) {
             return Status::Invalid(
                 fmt::format("field name '{}' in schema cannot be special field.", field_name));
```
zjw1111, line 109: `special field` or `system field`? same with the file name, func name, msg name

src/paimon/core/io/shredding_append_data_file_writer_factory.cpp
```diff
@@ -47,6 +47,8 @@ ShreddingAppendDataFileWriterFactory::CreateWriter() const {
     if (!shredding_context_) {
         return Status::Invalid("Shared-shredding append writer requires a shredding context.");
     }
+    std::shared_ptr<LongCounter> seq_num_counter =
+        options_.DataEvolutionEnabled() ? std::make_shared<LongCounter>(0) : seq_num_counter_;
```
zjw1111, line 51: move to base class?

test/inte/clean_inte_test.cpp
```diff
@@ -405,7 +405,7 @@ TEST_F(CleanInteTest, TestDropPartitionAndExpireSnapshot) {
     ASSERT_EQ(3u, manifests[1].NumAddedFiles());
 }
 
-TEST_F(CleanInteTest, TestDropPartitionAndExpireSnapshotWithIOException) {
+TEST_F(CleanInteTest, DISABLED_TestDropPartitionAndExpireSnapshotWithIOException) {
```
zjw1111, line 408: fix the disabled case

src/paimon/core/operation/commit/commit_changes_provider_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+    ASSERT_OK_AND_ASSIGN(std::shared_ptr<CommitChanges> provided, provider->Provide(std::nullopt));
+
+    const std::vector<ManifestEntry>& provided_delta = provided->delta_files;
+    const std::vector<ManifestEntry>& provided_changelog = provided->changelog_files;
+    const std::vector<IndexManifestEntry>& provided_index = provided->index_entries;
+
+    ASSERT_EQ(delta_files.size(), provided_delta.size());
+    ASSERT_EQ(changelog_files.size(), provided_changelog.size());
+    ASSERT_EQ(index_entries.size(), provided_index.size());
+    EXPECT_EQ("delta-1", provided_delta[0].FileName());
```
zjw1111, line 102: prefer use ASSERT_* in test, and grep other files

src/paimon/core/operation/commit/commit_scanner.h
```diff
@@ -0,0 +1,115 @@
...
+class ManifestFile;
+class ManifestList;
+class MemoryPool;
+class ScanFilter;
+class SchemaManager;
+class Snapshot;
+class SnapshotManager;
+class TableSchema;
+
+// Manifest entries scanner for commit operations.
```
zjw1111, line 53: `///`

src/paimon/core/operation/commit/compacted_changelog_path_resolver.h
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <string>
+
+namespace paimon {
+
+class CompactedChangelogPathResolver {
```
zjw1111, line 23: Add class and interface comments

src/paimon/core/operation/commit/compacted_changelog_path_resolver.h
```diff
@@ -0,0 +1,30 @@
...
+ public:
+    static bool IsCompactedChangelogPath(const std::string& path);
```
zjw1111, line 25: private

src/paimon/core/operation/commit/row_tracking_commit_utils.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "paimon/common/data/blob_utils.h"
+#include "paimon/common/table/special_fields.h"
+#include "paimon/core/manifest/file_source.h"
+#include "paimon/status.h"
+
+namespace paimon {
+
+namespace {
+
+bool IsVectorStoreFile(const std::string& file_name) {
```
zjw1111, line 33: same as func in src/paimon/core/operation/commit/conflict_detection.cpp

src/paimon/core/operation/commit/conflict_detection.cpp
```diff
@@ -0,0 +1,610 @@
...
+    RangeHelper<ManifestEntry>& range_helper, const std::vector<ManifestEntry>& data_files) const {
+    std::vector<ManifestEntry> data_files_copy = data_files;
+    PAIMON_ASSIGN_OR_RAISE(std::vector<std::vector<ManifestEntry>> data_file_groups,
+                           range_helper.MergeOverlappingRanges(std::move(data_files_copy)));
+    for (const std::vector<ManifestEntry>& data_file_group : data_file_groups) {
+        PAIMON_ASSIGN_OR_RAISE(bool all_data_ranges_same,
+                               range_helper.AreAllRangesSame(data_file_group));
+        if (!all_data_ranges_same) {
+            return Status::Invalid(
+                "For Data Evolution table, multiple MERGE INTO/COMPACT operations have "
```
zjw1111, line 414: there is no MERGE INTO operation in paimon-cpp?

src/paimon/core/core_options.cpp
```diff
@@ -652,6 +679,19 @@ struct CoreOptions::Impl {
             Options::SEQUENCE_FIELD, Options::FIELDS_SEPARATOR, &sequence_field));
         // Parse sequence.field.sort-order - order of sequence field, default "ascending"
         PAIMON_RETURN_NOT_OK(parser.ParseSortOrder(&sequence_field_sort_order));
+        // Parse write-sequence-number-init-mode - sequence init mode for write path
```
zjw1111, line 682: Option name, write- or write.

src/paimon/core/schema/schema_validation.cpp
lucasfang: there is no IsSpecialFieldName function any more

src/paimon/core/operation/commit/conflict_detection.cpp
lucasfang: Align error msg with Java Paimon

PR #398, opened 2026-07-02
fix: release global index writer before reader to avoid mem issue when write index failed

lxy-9602:
### Purpose

No Linked issue.

Keep `batch_reader` alive until `global_index_writer` is destroyed because some index writers may retain Arrow arrays imported from reader-produced batches.


zjw1111 (approved):
LGTM.

PR #402, opened 2026-07-07
fix(blob): allow blob files with different schema ids in a bunch

duanyyyyyyy:
### Purpose

`DataEvolutionSplitRead::BlobBunch::Add` rejected a bunch whose blob files
carry different schema ids ("All files in a blob bunch should have the same
schema id.").
Linked issue: close #xxx

After schema evolution that adds/drops OTHER columns, a single blob field's
contiguous files can legitimately have different schema ids — e.g. an old
appended file (schema N) followed by a compacted file (schema M) covering the
next row range. `MergeRangesAndSort` groups them into one bunch purely by row
id, so the check fires and a normal read fails.

Paimon Java `SpecialFieldBunch.add` guards this exact check with
`if (!isBlobFile(file))` — blob files are exempt because a blob column's on-disk
layout is schema-independent, and the bunch is read with the first file's schema
regardless. That guard was dropped when porting to C++.

## What this PR does
- Restore the `if (!IsBlobFile(file))` guard so blob files may span schema ids;
  the `write_cols` (field identity) check is unchanged.
- Add `TestAddBlobFilesWithDifferentSchemaId`.

### DIFF
--- a/src/paimon/core/operation/data_evolution_split_read.cpp
+++ b/src/paimon/core/operation/data_evolution_split_read.cpp
@@ -94,9 +94,7 @@ Status DataEvolutionSplitRead::BlobBunch::Add(const std::shared_ptr<DataFileMeta
             }
         }
         if (!files_.empty()) {
-            if (file->schema_id != files_[0]->schema_id) {
-                return Status::Invalid("All files in a blob bunch should have the same schema id.");
-            }
+            // Paimon Java does not check the schema id for blob files in a bunch.
             if (file->write_cols != files_[0]->write_cols) {
                 return Status::Invalid(
                     "All files in a blob bunch should have the same write columns.");
--- a/src/paimon/core/operation/data_evolution_split_read_test.cpp
+++ b/src/paimon/core/operation/data_evolution_split_read_test.cpp
@@ -83,13 +83,13 @@ class DataEvolutionSplitReadTest : public ::testing::Test {
 
     std::shared_ptr<DataFileMeta> CreateBlobFile(
         const std::string& file_name, int64_t first_row_id, int64_t row_count,
-        int64_t max_sequence_number,
-        const std::optional<std::vector<std::string>>& write_cols) const {
+        int64_t max_sequence_number, const std::optional<std::vector<std::string>>& write_cols,
+        int64_t schema_id = 0) const {
         return DataFileMeta::ForAppend(file_name + ".blob", /*file_size=*/row_count,
                                        /*row_count=*/row_count,
                                        /*row_stats=*/SimpleStats::EmptyStats(),
-                                       /*min_sequence_number=*/0, max_sequence_number,
-                                       /*schema_id=*/0, FileSource::Append(),
+                                       /*min_sequence_number=*/0, max_sequence_number, schema_id,
+                                       FileSource::Append(),
                                        /*value_stats_cols=*/std::nullopt,
                                        /*external_path=*/std::nullopt, first_row_id, write_cols)
             .value();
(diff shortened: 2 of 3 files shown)


src/paimon/core/operation/data_evolution_split_read_test.cpp
```diff
@@ -247,6 +247,28 @@ TEST_F(DataEvolutionSplitReadTest, TestAddBlobFileWithDifferentWriteCols) {
...
+                       /*max_sequence_number=*/1,
+                       /*write_cols=*/std::optional<std::vector<std::string>>({"blob_col"}),
+                       /*schema_id=*/1);
+    auto blob_bunch = std::make_shared<DataEvolutionSplitRead::BlobBunch>(
+        INT64_MAX, /*has_row_ids_selection=*/false);
+    ASSERT_OK(blob_bunch->Add(blob_entry));
+    ASSERT_OK(blob_bunch->Add(blob_tail));
+    ASSERT_EQ(blob_bunch->Files().size(), 2);
+}
+
```
lxy-9602, line 271: Could you add an integration test in `blob_table_inte_test.cpp` that uses `WriteNextSchema` to simulate an ALTER TABLE scenario?

src/paimon/core/operation/data_evolution_split_read.cpp
```diff
@@ -94,8 +94,18 @@ Status DataEvolutionSplitRead::BlobBunch::Add(const std::shared_ptr<DataFileMeta
...
-                return Status::Invalid("All files in a blob bunch should have the same schema id.");
+            // Mirror Paimon Java (SpecialFieldBunch.add): the schema id must agree only for
+            // non-blob (vector-store) files. Blob files are allowed to span schema ids -- a
+            // blob column's on-disk layout is schema-independent, so after schema evolution
+            // (which adds/drops OTHER columns) the same blob field can have contiguous files
+            // written under different schema ids that MergeRangesAndSort groups into one
+            // bunch. The Java guard `if (!isBlobFile(file))` around this check was dropped
+            // when porting, which turned that legitimate case into a spurious error.
+            if (!BlobUtils::IsBlobFile(file->file_name)) {
+                if (file->schema_id != files_[0]->schema_id) {
```
lxy-9602: `BlobBunch::Add` already rejects non-blob files at the beginning, so the new `if (!BlobUtils::IsBlobFile(file->file_name))` guard around the schema-id check is unreachable. Functionally this is fine, but it may confuse readers. Could we remove the schema-id check entirely for `BlobBunch` and replace the comment with something like: “BlobBunch only contains blob files. Unlike vector-store files, blob files may span schema ids as long as they belong to the same blob field.”

test/inte/blob_table_inte_test.cpp
```diff
@@ -840,6 +840,85 @@ TEST_P(BlobTableInteTest, TestMultipleAppendsDifferentFirstRowIds) {
...
+    // Without the fix the read fails with
+    // "All files in a blob bunch should have the same schema id."; with it both rows read.
+    auto expected_array = std::dynamic_pointer_cast<arrow::StructArray>(
+        arrow::ipc::internal::json::ArrayFromJSON(arrow::struct_(fields_), R"([
+        [1, "a", "b"],
+        [2, "c", "d"]
+    ])")
+            .ValueOrDie());
+    ASSERT_OK(ScanAndRead(table_path, schema->field_names(), expected_array));
+}
```
lxy-9602, line 1209: I tested this integration test locally against the unmodified/pre-fix code, and it still passes. So I don’t think this test actually hits the regression scenario.

The test creates two f1 blob files with adjacent row ranges: [0, 0] under schema 0 and [1, 1] under schema 1. However, RangeHelper::MergeOverlappingRanges only merges overlapping ranges, not adjacent ranges. As a result, these two files are split into different `need_merge_files` groups, and each group creates its own `BlobBunch` with only one blob file. The old schema-id check is therefore never exercised.

To cover the real scenario, the integration test needs to ensure that two blob files with different schema ids are added to the same BlobBunch, for example all with first row id 0.

test/inte/blob_table_inte_test.cpp
```diff
@@ -840,6 +840,85 @@ TEST_P(BlobTableInteTest, TestMultipleAppendsDifferentFirstRowIds) {
...
+
 TEST_P(BlobTableInteTest, TestMoreDataWithDataEvolution) {
```
lxy-9602, line 1211: Could you check `TestOnlySomeColumns`? If only one blob field is written, I would expect this to fail. Does this test actually pass locally? Also, the rows for backfilling the added columns need to match the row count of the previous files. For example, if `"f0" and "f2"` write two rows in the first round, then the second round should also write two rows.

lxy-9602:
Thanks for the contribution! Since we’re close to the release, I’ve put this fix into #460, so I’ll close this PR for now. Feel free to reach out anytime if anything comes up.

PR #403, opened 2026-07-08
feat: add read-optimized system table

gripleaf:
### Purpose

  Add support for the read-optimized `$ro` system table. For primary-key tables, `$ro` reads only the highest LSM level produced by full compaction and skips merge-on-read across levels. Append-only tables keep the same read behavior as the base table, including streaming reads.

  This also refactors read-optimized scan configuration into `DataTableBatchScan`, uses context builders for `$ro` scan/read contexts, and documents freshness and value-stats filtering behavior.

  ### Tests

E2E test:
  - `SystemTableTest.TestReadOptimizedSystemTableRegistration`
  - `SystemTableTest.TestReadOptimizedSystemTablePathParsing`
  - `TableScanTest.TestReadOptimizedPrimaryKeyStreamingScanUnsupported`
  - `SystemTableReadInteTest.TestReadOptimizedSystemTable`
  - `SystemTableReadInteTest.TestReadOptimizedAppendOnlySystemTableWithStreamingScan`
  - `SystemTableReadInteTest.TestReadOptimizedSystemTableWithBranch`
  - `SystemTableReadInteTest.TestReadOptimizedSystemTableWithFirstRowMergeEngine`

  ### API and Format

  No public API, storage format, or protocol changes.

  ### Documentation

  Added user guide documentation for system tables and the read-optimized $ro system table.

  ### Generative AI tooling


lxy-9602:
@suxiaogang223 Would you mind helping take a look when you have time, since you’re most familiar with system table. Thanks a lot!

src/paimon/core/table/source/table_scan.cpp
```diff
@@ -225,9 +226,15 @@ Result<std::unique_ptr<TableScan>> NewDataTableScan(const std::shared_ptr<ScanCo
...
+    const auto read_optimized_iter = context->GetOptions().find(kReadOptimizedScanOption);
+    const bool read_optimized =
+        read_optimized_iter != context->GetOptions().end() && read_optimized_iter->second == "true";
     // merge options
     auto options = table_schema->Options();
     for (const auto& [key, value] : context->GetOptions()) {
+        if (key == kReadOptimizedScanOption) {
+            continue;
+        }
         options[key] = value;
```
lxy-9602, line 236: It seems we may not need to assume users will set internal options like `kReadOptimizedScanOption` here, so the continue branch might be avoidable. Also, would it make sense to use `OptionsUtils::GetValueFromMap` for reading the option value, rather than directly using raw strings like `true`?

src/paimon/core/table/system/read_optimized_system_table.cpp
```diff
@@ -0,0 +1,123 @@
...
+        .WithFileSystem(context->GetSpecificFileSystem())
+        .WithFileSystemSchemeToIdentifierMap(context->GetFileSystemSchemeToIdentifierMap())
+        .SetPrefetchCacheMode(context->GetPrefetchCacheMode())
+        .WithCacheConfig(context->GetCacheConfig())
+        .WithCache(context->GetCache());
+    if (!context->GetReadFieldIds().empty()) {
+        builder.SetReadFieldIds(context->GetReadFieldIds());
+    } else {
+        builder.SetReadFieldNames(context->GetReadFieldNames());
+    }
```
lxy-9602, line 120: I understand that `ReadFieldIds` has higher priority than `ReadFieldNames`, but I’d prefer to keep that assumption localized within `InternalReadContext`. My suggestion would be for `ReadContextBuilder` to copy each field independently, and let `InternalReadContext` decide which one takes precedence.

test/inte/read_inte_test.cpp
```diff
@@ -291,6 +292,17 @@ Result<SystemTableReadResult> ReadSystemTable(const std::string& system_table_pa
     return SystemTableReadResult(std::move(batch_reader), result);
 }
 
+Status WriteAndFullCompact(TestHelper* helper, std::unique_ptr<RecordBatch>&& batch,
+                           int64_t commit_identifier) {
+    PAIMON_RETURN_NOT_OK(helper->write_->Write(std::move(batch)));
+    PAIMON_RETURN_NOT_OK(
```
lxy-9602, line 308: Could you place the output parameter at the end of the parameter list?

test/inte/read_inte_test.cpp
```diff
@@ -783,6 +795,198 @@ TEST(SystemTableReadInteTest, TestReadMetadataSystemTables) {
...
+    std::shared_ptr<arrow::DataType> expected_type =
+        arrow::struct_({arrow::field("_VALUE_KIND", arrow::int8()),
+                        arrow::field("k", arrow::int32()), arrow::field("v", arrow::int32())});
+    std::shared_ptr<arrow::ChunkedArray> expected;
+    ASSERT_TRUE(arrow::ipc::internal::json::ChunkedArrayFromJSON(
+                    expected_type, {R"([[0, 1, 10], [0, 2, 20]])"}, &expected)
+                    .ok());
+    ASSERT_TRUE(result.array->Equals(expected)) << result.array->ToString();
+}
+
```
lxy-9602, line 1149: Could you add a test case to verify that system table reads support partial-column projection and that predicate pushdown is effective? For the predicate setup, you can refer to `NestedColumnPruningInteTest.NestedStructMapSelectedKeysWithPredicate`. If the Parquet/ORC row group or stripe boundaries are configured properly, the test should be able to observe and validate that predicate pushdown is actually taking effect.

test/inte/read_inte_test.cpp
```diff
@@ -783,6 +795,198 @@ TEST(SystemTableReadInteTest, TestReadMetadataSystemTables) {
...
+    ASSERT_OK_AND_ASSIGN(SystemTableReadResult result,
+                         ReadSystemTable(table_path + "$branch_rt$ro", options));
+    std::shared_ptr<arrow::DataType> expected_type =
+        arrow::struct_({arrow::field("_VALUE_KIND", arrow::int8()),
+                        arrow::field("k", arrow::int32()), arrow::field("v", arrow::int32())});
+    std::shared_ptr<arrow::ChunkedArray> expected;
+    ASSERT_TRUE(arrow::ipc::internal::json::ChunkedArrayFromJSON(
+                    expected_type, {R"([[0, 1, 10], [0, 2, 20]])"}, &expected)
+                    .ok());
+    ASSERT_TRUE(result.array->Equals(expected)) << result.array->ToString();
```
lxy-9602, line 1099: We could also take this opportunity to verify the result on the main branch in this test.

include/paimon/read_context.h
```diff
@@ -204,8 +204,8 @@ class PAIMON_EXPORT ReadContextBuilder {
     /// @param read_field_ids Vector of field ids to read from the table.
     /// @return Reference to this builder for method chaining.
     /// @note Currently supports top-level field selection.
-    /// @note SetReadFieldIds() and SetReadFieldNames() are mutually exclusive.
-    ///       Calling both will ignore the read schema set by SetReadFieldNames().
+    /// @note If both SetReadFieldIds() and SetReadFieldNames() are set, both values are
+    ///       preserved in ReadContext. InternalReadContext decides precedence.
     ReadContextBuilder& SetReadFieldIds(const std::vector<int32_t>& read_field_ids);
```
lxy-9602, line 210: Please keep this as-is — it is meant to notify users about the internal behavior.

test/inte/read_inte_test.cpp
```diff
@@ -895,6 +917,74 @@ TEST(SystemTableReadInteTest, TestReadOptimizedAppendOnlySystemTableWithStreamin
...
+                                                  {Options::MANIFEST_FORMAT, "avro"},
+                                                  {Options::BUCKET, "1"},
+                                                  {Options::BUCKET_KEY, "k"}};
+    auto dir = UniqueTestDirectory::Create();
+    ASSERT_TRUE(dir);
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<TestHelper> helper,
+                         TestHelper::Create(dir->Str(), schema,
+                                            /*partition_keys=*/{},
+                                            /*primary_keys=*/{}, options,
+                                            /*is_streaming_mode=*/true));
```
lxy-9602, line 944: What I’m really trying to test here is predicate behavior on PK tables, because RO tables additionally enable `EnableValueFilter`, which could affect the result.

test/inte/read_inte_test.cpp
```diff
@@ -783,6 +817,274 @@ TEST(SystemTableReadInteTest, TestReadMetadataSystemTables) {
...
+
+    ASSERT_OK_AND_ASSIGN(
+        std::unique_ptr<RecordBatch> batch,
+        TestHelper::MakeRecordBatch(arrow::struct_(fields), R"([[1, 10], [2, 20]])",
+                                    /*partition_map=*/{}, /*bucket=*/0, {}));
+    ASSERT_OK(helper->WriteAndCommit(std::move(batch), /*commit_identifier=*/0,
+                                     /*expected_commit_messages=*/std::nullopt));
+
+    ASSERT_OK_AND_ASSIGN(SystemTableReadResult result,
+                         ReadSystemTable(table_path + "$ro", options, /*streaming_mode=*/true));
```
lszskye, line 909: please call `ReadSystemTable` multiple times if you want to test `StreamingScan`

lxy-9602: @lszskye I think there may be a small misunderstanding here. `ReadSystemTable` performs a scan first and then reads. What we usually mean by streaming read is doing a single scan and then reading in a while loop from that same scan, rather than calling `ReadSystemTable` multiple times. For reference, please see `TestWithPKWithDvStreamFromSnapshot4` in `scan_and_read_inte_test.cpp`.

src/paimon/core/table/system/read_optimized_system_table.cpp
```diff
@@ -0,0 +1,120 @@
...
+        .SetRowToBatchThreadNumber(context->GetRowToBatchThreadNumber())
+        .WithMemoryPool(context->GetMemoryPool())
+        .WithExecutor(context->GetExecutor())
+        .WithFileSystem(context->GetSpecificFileSystem())
+        .WithFileSystemSchemeToIdentifierMap(context->GetFileSystemSchemeToIdentifierMap())
+        .SetPrefetchCacheMode(context->GetPrefetchCacheMode())
+        .WithCacheConfig(context->GetCacheConfig())
+        .WithCache(context->GetCache())
+        .SetReadFieldNames(context->GetReadFieldNames())
+        .SetReadFieldIds(context->GetReadFieldIds());
```
suxiaogang223, line 113: Could we preserve the projected Arrow schema from `ReadContext::SetReadSchema()` here? The wrapper currently copies field names and IDs only. For nested-column pruning those vectors are normally empty, so `$ro` silently falls back to the full table schema and may return a different shape from the requested projection. This is separate from the existing top-level projection test; could we transfer or re-export the read schema and add a nested projection test for `$ro`?

docs/source/user_guide/system_tables.rst
```diff
@@ -0,0 +1,60 @@
...
+The read-optimized system table is addressed by the ``$ro`` suffix:
+
+.. code-block:: text
+
+   /warehouse/db.db/orders$ro
+
+For primary-key tables, ``$ro`` only plans data files from the highest LSM
+level, which is the level produced by full compaction. This avoids merging data
+from multiple LSM levels during query planning and reading. The tradeoff is that
+the result may lag behind the latest committed data until a full compaction
```
suxiaogang223, line 38: Could we document that `$ro` does not necessarily represent one coherent older snapshot? Different buckets may complete full compaction at different times, so the top-level files selected from the latest snapshot can contain bucket states produced at different snapshots. The current “may lag” wording does not make this consistency caveat clear.

suxiaogang223 (changes requested):
I found two additional issues that are not covered by the existing review threads.

gripleaf:
@lxy-9602 @suxiaogang223 I have fixed the issues according to the comments. Please take some time to review again when you get a chance.

lxy-9602 (approved):
+1

PR #406, opened 2026-07-09
feat(variant): support variant data type

### LINKED ISSUE #91: [Feature] Introduce native support for variant data type
### Search before asking

### Motivation

Variant was born from Apache Spark, and the encoding rules have been incorporated into Parquet https://github.com/apache/parquet-format/blob/master/VariantEncoding.md. Paimon has already supported `variant` data type in https://github.com/apache/paimon/pull/4729.

### Solution

Introduce native support for `variant` data type.

### Anything else?

_No response_

SteNicholas:
### Purpose

Linked issue: close #91

Introduce native support for the `variant` data type for storing and efficiently querying semi-structured data (e.g. JSON), following the parquet-format [VariantEncoding.md](https://github.com/apache/parquet-format/blob/master/VariantEncoding.md) / [VariantShredding.md](https://github.com/apache/parquet-format/blob/master/VariantShredding.md) specifications:

- Variant binary codec, JSON parsing/rendering, and the public `paimon::Variant` API with path access (e.g. `$.a[0]`) to scalar targets and, via `VariantGetArrow`, nested `STRUCT` / `LIST` / `MAP` / `VARIANT` targets. Floating-point values render exactly as Java's `Double.toString` / `Float.toString` in both JSON output and STRING casts.
- `VARIANT` in the type system: parsed from / serialized to `"VARIANT"`, represented in Arrow as `struct<value: binary, metadata: binary>` marked with `paimon.extension.type=paimon.type.variant`; field-id assignment and schema validation treat variant structs as leaves.
- Parquet read/write with an unannotated group holding `REQUIRED BINARY value (field id 0)` / `metadata (field id 1)`; ORC/Avro report `NotImplemented` for `VARIANT` columns.
- Variant shredding on a generic shredding write-plan framework shared with MAP shared-shredding: configured via `variant.shreddingSchema` (top-level columns, as in Java), or inferred per file from sampled rows when `variant.inferShreddingSchema` is enabled (tuned by the `variant.shredding.*` options, which are validated when the table options are parsed). Inference and write conversion also cover VARIANT columns nested inside ROW types (struct-only recursion by field-index path, as in Java), and one `maxSchemaWidth` budget is shared across all variant columns of a schema.
- Read-side generic shredding read plans: shredded files are detected structurally and reassembled back into variants transparently, including variants nested inside ROW columns; a variant column can also be read as a variant-access projection built with `VariantAccessBuilder`, extracting specific paths while pruning the scan to `metadata` and the requested `typed_value` sub-columns.

### Tests

- UT: `generic_variant_test.cpp` (golden-byte compatibility, malformed/corrupted input), `variant_test.cpp` (public API), `variant_get_test.cpp` (path access, scalar and nested targets), `variant_type_utils_test.cpp`, `variant_json_utils_test.cpp` (Java `Double.toString` / `Float.toString` parity including the special values), `variant_shredding_test.cpp` (shred/reassemble round trips), `infer_variant_shredding_schema_test.cpp` (including the shared width budget), `variant_shredding_write_plan_factory_test.cpp` (including nested-ROW variants and null-ancestor rows), `infer_shredding_file_writer_test.cpp`, plus the `variant.*` option cases in `core_options_test.cpp` and VARIANT cases in `data_type_test.cpp`, `data_type_json_parser_test.cpp`, `arrow_schema_validator_test.cpp`, `table_schema_test.cpp`, `orc_format_writer_test.cpp` and `avro_schema_converter_test.cpp`.
- Format IT: `variant_parquet_test.cpp` (physical layout, unshredded/shredded round trips, variant-access reads with `typed_value` pruning).
- IT: `test/inte/variant_table_inte_test.cpp` (append and primary-key tables with deeply nested mixed-type documents, variant-access projection read, ORC rejection, and IO-exception injection on both the write and the scan+read paths).

### API and Format

- New public API: `include/paimon/data/variant.h` (`paimon::Variant`, `VariantCastArgs`, `VariantAccessBuilder`), `FieldType::VARIANT` in `include/paimon/defs.h`, and the new `variant.*` table options.
- The storage format follows the parquet-format VariantEncoding.md / VariantShredding.md specifications; the parquet physical layout is byte-compatible with Java Paimon. No changes to existing formats or protocols.

### Documentation

`docs/source/user_guide/data_types.rst` documents the VARIANT type, its Arrow representation, shredding configuration and variant-access reads.

### Generative AI tooling


src/paimon/common/data/shredding/shredding_write_plan_factory.h
```diff
@@ -0,0 +1,69 @@
...
+    virtual Result<std::shared_ptr<ShreddingBatchConverter>> CreateConverter(
+        const std::string& file_format_identifier,
+        const std::vector<std::shared_ptr<arrow::Array>>& sample_batches) const = 0;
+
+    /// The per-file metadata finalizer persisted into the file footer, or nullptr when the
+    /// physical schema is self-describing (as it is for VARIANT shredding).
+    virtual MetadataFinalizer CreateMetadataFinalizer(
+        const std::shared_ptr<ShreddingBatchConverter>& converter) const {
+        return nullptr;
+    }
```
lxy-9602: Could we avoid adding a default implementation here if possible, and keep this function as const = 0 pure virtual instead?

src/paimon/common/data/variant/infer_variant_shredding_schema.cpp
```diff
@@ -0,0 +1,361 @@
...
+        case VariantValueType::kObject: {
+            if (max_depth <= 0) {
+                return SimpleSchema::Variant();
+            }
+            PAIMON_ASSIGN_OR_RAISE(int32_t size, variant.ObjectSize());
+            auto result = std::make_shared<SimpleSchema>();
+            result->is_object = true;
+            result->fields.reserve(size);
+            for (int32_t i = 0; i < size; ++i) {
+                PAIMON_ASSIGN_OR_RAISE(auto field, variant.GetFieldAtIndex(i));
```
lxy-9602: Could we avoid `PAIMON_ASSIGN_OR_RAISE(auto, ...)` in production code if possible, and use a concrete type here instead?

src/paimon/common/data/shredding/shredding_file_reader.h
```diff
@@ -0,0 +1,65 @@
...
+#include "paimon/common/data/shredding/shredding_read_plan.h"
+#include "paimon/memory/memory_pool.h"
+#include "paimon/reader/file_batch_reader.h"
+
+namespace paimon {
+
+/// A file batch reader wrapper that pushes per-column physical (shredded, possibly pruned)
+/// subtrees down to the inner reader and assembles the physical batches back into the logical
+/// columns as planned by `ShreddingColumnReadPlan`s.
+class ShreddingFileReader : public FileBatchReader {
```
lszskye, line 34: Please add `ShreddingFileReader` in read chain in class `RawFileSplitRead`, class `MergeFileSplitRead` and class `DataEvolutionSplitRead`

Image

src/paimon/common/types/data_type.cpp
```diff
@@ -116,9 +122,16 @@ std::string DataType::DataTypeToString(const std::shared_ptr<arrow::DataType>& t
...
+        case arrow::Type::type::STRUCT: {
+            if (VariantTypeUtils::IsVariantMetadata(metadata_)) {
+                return "VARIANT";
+            }
+            [[fallthrough]];
+        }
         case arrow::Type::type::LARGE_BINARY: {
             // TODO(xinyu): change binary to large binary?
-            if (BlobUtils::IsBlobMetadata(metadata_)) {
+            if (type->id() == arrow::Type::type::LARGE_BINARY &&
```
lszskye: Already in case arrow::Type::type::LARGE_BINARY, why need `type->id() == arrow::Type::type::LARGE_BINARY `

src/paimon/common/utils/field_type_utils.h
```diff
@@ -46,6 +47,18 @@ class FieldTypeUtils {
...
 
+    /// Converts an arrow field to a `FieldType`, disambiguating metadata-marked extension types
+    /// (a VARIANT field is physically a STRUCT with the variant metadata marker).
+    static Result<FieldType> ConvertToFieldType(const std::shared_ptr<arrow::Field>& field) {
+        if (field->type()->id() == arrow::Type::type::STRUCT && field->HasMetadata()) {
+            auto extension_name = field->metadata()->Get(VariantDefs::kExtensionTypeKey);
+            if (extension_name.ok() && *extension_name == VariantDefs::kExtensionTypeValue) {
+                return FieldType::VARIANT;
+            }
+        }
```
lszskye, line 55: may use VariantTypeUtils::IsVariantField

src/paimon/common/data/variant/infer_variant_shredding_schema_test.cpp
```diff
@@ -0,0 +1,195 @@
...
+TEST_F(InferVariantShreddingSchemaTest, MaxSchemaWidthLimit) {
+    InferVariantShreddingSchema narrow_infer{/*max_schema_width=*/3, /*max_schema_depth=*/50,
+                                             /*min_field_cardinality_ratio=*/0.1};
+    auto samples = Samples({R"({"a": 1, "b": 2, "c": 3, "d": 4, "e": 5})"});
+    ASSERT_OK_AND_ASSIGN(std::shared_ptr<arrow::DataType> inferred,
+                         narrow_infer.InferColumnShreddingType(samples));
+    if (inferred != nullptr) {
+        ASSERT_EQ(inferred->id(), arrow::Type::STRUCT);
+        ASSERT_LT(inferred->num_fields(), 5);
+    }
```
lxy-9602: This test seems a bit ambiguous at the moment, since multiple scenarios could make it pass. Could we tighten it up with more specific checks?

src/paimon/common/data/variant/variant.cpp
```diff
@@ -0,0 +1,187 @@
...
+VariantAccessBuilder::VariantAccessBuilder() : impl_(std::make_unique<Impl>()) {}
+VariantAccessBuilder::~VariantAccessBuilder() = default;
+
+Status VariantAccessBuilder::AddField(struct ArrowSchema* target_type, const std::string& path,
+                                      bool fail_on_error, const std::string& zone_id) {
+    PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Field> target,
+                                      arrow::ImportField(target_type));
+    // Validate the path eagerly so mistakes fail at build time, not at read time.
+    PAIMON_RETURN_NOT_OK(VariantPathSegment::Parse(path).status());
+    // Keep the target field's own metadata (e.g. the variant extension marker of a
```
lxy-9602, line 152: `PAIMON_RETURN_NOT_OK(VariantPathSegment::Parse(path));`

src/paimon/common/data/variant/variant_access_utils.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include "fmt/format.h"
+#include "paimon/common/data/variant/variant_defs.h"
+#include "paimon/common/types/data_field.h"
+
+namespace paimon {
+
+namespace {
+
+Result<std::string> GetDescription(const std::shared_ptr<arrow::Field>& field) {
+    if (field->metadata() == nullptr) {
```
lxy-9602, line 32: It seems that this function may not need to return a `Result`.

src/paimon/core/core_options.cpp
```diff
@@ -1191,6 +1191,42 @@ Result<bool> CoreOptions::FieldCollectAggDistinct(const std::string& field_name)
     return distinct;
 }
 
+std::optional<std::string> CoreOptions::GetVariantShreddingSchema() const {
```
lszskye, line 1328: Please add ut for CoreOptions

src/paimon/common/data/variant/variant_reassembler.cpp
```diff
@@ -0,0 +1,260 @@
...
+        case VariantSchema::ScalarKind::kDate:
+            return builder->AppendDate(
+                static_cast<const arrow::Date32Array&>(typed_array).Value(row));
+        case VariantSchema::ScalarKind::kTimestampLtz:
+            return builder->AppendTimestamp(
+                static_cast<const arrow::TimestampArray&>(typed_array).Value(row));
+        case VariantSchema::ScalarKind::kTimestampNtz:
+            return builder->AppendTimestampNtz(
+                static_cast<const arrow::TimestampArray&>(typed_array).Value(row));
+        default:
```
lxy-9602, line 109: I’m not entirely sure whether you’ll validate the precision of the Arrow array, since the value in an Arrow array represents the numeric value under the corresponding precision. For example, for 1s, the value would be 1000 if the precision is ms, and 10^9 if the precision is ns.

src/paimon/common/data/variant/variant_reassembler.cpp
```diff
@@ -0,0 +1,260 @@
...
+}
+
+Result<std::shared_ptr<arrow::Array>> VariantReassembler::AssembleVariantArray(
+    const std::shared_ptr<arrow::StructArray>& shredded,
+    const std::shared_ptr<VariantSchema>& schema, arrow::MemoryPool* pool) {
+    if (schema->top_level_metadata_idx < 0) {
+        return VariantBinaryUtil::MalformedVariant();
+    }
+    std::shared_ptr<MemoryPool> paimon_pool = GetDefaultPool();
+    auto output_type = VariantTypeUtils::UnshreddedStructType();
```
lxy-9602, line 226: `GetDefaultPool()` may cause allocations outside the managed pool. It’s recommended to keep allocations within the passed-in pool, e.g. by passing two pools if necessary.

src/paimon/common/data/variant/variant_reassembler.cpp
```diff
@@ -0,0 +1,260 @@
...
+        std::string_view metadata = cursor.GetBinary(schema->top_level_metadata_idx);
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(struct_builder->Append());
+        if (unshredded) {
+            // Rebuilding is unnecessary for unshredded variants.
+            if (cursor.IsNullAt(schema->variant_idx)) {
+                return VariantBinaryUtil::MalformedVariant();
+            }
+            PAIMON_RETURN_NOT_OK_FROM_ARROW(
+                value_builder->Append(cursor.GetBinary(schema->variant_idx)));
+            PAIMON_RETURN_NOT_OK_FROM_ARROW(metadata_builder->Append(metadata));
```
lxy-9602, line 256: I’d like to confirm whether unshredded really needs to copy the data. My understanding is that the physical storage is just metadata and value, so it might be possible to return them directly, with at most a row-by-row check for `MalformedVariant`. If this could improve read performance, could you add a TODO to revisit it later? Given the complexity of the current PR, I don’t think it needs to be implemented in this PR.

src/paimon/common/data/variant/variant_shredding_write_plan.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "paimon/common/types/data_type_json_parser.h"
+#include "rapidjson/document.h"
+
+namespace paimon {
+
+Result<std::shared_ptr<VariantShreddingWritePlan>> VariantShreddingWritePlan::Create(
+    const std::shared_ptr<arrow::Schema>& logical_schema,
+    const std::map<std::string, std::shared_ptr<arrow::DataType>>& column_shredding_types) {
+    auto plan = std::shared_ptr<VariantShreddingWritePlan>(new VariantShreddingWritePlan());
+    plan->logical_schema_ = logical_schema;
```
lxy-9602: Let’s avoid using a default constructor followed by assigning values in `Create` if possible. Could we instead have Create call a constructor that takes the required parameters?

src/paimon/common/data/variant/variant_binary_util.h
```diff
@@ -0,0 +1,198 @@
...
+    static Status UnexpectedType(VariantValueType type);
+
+    /// Checks the validity of an index `pos` in a buffer of `length` bytes. Returns
+    /// `MALFORMED_VARIANT` if it is out of bound.
+    static Status CheckIndex(int32_t pos, int32_t length);
+
+    /// Writes the least significant `num_bytes` bytes in `value` into
+    /// `bytes[pos, pos + num_bytes)` in little endian.
+    static void WriteLong(uint8_t* bytes, int32_t pos, int64_t value, int32_t num_bytes);
+
```
lxy-9602, line 119: Please place the output parameters at the end of the function parameter list.

src/paimon/common/data/variant/variant_binary_util.cpp
```diff
@@ -0,0 +1,589 @@
...
+    } else {
+        result.append("0.");
+        result.append(static_cast<size_t>(scale) - digits.size(), '0');
+        result.append(digits);
+    }
+    return result;
+}
+
+Status VariantBinaryUtil::MalformedVariant() {
+    return Status::Invalid("MALFORMED_VARIANT");
```
lxy-9602: Could we include an error message parameter? Many places currently just return `MalformedVariant`, which may make debugging harder later.

src/paimon/common/data/variant/variant_get.h
```diff
@@ -0,0 +1,83 @@
...
+    /// Supported targets beyond scalars: a variant-marked struct field (the variant is deeply
+    /// re-encoded), STRUCT (from a variant object, children matched by name; unmatched children
+    /// are null), MAP with string keys (from a variant object), and LIST (from a variant array).
+    /// A nullptr `variant`, a variant null, or an invalid cast with `fail_on_error == false`
+    /// appends null.
+    static Status CastToBuilder(const std::shared_ptr<GenericVariant>& variant,
+                                const std::shared_ptr<arrow::Field>& target_field,
+                                arrow::ArrayBuilder* builder, const VariantCastArgs& cast_args,
+                                const std::shared_ptr<MemoryPool>& pool);
+
```
lxy-9602, line 74: Please put all output parameters at the end of the function parameter list.

src/paimon/common/data/variant/variant_get.h
```diff
@@ -0,0 +1,83 @@
...
+    /// Extracts a sub-variant by `path`, casts it to the (possibly nested) type of
+    /// `target_field`, and returns a length-1 arrow array holding the result (a null slot
+    /// represents SQL NULL). `arrow_pool` must outlive the returned array.
+    static Result<std::shared_ptr<arrow::Array>> GetAsArrow(
+        const std::shared_ptr<GenericVariant>& variant, const std::string& path,
+        const std::shared_ptr<arrow::Field>& target_field, const VariantCastArgs& cast_args,
+        const std::shared_ptr<MemoryPool>& pool, arrow::MemoryPool* arrow_pool);
```
lxy-9602: Could `pool` and `arrow_pool` use a consistent type, either both raw pointers or both smart pointers? The current usage looks a bit inconsistent.

src/paimon/common/data/variant/variant_json_utils.h
```diff
@@ -0,0 +1,59 @@
...
+class VariantJsonUtils {
+ public:
+    VariantJsonUtils() = delete;
+    ~VariantJsonUtils() = delete;
+
+    /// Appends `str` as a quoted JSON string with escaping into `out`.
+    static void AppendEscapedJson(std::string_view str, std::string* out);
+
+    /// Formats a double like `java.lang.Double.toString`, e.g. `1.0`, `-0.001`, `1.0E7`.
+    static std::string JavaDoubleToString(double value);
```
lszskye, line 38: This function appears to be intended to mimic the behavior of Java’s `Double.toString`. Could you add unit tests for this function, focusing on whether special values such as `Double.MIN_VALUE`, `Double.MAX_VALUE`, `-Infinity`, `-0.0`, `+0.0`, `+Infinity`, and `NaN` are consistent with Java?

src/paimon/common/data/variant/variant_get_test.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+TEST_F(VariantGetTest, NestedTargetNullSemantics) {
+    auto target = arrow::field("r", arrow::struct_({arrow::field("a", arrow::int64())}));
+    ASSERT_OK_AND_ASSIGN(std::shared_ptr<arrow::Array> array, GetAsArrow("$.missing", target));
+    ASSERT_TRUE(array->IsNull(0));
+    // A variant null yields SQL NULL even with fail_on_error == true.
+    cast_args_.fail_on_error = true;
+    ASSERT_OK_AND_ASSIGN(array, GetAsArrow("$.nullField", target));
+    ASSERT_TRUE(array->IsNull(0));
+}
```
lxy-9602, line 333: 1. Could we add a test where one STRUCT child fails to cast with `fail_on_error=false`, verifying that the child is null while the parent remains non-null?

2. Could we also add a test for casting a Variant null to a Variant target, to clarify the expected null semantics?

src/paimon/common/data/variant/variant_shredding_read_plan_factory.cpp
```diff
@@ -0,0 +1,341 @@
...
+            }
+            std::string_view metadata = static_cast<const arrow::BinaryArray&>(
+                                            *physical_struct.field(schema_->top_level_metadata_idx))
+                                            .GetView(row);
+            PAIMON_RETURN_NOT_OK_FROM_ARROW(struct_builder->Append());
+            for (size_t i = 0; i < specs_.size(); ++i) {
+                PAIMON_RETURN_NOT_OK(
+                    ExtractField(physical_struct, row, metadata, specs_[i],
+                                 struct_builder->field_builder(static_cast<int>(i))));
+            }
```
lxy-9602, line 388: int -> int32_t

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return configured_schema_.has_value();
+}
+
+bool VariantShreddingWritePlanFactory::ContainsVariantField() const {
+    for (const auto& field : write_schema_->fields()) {
+        if (VariantTypeUtils::IsVariantField(field)) {
+            return true;
+        }
+    }
+    return false;
```
lxy-9602: `ContainsVariantField` only checks top-level fields and does not recurse into nested ones. To avoid ambiguity with `VariantTypeUtils::ContainsVariantField` (which is recursive), could we rename it to `ContainsTopLevelVariantField`?

src/paimon/common/data/variant/variant_shredding_test.cpp
```diff
@@ -0,0 +1,248 @@
...
+class VariantShreddingTest : public ::testing::Test {
+ public:
+    // Shreds the given JSON documents (nullptr = null variant) with the given shredding type,
+    // asserts the reassembled variants render back to the same JSON, and returns the shredded
+    // array for structural checks.
+    std::shared_ptr<arrow::StructArray> RoundTrip(
+        const std::shared_ptr<arrow::DataType>& shredding_type,
+        const std::vector<const char*>& jsons) {
+        auto physical_result = VariantShreddingUtils::VariantShreddingSchema(shredding_type);
+        EXPECT_TRUE(physical_result.ok()) << physical_result.status().ToString();
```
lxy-9602: `EXPECT_TRUE(result.ok())` -> `EXPECT_OK_AND_ASSIGN(auto data, Func())`

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
```diff
@@ -0,0 +1,144 @@
...
+            config_status_ = result.status();
+        }
+    };
+    resolve(options.VariantInferShreddingSchemaEnabled(), &infer_enabled_);
+    resolve(options.GetVariantShreddingMaxSchemaWidth(), &max_schema_width_);
+    resolve(options.GetVariantShreddingMaxSchemaDepth(), &max_schema_depth_);
+    resolve(options.GetVariantShreddingMinFieldCardinalityRatio(), &min_field_cardinality_ratio_);
+    resolve(options.GetVariantShreddingMaxInferBufferRow(), &max_infer_buffer_row_);
+}
+
```
lxy-9602, line 133: If config parsing can fail, it should fail at `VariantShreddingWritePlanFactory` creation time rather than being deferred via `config_status_`. Please add a `Create` function for `VariantShreddingWritePlanFactory` and remove `config_status_`.

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    std::shared_ptr<VariantShreddingWritePlan> plan;
+    if (HasConfiguredShreddingSchema()) {
+        PAIMON_ASSIGN_OR_RAISE(plan, VariantShreddingWritePlan::FromConfiguredSchema(
+                                         write_schema_, configured_schema_.value()));
+    } else {
+        InferVariantShreddingSchema inferrer(max_schema_width_, max_schema_depth_,
+                                             min_field_cardinality_ratio_);
+        std::map<std::string, std::shared_ptr<arrow::DataType>> column_shredding_types;
```
lxy-9602: Could we align the `maxSchemaWidth` budget with Java? Java shares one `MaxFields` instance across all Variant columns, while C++ recreates it in each `InferColumnShreddingType` call, effectively applying the limit per column. This may allow the total inferred schema width to exceed the configured limit when multiple Variant columns exist. It would also be helpful to add a multi-column test with a small width limit.

src/paimon/common/data/variant/variant_shredding_write_plan_factory_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+    Result<CoreOptions> MakeOptions(std::map<std::string, std::string> options) {
+        // Keep the manifest format resolvable in test binaries without the avro plugin.
+        options.emplace("manifest.format", "parquet");
+        return CoreOptions::FromMap(options);
+    }
+
+    std::shared_ptr<arrow::Array> BuildBatch(const std::vector<const char*>& jsons) {
+        auto result = BuildVariantBatch(schema_->field(0), schema_->field(1), jsons, pool_);
+        EXPECT_TRUE(result.ok()) << result.status().ToString();
```
lxy-9602, line 53: Could we avoid having `BuildVariantBatch` as a global function? Prefer putting it into a test utility class as a static helper if possible.

src/paimon/common/data/variant/variant_shredding_writer.h
```diff
@@ -0,0 +1,103 @@
...
+        std::unique_ptr<Node> array_element;
+    };
+
+    VariantShreddedColumnWriter(const std::shared_ptr<VariantSchema>& schema,
+                                std::unique_ptr<arrow::ArrayBuilder>&& root_builder);
+
+    static Status BuildNode(const std::shared_ptr<VariantSchema>& schema,
+                            arrow::StructBuilder* group, Node* node);
+
+    Status AppendVariantNode(Node* node, const GenericVariant& variant);
```
lxy-9602: Could you please place the output parameters at the end of the function parameter list? (for `AppendVariantNode` and `TryTypedShred`)

src/paimon/common/data/variant/variant_shredding_writer.cpp
```diff
@@ -0,0 +1,468 @@
...
+    }
+    return result;
+}
+
+Status AppendDecimalTo(arrow::ArrayBuilder* builder, __int128_t unscaled) {
+    auto* decimal_builder = static_cast<arrow::Decimal128Builder*>(builder);
+    arrow::Decimal128 value(static_cast<int64_t>(unscaled >> 64),
+                            static_cast<uint64_t>(static_cast<__uint128_t>(unscaled)));
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(decimal_builder->Append(value));
+    return Status::OK();
```
lxy-9602, line 57: Please place the output parameters at the end of the function parameter list.

src/paimon/common/data/variant/variant_path_segment.cpp
```diff
@@ -0,0 +1,115 @@
...
+bool TryParseQuotedKey(std::string_view* remaining, std::string* key) {
+    std::string_view s = *remaining;
+    if (s.size() < 4 || s[0] != '[' || (s[1] != '\'' && s[1] != '"')) {
+        return false;
+    }
+    char quote = s[1];
+    size_t end = s.find(quote, 2);
+    if (end == std::string_view::npos || end == 2 || end + 1 >= s.size() || s[end + 1] != ']') {
+        return false;
+    }
```
lszskye, line 63: Just a curious question: since `end` cannot be `2`, a valid JSON object like `{"": 1}` can be encoded, but cannot be queried through a variantGet path. Is that correct?

src/paimon/core/io/infer_shredding_file_writer_test.cpp
```diff
@@ -0,0 +1,212 @@
...
+    std::unique_ptr<InferShreddingFileWriter<::ArrowArray*, std::shared_ptr<DataFileMeta>>>
+    MakeWriter(int32_t buffer_rows) {
+        std::map<std::string, std::string> option_map = {
+            {"variant.inferShreddingSchema", "true"},
+            {"variant.shredding.maxInferBufferRow", std::to_string(buffer_rows)},
+            // Keep the manifest format resolvable in test binaries without the avro plugin.
+            {"manifest.format", "parquet"}};
+        auto options = CoreOptions::FromMap(option_map);
+        EXPECT_TRUE(options.ok()) << options.status().ToString();
+        auto plan_factory =
```
lxy-9602: Please use `EXPECT_OK_AND_ASSIGN`

src/paimon/core/operation/abstract_split_read.cpp
```diff
@@ -281,6 +286,32 @@ AbstractSplitRead::ApplySharedShreddingReaderIfNeeded(
...
+    }
+    if (!has_variant_field) {
+        return std::move(file_reader);
+    }
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<::ArrowSchema> file_schema,
+                           file_reader->GetFileSchema());
+    PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Schema> file_arrow_schema,
+                                      arrow::ImportSchema(file_schema.get()));
+    PAIMON_ASSIGN_OR_RAISE(auto plans, VariantShreddingReadPlanFactory::CreateReadPlans(
+                                           read_schema, file_arrow_schema, pool_));
```
lxy-9602: Please avoid `PAIMON_ASSIGN_OR_RAISE(auto, ...)` in non-test / production code.

src/paimon/common/data/variant/variant_builder.cpp
```diff
@@ -0,0 +1,652 @@
...
+    bool Ok(const Status& status) {
+        if (!status.ok()) {
+            status_ = status;
+            return false;
+        }
+        return true;
+    }
+
+    // Records the offset of the value that is about to be appended in the enclosing container.
+    Status BeforeValue() {
```
lszskye: This function does not appear to return a bad status under any circumstances. Can we use `void BeforeValue()`

src/paimon/core/core_options.cpp
```diff
@@ -1191,6 +1191,42 @@ Result<bool> CoreOptions::FieldCollectAggDistinct(const std::string& field_name)
...
+    if (it == impl_->raw_options.end() || it->second.empty()) {
+        return std::nullopt;
+    }
+    return it->second;
+}
+
+Result<bool> CoreOptions::VariantInferShreddingSchemaEnabled() const {
+    return OptionsUtils::GetValueFromMap<bool>(impl_->raw_options,
+                                               Options::VARIANT_INFER_SHREDDING_SCHEMA, false);
+}
```
lxy-9602, line 1341: I wonder if it would be better to validate this in parse, like the other options, rather than only when it is used. That might help catch invalid configuration earlier.

src/paimon/testing/utils/variant_test_data.h
```diff
@@ -0,0 +1,64 @@
...
+#include "paimon/memory/memory_pool.h"
+#include "paimon/result.h"
+
+namespace paimon::test {
+
+/// Builds a `[id, v]` struct array where `v` holds the variant encodings of `jsons` (nullptr
+/// means a null variant) and `id` is a running int32 starting at `id_offset`. `variant_field`
+/// must be a variant-marked field (see `VariantTypeUtils::ToArrowField`).
+inline Result<std::shared_ptr<arrow::StructArray>> BuildVariantBatch(
+    const std::shared_ptr<arrow::Field>& id_field,
```
lxy-9602, line 46: Please avoid using a global function if possible. Prefer putting it into a test utility class as a static function.

test/inte/variant_table_inte_test.cpp
```diff
@@ -0,0 +1,233 @@
...
+        result_struct->field(struct_type->GetFieldIndex("v")));
+    const auto& age = static_cast<const arrow::Int64Array&>(*v_column->field(0));
+    const auto& other = static_cast<const arrow::StringArray&>(*v_column->field(1));
+    ASSERT_EQ(age.Value(0), 35);
+    ASSERT_EQ(age.Value(1), 25);
+    ASSERT_TRUE(v_column->IsNull(2));
+    ASSERT_TRUE(other.IsNull(0));
+    ASSERT_EQ(other.GetString(1), "Hello");
+}
+
```
lxy-9602, line 263: Could we add a more comprehensive JSON test case here, with deeper nesting and a mix of different value types?

src/paimon/format/parquet/variant_parquet_test.cpp
```diff
@@ -0,0 +1,568 @@
...
+        ASSERT_OK_AND_ASSIGN(std::shared_ptr<VariantShreddingBatchConverter> converter,
+                             VariantShreddingBatchConverter::Create(plan, pool_));
+        auto logical = BuildArray(jsons);
+        auto c_logical = std::make_unique<ArrowArray>();
+        ASSERT_TRUE(arrow::ExportArray(*logical, c_logical.get()).ok());
+        ASSERT_OK_AND_ASSIGN(std::unique_ptr<ArrowArray> c_physical,
+                             converter->Convert(c_logical.get()));
+        ASSERT_OK_AND_ASSIGN(
+            std::shared_ptr<arrow::Schema> write_schema,
+            ParquetFieldIdConverter::AddParquetIdsFromPaimonIds(converter->GetPhysicalSchema()));
```
lxy-9602: Could we reuse `WriteFile` here?

test/inte/variant_table_inte_test.cpp
lxy-9602: Also, could you generate a test dataset using the Java implementation, put it under `test_data`, and have the C++ side read it to verify compatibility?

src/paimon/format/parquet/variant_parquet_test.cpp
```diff
@@ -0,0 +1,568 @@
...
+}  // namespace
+
+TEST_F(VariantParquetTest, PhysicalLayoutMatchesJava) {
+    auto array = BuildArray({R"({"a": 1, "b": "hello"})", nullptr, "[1,2,3]"});
+    WriteFile(array);
+
+    // The on-disk layout must match the Java ParquetSchemaConverter: an (unannotated) group
+    // with two REQUIRED BINARY fields `value` (id 0) and `metadata` (id 1).
+    auto file = arrow::io::ReadableFile::Open(file_path_, arrow_pool_.get());
+    ASSERT_TRUE(file.ok());
```
lxy-9602, line 480: I’m a bit curious—why do we need to use the raw Parquet reader here? Would it also work to use the Paimon Parquet reader to get the file schema?

test/inte/variant_table_inte_test.cpp
lxy-9602: Could you also add a test covering an I/O exception scenario? The goal is to verify that read/write operations report an error cleanly, without causing a core dump or memory leak. You can refer to `ScanAndReadInteTest.TestWithAppendSnapshotIOException`.

src/paimon/common/data/variant/variant_shredding_write_plan.cpp
lszskye: +1

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
```diff
@@ -0,0 +1,144 @@
...
+            return true;
+        }
+    }
+    return false;
+}
+
+Result<std::shared_ptr<ShreddingBatchConverter>> VariantShreddingWritePlanFactory::CreateConverter(
+    const std::string& file_format_identifier,
+    const std::vector<std::shared_ptr<arrow::Array>>& sample_batches) const {
+    PAIMON_RETURN_NOT_OK(config_status_);
```
lszskye: Even if the user has configured variant.shreddingSchema, CreateConverter() will still fail if variant.shredding.maxSchemaWidth is invalid. This behavior is different from the Java implementation.

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
```diff
@@ -0,0 +1,144 @@
...
+                                             min_field_cardinality_ratio_);
+        std::map<std::string, std::shared_ptr<arrow::DataType>> column_shredding_types;
+        for (int i = 0; i < write_schema_->num_fields(); ++i) {
+            const std::shared_ptr<arrow::Field>& field = write_schema_->field(i);
+            if (!VariantTypeUtils::IsVariantField(field)) {
+                continue;
+            }
+            std::vector<std::shared_ptr<GenericVariant>> samples;
+            for (const auto& sample_batch : sample_batches) {
+                const auto& struct_array = static_cast<const arrow::StructArray&>(*sample_batch);
```
lszskye: If you’re fairly certain about the type, use `arrow::internal::checked_pointer_cast` will be better

src/paimon/common/data/variant/variant_shredding_write_plan_factory_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+class VariantShreddingWritePlanFactoryTest : public ::testing::Test {
+ public:
+    void SetUp() override {
+        pool_ = GetDefaultPool();
+        std::vector<DataField> fields = {DataField(1, arrow::field("id", arrow::int32())),
+                                         DataField(2, VariantTypeUtils::ToArrowField("v"))};
+        schema_ = DataField::ConvertDataFieldsToArrowSchema(fields);
+    }
+
+    Result<CoreOptions> MakeOptions(std::map<std::string, std::string> options) {
```
lszskye: Result<CoreOptions> MakeOptions(std::map<std::string, std::string> options) const

src/paimon/core/io/shredding_key_value_data_file_writer_factory.cpp
```diff
@@ -59,8 +75,6 @@ ShreddingKeyValueDataFileWriterFactory::CreateWriter() const {
         ArrowArrayMove(physical.get(), array);
         return Status::OK();
     };
-
-    auto format = options_.GetWriteFileFormat(level_);
     PAIMON_ASSIGN_OR_RAISE(WriterResources resources,
```
lszskye, line 78: why remove auto format = options_.GetWriteFileFormat(level_);

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+int32_t VariantShreddingWritePlanFactory::InferBufferRowCount() const {
+    return max_infer_buffer_row_;
+}
+
+bool VariantShreddingWritePlanFactory::HasConfiguredShreddingSchema() const {
+    return configured_schema_.has_value();
+}
+
+bool VariantShreddingWritePlanFactory::ContainsVariantField() const {
```
lszskye: The Java implementation supports VARIANT fields nested inside ROW types during schema inference and write conversion. InferVariantShreddingSchema recursively discovers VARIANT paths through nested RowTypes, and VariantShreddingWritePlan recursively converts the corresponding nested rows.
In contrast, the C++ implementation only iterates over the top-level fields of write_schema_ and checks VariantTypeUtils::IsVariantField(field). As a result, a VARIANT nested inside a struct is neither detected, inferred, nor converted.
Could we recursively handle VARIANT fields nested inside structs to align with the Java implementation? VARIANT fields nested inside arrays or maps do not need to be included, as Java does not support shredding them either.

src/paimon/common/data/variant/variant_shredding_read_plan_factory.cpp
```diff
@@ -0,0 +1,341 @@
...
+    Result<std::shared_ptr<arrow::Array>> Assemble(const std::shared_ptr<arrow::Array>& physical,
+                                                   arrow::MemoryPool* pool) const override {
+        if (physical->type_id() != arrow::Type::STRUCT) {
+            return Status::Invalid(fmt::format("cannot cast shredded variant field {} to a struct",
+                                               physical_field_->name()));
+        }
+        const auto& physical_struct = static_cast<const arrow::StructArray&>(*physical);
+        std::unique_ptr<arrow::ArrayBuilder> builder;
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::MakeBuilder(pool, logical_field_->type(), &builder));
+        auto* struct_builder = static_cast<arrow::StructBuilder*>(builder.get());
```
lszskye, line 371: call builder->Reserve here

src/paimon/common/data/variant/variant_get.cpp
```diff
@@ -0,0 +1,407 @@
...
+                           FieldTypeUtils::ConvertToFieldType(input_type->id()));
+    PAIMON_ASSIGN_OR_RAISE(FieldType target_field_type,
+                           FieldTypeUtils::ConvertToFieldType(target_type->id()));
+    std::shared_ptr<CastExecutor> executor =
+        CastExecutorFactory::GetCastExecutorFactory()->GetCastExecutor(input_field_type,
+                                                                       target_field_type);
+    if (executor == nullptr) {
+        return InvalidCast(variant, target_type, cast_args);
+    }
+    Result<Literal> cast_result = executor->Cast(*input, target_type);
```
lszskye, line 193: `GenericVariant::ToJson()` formats floating-point values using `JavaDoubleToString()`, while `VariantGetExecutor::CastVariant()` delegates DOUBLE-to-STRING conversion to Arrow's cast executor. Since these implementations produce different string representations for the same value, this causes two consistency issues:
(1)Paimon C++ may return different results for the same Variant value depending on whether it is converted through ToJson() or variant_get(..., STRING).
(2)variant_get(..., STRING) may return different results in Paimon C++ and Java, because Java uses Double.toString() for this conversion.

src/paimon/common/data/shredding/shredding_file_reader.cpp
```diff
@@ -0,0 +1,131 @@
...
+    auto resolved_schema = arrow::schema(resolved_fields, logical_read_schema->metadata());
+    auto c_resolved_schema = std::make_unique<::ArrowSchema>();
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::ExportSchema(*resolved_schema, c_resolved_schema.get()));
+    return reader_->SetReadSchema(c_resolved_schema.get(), predicate, selection_bitmap);
+}
+
+Result<BatchReader::ReadBatch> ShreddingFileReader::NextBatch() {
+    return Status::Invalid("paimon inner reader ShreddingFileReader should use NextBatchWithBitmap");
+}
+
```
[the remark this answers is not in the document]
SteNicholas: `NextBatch()` intentionally returns `Invalid` to stay consistent with the existing inner-reader convention: `MapSharedShreddingFileReader::NextBatch()` does the same, and the paimon read chain always drives inner readers through `NextBatchWithBitmap()`.

src/paimon/common/data/shredding/shredding_write_plan_factory.h
SteNicholas: Done: `CreateMetadataFinalizer` is `const = 0` now, and `VariantShreddingWritePlanFactory` overrides it returning nullptr (the shredded physical schema is self-describing).

src/paimon/common/data/variant/infer_variant_shredding_schema.cpp
SteNicholas: Done: replaced with the explicit type `std::optional<GenericVariant::ObjectField>` here and at the other `GetFieldAtIndex` call sites.

src/paimon/common/data/shredding/shredding_file_reader.h
SteNicholas: The wrapping itself happens in `AbstractSplitRead::CreateFieldMappingReader`, which all three split reads share via `CreateRawFileReaders`, so the read path was already covered. Added `(ShreddingFileReader)` to the Readers Overview doc of `RawFileSplitRead`, `MergeFileSplitRead` and `DataEvolutionSplitRead`.

src/paimon/common/types/data_type.cpp
SteNicholas: The guard existed because the `STRUCT` case fell through into `LARGE_BINARY`. Restructured the switch so both cases `break` to a shared throw instead of falling through; the redundant check is gone.

src/paimon/common/utils/field_type_utils.h
SteNicholas: Done: `ConvertToFieldType(field)` uses `VariantTypeUtils::IsVariantField` now.

src/paimon/common/data/variant/infer_variant_shredding_schema_test.cpp
SteNicholas: Done: the test now asserts the exact inferred type (`struct{a: int64}`, derived from the budget arithmetic: root costs 1, `a` costs 2) and additionally covers the budget being shared across columns.

src/paimon/common/data/variant/variant.cpp
SteNicholas: Done.

src/paimon/common/data/variant/variant_access_utils.cpp
SteNicholas: Done: `GetDescription` returns `std::string` now.

src/paimon/core/core_options.cpp
SteNicholas: Done: added `CoreOptionsTest.TestVariantOptions` covering the defaults, configured values, the `parquet.variant.shreddingSchema` fallback key, and invalid values failing at parse time.

src/paimon/common/data/variant/variant_reassembler.cpp
SteNicholas: Good catch: the reassembler assumed microseconds but nothing validated it. `BuildVariantSchema` now accepts a TIMESTAMP `typed_value` only with microsecond unit (the variant binary stores micros) and rejects other precisions as an invalid shredding schema; `ScalarSchemaToArrowType` maps back to `timestamp(MICRO)`.

src/paimon/common/data/variant/variant_reassembler.cpp
SteNicholas: Done: `AssembleVariantArray` now takes two pools, the caller-provided paimon pool for variant rebuilding and the arrow pool for the output array; `GetDefaultPool()` is gone.

src/paimon/common/data/variant/variant_reassembler.cpp
SteNicholas: Done: added the TODO to revisit returning the physical arrays directly for unshredded files.

src/paimon/common/data/variant/variant_shredding_write_plan.cpp
SteNicholas: Done: `VariantShreddingWritePlan` now has a private constructor taking all members; `Create`/`CreateFromPaths` build the values and call it.

src/paimon/common/data/variant/variant_binary_util.h
SteNicholas: Done: `WriteLong(int64_t value, int32_t num_bytes, uint8_t* bytes, int32_t pos)`.

src/paimon/common/data/variant/variant_binary_util.cpp
SteNicholas: Done: added an optional `message` parameter (appended as `MALFORMED_VARIANT: <message>`) and filled the call sites with specific context (out-of-bounds layouts, non-monotonic metadata offsets, null shredded groups, etc.).

src/paimon/common/data/variant/variant_get.h
SteNicholas: Done: `builder` moved to the end (`CastToBuilder(variant, target_field, cast_args, pool, builder)`).

src/paimon/common/data/variant/variant_get.h
SteNicholas: Done: both are smart pointers now (`const std::shared_ptr<MemoryPool>&` and `const std::shared_ptr<arrow::MemoryPool>&`).

src/paimon/common/data/variant/variant_json_utils.h
SteNicholas: Done: added `variant_json_utils_test.cpp` covering +-0.0, +-Infinity, NaN, MIN_VALUE, MAX_VALUE and MIN_NORMAL for both double and float, verified against Java 8 output. This exposed a real divergence: we printed `Double.MIN_VALUE` as `5.0E-324` while Java prints `4.9E-324` -- Java's `FloatingDecimal` emits at least two significant digits in scientific form. Fixed the formatter accordingly.

src/paimon/common/data/variant/variant_get_test.cpp
SteNicholas: Done: (1) `CastToStructTarget` now also covers a child failing to cast with `fail_on_error=false` (child null, parent row kept); (2) `CastToVariantTarget` covers a variant null cast to a VARIANT target -- it stays an encoded variant null (rendering as JSON `null`), while an unmatched path is SQL NULL.

src/paimon/common/data/variant/variant_shredding_read_plan_factory.cpp
SteNicholas: Done.

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
SteNicholas: Renamed to `ContainsShreddableVariantField`, and with the nested-ROW support it now recurses through structs only (arrays/maps excluded), so the name reflects exactly what is shreddable.

src/paimon/common/data/variant/variant_shredding_test.cpp
SteNicholas: Done: the round-trip helper uses `EXPECT_OK_AND_ASSIGN` throughout.

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
SteNicholas: Done: the `variant.*` options are parsed and validated in `CoreOptions::FromMap` now, `VariantShreddingWritePlanFactory::Create` builds the factory from the validated options, and `config_status_` is removed.

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
SteNicholas: Done: one `MaxFields` budget is created per schema (`CreateMaxFieldsBudget`) and shared across all variant columns, mirroring Java. Added multi-column small-limit tests at both the inferrer level and the factory level (`SharedWidthBudgetAcrossColumns`).

src/paimon/common/data/variant/variant_shredding_write_plan_factory_test.cpp
SteNicholas: Done: moved into the `VariantTestData` test utility class as a static helper.

src/paimon/common/data/variant/variant_shredding_writer.h
SteNicholas: Done for both: `AppendVariantNode(variant, node)` and `TryTypedShred(variant, variant_type, node, shredded)`.

src/paimon/common/data/variant/variant_shredding_writer.cpp
SteNicholas: Done: `AppendDecimalTo(unscaled, builder)`.

src/paimon/common/data/variant/variant_path_segment.cpp
SteNicholas: Yes, that is correct -- `{"": 1}` can be encoded but cannot be addressed through a variantGet path. This matches Java: `VariantPathSegment.KEY_PATTERN` uses `\['([^']+)']`, which also requires a non-empty key.

src/paimon/core/io/infer_shredding_file_writer_test.cpp
SteNicholas: Done.

src/paimon/core/operation/abstract_split_read.cpp
SteNicholas: Done: replaced with the explicit `std::map<std::string, std::shared_ptr<ShreddingColumnReadPlan>>` type.

src/paimon/common/data/variant/variant_builder.cpp
SteNicholas: Done: `void BeforeValue()`.

src/paimon/core/core_options.cpp
SteNicholas: Done: all `variant.*` options are parsed and validated in `CoreOptions::FromMap` (`ParseVariantOptions`, including range checks), like the other options; the getters just return the parsed members. Invalid configuration now fails when the options are created rather than when they are used.

src/paimon/testing/utils/variant_test_data.h
SteNicholas: Done: wrapped into the `VariantTestData` class as a static function.

test/inte/variant_table_inte_test.cpp
SteNicholas: Done for two of the three: (1) extended `TestAppendTable` with a comprehensive document set (deep object/array alternation, escaped and unicode strings, wide integers, decimals, exponent doubles, empty containers); (2) added `TestReadWithIOException` and `TestWriteWithIOException` following the `TestWithAppendSnapshotIOException` pattern, injecting an IO error at every position of the scan+read and create+write+commit paths.

src/paimon/format/parquet/variant_parquet_test.cpp
SteNicholas: Done: `WriteShreddedFile` now delegates to a schema-parameterized `WriteFile` overload.

src/paimon/format/parquet/variant_parquet_test.cpp
SteNicholas: The raw parquet reader is needed here because the test asserts parquet-level physical properties -- REQUIRED repetition, BYTE_ARRAY physical types, field ids and the absence of a logical-type annotation -- which are not visible in the Arrow schema surfaced by the paimon reader. Added a comment in the test explaining this.

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
SteNicholas: Fixed by moving the option parsing/validation into `CoreOptions::FromMap`: `CreateConverter` no longer parses or validates configuration at all, so a configured `variant.shreddingSchema` can never fail on an unrelated option there. (An invalid `maxSchemaWidth` value now fails once at options-parse time, which is the established convention for all options in the C++ implementation.)

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
SteNicholas: Done: using `arrow::internal::checked_cast` for the known array types here and in the converter.

src/paimon/common/data/variant/variant_shredding_write_plan_factory_test.cpp
SteNicholas: Done.

src/paimon/core/io/shredding_key_value_data_file_writer_factory.cpp
SteNicholas: It wasn't removed -- it moved into `CreateShreddedWriter` when the write path was split into `CreateWriter`/`CreateShreddedWriter`. The per-level format is still used both for the format identifier and for `CreateWriterResources`.

src/paimon/common/data/variant/variant_shredding_write_plan_factory.cpp
SteNicholas: Done: inference now discovers VARIANT paths through nested RowTypes by field-index path (mirroring `InferVariantShreddingSchema.getPathsToVariant`), `VariantShreddingWritePlan`/the batch converter recursively convert the nested structs, and the read side reassembles nested shredded variants transparently. Arrays/maps are excluded and the configured `variant.shreddingSchema` stays top-level-only, both as in Java. Rows whose enclosing struct is null are skipped when sampling and shredded to null (their child slot contents are unspecified in Arrow). Covered by new inferrer, factory and reader tests.

src/paimon/common/data/variant/variant_shredding_read_plan_factory.cpp
SteNicholas: Done: `Reserve(physical_struct.length())` before the row loop.

src/paimon/common/data/variant/variant_get.cpp
SteNicholas: Done: `CastVariant` now renders DOUBLE/FLOAT to STRING via `JavaDoubleToString`/`JavaFloatToString`, so `ToJson()` and `variant_get(..., STRING)` agree with each other and with Java's `Double.toString`/`Float.toString`.

SteNicholas:
@lxy-9602, @lszskye, thanks for detailed review. I have address above comments. PTAL.

lxy-9602:
Thanks for adding write support for Variants nested in structs. The corresponding read path still seems to differ from Java: Java recursively recognizes a nested Variant access type and extracts the requested fields, while `BuildNestedVariantPlan` currently treats it as a full Variant and reassembles `struct<value, metadata>`, which does not match the requested access struct. Could we support nested Variant access here, or explicitly reject it with a clear error? A nested read/access test would also be helpful.

SteNicholas:
@lxy-9602 Thanks — the diagnosis was exact.

Root cause: `VariantAccessBuilder::Build` stamps the variant extension marker on the projection field, so `IsVariantField()` returns true for a variant-access type, and `BuildNestedVariantPlan` tested that before ever considering the access type. Two failure modes followed: on a shredded file it reassembled `struct<value, metadata>` into a column declared with the access struct type; on an unshredded file it built no plan at all, so the projection type reached the format reader.

Supported it rather than rejecting, following `VariantShreddingReadPlanFactory` on the Java side:

- Recognise the access type at every nesting level and share one leaf-plan factory between the top level and nested positions, so a nested projection also prunes the scan (`typed_value` narrowed to the requested keys, `value` dropped when everything requested is shredded).
- Descend through ARRAY and MAP as well, as `buildPhysicalType` / `assembleVector` do in Java. One deliberate difference: the C++ parquet reader rejects partial projection inside a repeated group, so there the file subtree is pushed down verbatim and only reassembled back — no pruning below a repeated level.

Adding the end-to-end coverage you asked for surfaced two further gaps:

- `NestedProjectionUtils::PruneDataType` fails fast on any projection inside a repeated group, which blocked the entire table read path for ARRAY/MAP. Replacing a variant by its access projection drops no field, so it now passes through while a real partial projection keeps failing. This mirrors the existing ROW escape hatch there; Java's `FormatReaderMapping.pruneDataType` recurses freely because its readers support partial projection inside repeated groups.
- `ArrayType::ToJson` / `MapType::ToJson` dropped their children's metadata, so `ARRAY<VARIANT>` serialized as `ARRAY<ROW<value, metadata>>` and such a table could not be created at all — that struct's fixed child ids 0/1 collide with the global field id check on read-back. Split into its own commit since it affects any extension type nested in an array or map, VARIANT and BLOB alike.

Tests: 8 format-level cases in `variant_parquet_test.cpp` (nested ROW on shredded and unshredded files, ARRAY, MAP, `ARRAY<ROW<VARIANT>>`, plus the partial-projection guard), 4 in `nested_projection_utils_test.cpp`, 1 in `data_type_test.cpp`, and 2 end-to-end in `variant_table_inte_test.cpp`.

PTAL.

lxy-9602 (dismissed):
+1

lxy-9602 (approved):
+1

lszskye (approved):
+1

PR #407, opened 2026-07-10
chore: adjust set read_ranges_freshed_ in SetReadRanges

lszskye:
### Purpose
adjust set read_ranges_freshed_ in SetReadRanges


lxy-9602 (approved):
+1

PR #412, opened 2026-07-10
fix: compile with correct flags with AppleClang

### LINKED ISSUE #411: [Bug] failed to build on macos
### Search before asking

### Paimon-cpp version

master as of 34833166d8c7dfffe9981e42e05d39e79aec30db

### Minimal reproduce step

configure and build on macos, with no extra configure args

### What doesn't meet your expectations?

it should compile but did not

### Anything else?

BuildUtils.cmake should match against "AppleClang" as well, just like that in SetupCxxFlags.cmake

smdsbz:
### Purpose

Linked issue: close #411


zjw1111 (approved):
Nice catch! +1

PR #414, opened 2026-07-13
feat(blob view): support blob-view.resolve.enabled to preserve blob view references at read time

### LINKED ISSUE #413: [Feature] Support blob-view.resolve.enabled to preserve BlobViewStruct references at read time
### Search before asking

### Motivation

Java Paimon provides the `blob-view.resolve.enabled` option (default `true`, see `CoreOptions.BLOB_VIEW_RESOLVE_ENABLED` in apache/paimon) to control whether `blob-view-field` values are resolved from upstream tables at read time. Setting it to `false` preserves the serialized `BlobViewStruct` bytes, which is required when forwarding blob view values from one blob-view table to another instead of materializing them.

paimon-cpp currently always resolves blob-view fields whenever they appear in the read schema: `DataEvolutionSplitRead::WrapWithBlobViewResolverIfNeeded` unconditionally runs the prescan, builds the resolver, and fails with `invalid config for blob view, supposed to set BLOB_VIEW_UPSTREAM_WAREHOUSE` if the upstream warehouse is not configured. There is no way to opt out of resolution, so the forwarding scenario supported by Java cannot be implemented with paimon-cpp, leaving a behavior gap between the two implementations.

### Solution

Align with the Java implementation:

- Add `Options::BLOB_VIEW_RESOLVE_ENABLED` (`"blob-view.resolve.enabled"`, default `true`) and parse it into `CoreOptions` alongside the other blob-view options, exposed as `CoreOptions::BlobViewResolveEnabled()`.
- In `DataEvolutionSplitRead::WrapWithBlobViewResolverIfNeeded`, return the inner reader as-is when the option is disabled: the read then passes the serialized `BlobViewStruct` bytes through untouched, skips the prescan/resolver entirely, and no longer requires `blob-view-upstream-warehouse` (matching Java, where no catalog access is needed when resolution is off).
- Since read-context options are merged on top of table schema options, the option can be set per read as well as in the table schema.
- Tests: `CoreOptionsTest` assertions for the default and parsed values, plus an integration test verifying that with resolve disabled the read returns byte-identical `BlobViewStruct` payloads (deserializable back to the original identifier/fieldId/rowId) against a non-existent upstream table, and that re-enabling the option on the same read attempts resolution again.

### Anything else?

The option key, default value, and semantics follow `CoreOptions.BLOB_VIEW_RESOLVE_ENABLED` in apache/paimon (introduced with BlobView support in apache/paimon#7602).

SteNicholas:
### Purpose

Linked issue: close #413

Align with `CoreOptions.BLOB_VIEW_RESOLVE_ENABLED` in apache/paimon. This PR adds the `blob-view.resolve.enabled` option (default `true`). When set to `false`, reading a table with `blob-view-field` passes the serialized `BlobViewStruct` bytes through without resolving them from upstream tables: `DataEvolutionSplitRead::WrapWithBlobViewResolverIfNeeded` returns the inner reader as-is, skipping the prescan/resolver entirely and no longer requiring `blob-view-upstream-warehouse`. This enables forwarding blob view references from one blob-view table to another instead of materializing them. Since read-context options are merged on top of table schema options, the option can be disabled per read while the table keeps the default, which is the primary usage pattern.

### Tests

- UT `CoreOptionsTest.TestDefaultValue`: the option defaults to `true`.
- UT `CoreOptionsTest.TestFromMap`: `blob-view.resolve.enabled=false` is parsed into `CoreOptions::BlobViewResolveEnabled()`.
- IT `BlobTableInteTest.TestForwardBlobViewReference`: end-to-end forward scenario aligned with `BlobTableTest#testForwardBlobViewReference` in apache/paimon — the source blob-view table keeps the default `true` without an upstream warehouse, so its default read fails while a read with the option dynamically disabled returns the serialized `BlobViewStruct` bytes byte-identically; the preserved references are then written into a second blob-view table, which still stores the original upstream references (verified via another resolve-disabled read) and resolves them to the actual upstream blob values on a default read.

### API and Format

Adds the public option constant `Options::BLOB_VIEW_RESOLVE_ENABLED` in `include/paimon/defs.h` and the getter `CoreOptions::BlobViewResolveEnabled()`. No storage format or protocol change.

### Documentation

The option is documented in the `include/paimon/defs.h` doc comment, following the existing blob-view options.

### Generative AI tooling

### DIFF
--- a/include/paimon/defs.h
+++ b/include/paimon/defs.h
@@ -408,6 +408,10 @@ struct PAIMON_EXPORT Options {
     /// serialized BlobViewStruct bytes inline in data files and resolve from upstream tables at
     /// read time. No default value.
     static const char BLOB_VIEW_FIELD[];
+    /// "blob-view.resolve.enabled" - Whether to resolve blob-view-field values from upstream
+    /// tables at read time. Set to false to preserve serialized BlobViewStruct bytes when
+    /// forwarding blob view values to another blob-view table. Default value is "true".
+    static const char BLOB_VIEW_RESOLVE_ENABLED[];
     /// "blob-view-upstream-warehouse" - Since the catalog capabilities are partially missing, when
     /// Blob View is enabled, cpp paimon cannot automatically obtain the upstream table warehouse
     /// path and requires manual configuration by the user. No default value.
--- a/src/paimon/common/defs.cpp
+++ b/src/paimon/common/defs.cpp
@@ -102,6 +102,7 @@ const char Options::BLOB_FIELD[] = "blob-field";
 const char Options::BLOB_DESCRIPTOR_FIELD[] = "blob-descriptor-field";
 const char Options::FALLBACK_BLOB_DESCRIPTOR_FIELD[] = "blob.stored-descriptor-fields";
 const char Options::BLOB_VIEW_FIELD[] = "blob-view-field";
+const char Options::BLOB_VIEW_RESOLVE_ENABLED[] = "blob-view.resolve.enabled";
 const char Options::BLOB_VIEW_UPSTREAM_WAREHOUSE[] = "blob-view-upstream-warehouse";
 const char Options::GLOBAL_INDEX_ENABLED[] = "global-index.enabled";
 const char Options::GLOBAL_INDEX_THREAD_NUM[] = "global-index.thread-num";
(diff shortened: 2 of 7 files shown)


lszskye (approved):
+1

PR #60, opened 2026-01-18
fix: glog linking error when libunwind is present

mrdrivingduck:
### Purpose

When glog is built on systems with libunwind installed, it automatically detects and links against libunwind for stack unwinding support. However, the `INTERFACE_LINK_LIBRARIES` property was not set, causing undefined reference errors when linking static libraries:

```
/usr/bin/ld: ../../glog_ep-install/lib/libglog.a(stacktrace.cc.o): in function `google::glog_internal_namespace_::GetStackTrace(void**, int, int)':
stacktrace.cc:(.text+0x68): undefined reference to `_Ux86_64_getcontext'
/usr/bin/ld: stacktrace.cc:(.text+0x73): undefined reference to `_ULx86_64_init_local'
/usr/bin/ld: stacktrace.cc:(.text+0xa1): undefined reference to `_ULx86_64_get_reg'
/usr/bin/ld: stacktrace.cc:(.text+0xb4): undefined reference to `_ULx86_64_step'
/usr/bin/ld: stacktrace.cc:(.text+0xf7): undefined reference to `_ULx86_64_step'
collect2: error: ld returned 1 exit status
```

This change dynamically detects libunwind availability and adds it to glog's link interface only when present, ensuring portability across different build environments.

According to the docs of glog, [libunwind is recommended](https://google.github.io/glog/0.7.1/unwinder/).

### Tests

For system without libunwind:

```
$ ldd ./usr/local/lib64/libpaimon.so 
        linux-vdso.so.1 (0x00007fff863ab000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f0b897cf000)
        libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f0b87c10000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f0b87aca000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f0b897b4000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f0b87aa8000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f0b878cc000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f0b897ee000)
```

For system with libunwind:

```
$ ldd ./usr/local/lib64/libpaimon.so
        linux-vdso.so.1 (0x00007ffcf1b3e000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f52b4a12000)
[...]

PR #415, opened 2026-07-13
fix(build): disable glog unwind without libunwind

mrdrivingduck:
### Purpose

Linked issue: N/A

This is a follow-up to #60. That change fixed static-link errors in regular configurations by adding a discovered libunwind library to the imported `glog` target's link interface. However, it made that decision after bundled glog had already been configured, so an explicit `LIBUNWIND_LIBRARY=FALSE` setting was not propagated to glog.

A downstream build can explicitly set `-DLIBUNWIND_LIBRARY=FALSE` to prevent paimon-cpp from using or propagating libunwind. Before this change, paimon-cpp skipped adding libunwind to the imported `glog` target, but glog still used its default unwind auto-detection. When libunwind was installed and discoverable by glog's CMake configuration, glog enabled libunwind support and compiled stacktrace code with `unw_*` references.

The inconsistency becomes visible when a downstream application links the resulting static glog library: paimon-cpp does not propagate `-lunwind`, while glog contains calls that require it, which can produce unresolved `unw_*` symbols. Downstream users must then carry a glog-specific patch or otherwise override glog's unwind configuration.

Use the same `LIBUNWIND_LIBRARY` result for both decisions. When libunwind is unavailable or explicitly disabled, pass `-DWITH_UNWIND=none` to bundled glog and omit the libunwind link dependency. Otherwise, retain the existing automatic discovery and linking behavior.

### Tests

- `cmake --build build --target glog_ep -j 4`
- `pre-commit run --files cmake_modules/ThirdpartyToolchain.cmake`
- `git diff --check`

### API and Format

No public API, storage format, or protocol changes.

### Documentation

No documentation changes.

### Generative AI tooling

Generated-by: Codex GPT-5.6 Terra

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -1809,6 +1809,8 @@ endmacro()
 
 macro(build_glog)
     message(STATUS "Building glog from source")
+    find_library(LIBUNWIND_LIBRARY NAMES unwind)
+
     set(GLOG_PREFIX "${CMAKE_CURRENT_BINARY_DIR}/glog_ep-install")
     set(GLOG_INCLUDE_DIR "${GLOG_PREFIX}/include")
     if(${UPPERCASE_BUILD_TYPE} STREQUAL "DEBUG")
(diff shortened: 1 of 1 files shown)


mrdrivingduck:
Additional note from testing: I could not reproduce the unresolved libunwind symbols on macOS, while the issue is reproducible on Linux.

This appears to be due to different linker and loader behavior for undefined symbols. When producing a `.dylib`, macOS ld64 can tolerate the `unw_*` references and defer their resolution until dynamic loading, so both the library build and extension loading may succeed. Linux exposes the missing libunwind dependency earlier during linking.

Therefore, a successful macOS build does not make the configuration consistent. When `LIBUNWIND_LIBRARY=FALSE`, bundled glog should still receive `-DWITH_UNWIND=none` rather than auto-detecting and enabling libunwind.

zjw1111 (approved):
Thanks for the follow-up! The fix is well-reasoned: making both glog's `WITH_UNWIND` setting and paimon's link propagation derive from the same `LIBUNWIND_LIBRARY` result resolves the unresolved `unw_*` symbols cleanly, and `-DWITH_UNWIND=none` is a valid driver value for glog v0.7.1. +1

PR #417, opened 2026-07-14
feat(blob): support blob.split-by-file-size to weigh blob files in scan splitting

### LINKED ISSUE #416: [Feature] Support blob.split-by-file-size to weigh blob files in scan splitting
### Search before asking

### Motivation

Java Paimon provides the `blob.split-by-file-size` option (no default value, see `CoreOptions.BLOB_SPLIT_BY_FILE_SIZE` in apache/paimon) to control whether blob file size is considered as a factor when performing scan splitting. When unset it derives from the negation of `blob-as-descriptor`: content reads count blob file sizes, while descriptor reads (which never fetch blob payloads) weigh a blob file only by the open file cost.

paimon-cpp currently has no such option: the weight function in `DataEvolutionSplitGenerator::SplitForBatch` unconditionally sums `file_size` of every file in a row-id range, including `.blob` files. This matches the Java default for content reads, but there is no way to opt out. Under `blob-as-descriptor` reads, where only ~100-byte descriptors are materialized, the huge blob file sizes still dominate split weights and shatter the scan into many tiny splits, leaving a behavior gap between the two implementations.

### Solution

Align with the Java implementation:

- Add `Options::BLOB_SPLIT_BY_FILE_SIZE` (`"blob.split-by-file-size"`, no default) and parse it into `CoreOptions` together with `blob-as-descriptor`, exposed as `CoreOptions::BlobSplitByFileSize()` which returns the explicit value or falls back to the negation of `blob-as-descriptor` (matching `CoreOptions#blobSplitByFileSize`).
- Add a `count_blob_size` flag to `DataEvolutionSplitGenerator`: the split weight function counts a blob file as its `file_size` when enabled, otherwise as the open file cost, while non-blob files always count as `file_size` (matching `DataEvolutionSplitGenerator#splitForBatch`).
- Pass `CoreOptions::BlobSplitByFileSize()` when constructing `DataEvolutionSplitGenerator` in the data-evolution scan path (matching `AppendOnlyFileStoreTable#splitGenerator`).
- Tests: `CoreOptionsTest` assertions for the default, the value derived from `blob-as-descriptor`, and explicit-value precedence; a `SplitGeneratorTest` case verifying that the same data/blob file layout packs into different splits depending on the flag.

### Anything else?

The option key, default derivation, and weighting semantics follow `CoreOptions.BLOB_SPLIT_BY_FILE_SIZE` and `DataEvolutionSplitGenerator` in apache/paimon (introduced in apache/paimon#7020).

SteNicholas:
### Purpose

Linked issue: close #416

Align with `CoreOptions.BLOB_SPLIT_BY_FILE_SIZE` in apache/paimon. This PR adds the `blob.split-by-file-size` option (no default value) to control whether blob file size is considered as a factor when performing scan splitting. When unset, `CoreOptions::BlobSplitByFileSize()` derives the value from the negation of `blob-as-descriptor`: content reads keep counting blob file sizes, while descriptor reads, which materialize only small descriptors instead of blob payloads, weigh a blob file only by the open file cost. `DataEvolutionSplitGenerator` gains a `count_blob_size` flag wired from the option in the data-evolution scan path, so huge blob files no longer dominate split weights and shatter the scan into many tiny splits when their contents are not read.

### Tests

- UT `CoreOptionsTest.TestDefaultValue`: the option defaults to `true` when neither it nor `blob-as-descriptor` is set.
- UT `CoreOptionsTest.TestFromMap`: with `blob-as-descriptor=true` and the option unset, `CoreOptions::BlobSplitByFileSize()` derives `false`.
- UT `CoreOptionsTest.TestFallback`: an explicit `blob.split-by-file-size=true` wins over the default derived from `blob-as-descriptor=true`.
- UT `SplitGeneratorTest.TestDataEvolutionBlobSplitByFileSize`: the same data/blob file layout packs into two splits when blob file sizes count in the weight, and into a single split when blob files weigh only the open file cost.

### API and Format

Adds the public option constant `Options::BLOB_SPLIT_BY_FILE_SIZE` in `include/paimon/defs.h` and the getter `CoreOptions::BlobSplitByFileSize()`. The `DataEvolutionSplitGenerator` constructor gains a `count_blob_size` parameter (internal, under `src/`). No storage format or protocol change.

### Documentation

The option is documented in the `include/paimon/defs.h` doc comment, following the existing blob options.

### Generative AI tooling

### DIFF
--- a/include/paimon/defs.h
+++ b/include/paimon/defs.h
@@ -119,7 +119,7 @@ struct PAIMON_EXPORT Options {
     static const char BLOB_TARGET_FILE_SIZE[];
 
     /// "blob.split-by-file-size" - Whether to consider blob file size as a factor when performing
-    /// scan splitting. Default is the negation of BLOB_AS_DESCRIPTOR.
+    /// scan splitting. When unset, defaults to the negation of BLOB_AS_DESCRIPTOR.
     static const char BLOB_SPLIT_BY_FILE_SIZE[];
 
     /// "partition.default-name" - The default partition name in case the dynamic partition column
--- a/include/paimon/defs.h
+++ b/include/paimon/defs.h
@@ -118,6 +118,10 @@ struct PAIMON_EXPORT Options {
     /// "blob.target-file-size" - Target size of a blob file. Default is TARGET_FILE_SIZE.
     static const char BLOB_TARGET_FILE_SIZE[];
 
+    /// "blob.split-by-file-size" - Whether to consider blob file size as a factor when performing
+    /// scan splitting. Default is the negation of BLOB_AS_DESCRIPTOR.
+    static const char BLOB_SPLIT_BY_FILE_SIZE[];
+
     /// "partition.default-name" - The default partition name in case the dynamic partition column
     /// value is null/empty string. Default is "__DEFAULT_PARTITION__".
     static const char PARTITION_DEFAULT_NAME[];
--- a/src/paimon/common/defs.cpp
+++ b/src/paimon/common/defs.cpp
@@ -33,6 +33,7 @@ const char Options::FILE_FORMAT[] = "file.format";
 const char Options::FILE_SYSTEM[] = "file-system";
 const char Options::TARGET_FILE_SIZE[] = "target-file-size";
 const char Options::BLOB_TARGET_FILE_SIZE[] = "blob.target-file-size";
+const char Options::BLOB_SPLIT_BY_FILE_SIZE[] = "blob.split-by-file-size";
 const char Options::PAGE_SIZE[] = "page-size";
 const char Options::PARTITION_DEFAULT_NAME[] = "partition.default-name";
 const char Options::FILE_COMPRESSION[] = "file.compression";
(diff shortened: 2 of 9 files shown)


include/paimon/defs.h
```diff
@@ -118,6 +118,10 @@ struct PAIMON_EXPORT Options {
     /// "blob.target-file-size" - Target size of a blob file. Default is TARGET_FILE_SIZE.
     static const char BLOB_TARGET_FILE_SIZE[];
 
+    /// "blob.split-by-file-size" - Whether to consider blob file size as a factor when performing
+    /// scan splitting. Default is the negation of BLOB_AS_DESCRIPTOR.
```
lxy-9602: Just a minor suggestion: on the Java side, `BLOB_SPLIT_BY_FILE_SIZE` uses `noDefaultValue()`, with the actual fallback handled in the getter, and the generated docs show the default as (none). It might be better to update the C++ comment to something like: When unset, defaults to the negation of `BLOB_AS_DESCRIPTOR`.

SteNicholas: Good catch. Updated the doc comment to `When unset, defaults to the negation of BLOB_AS_DESCRIPTOR.` to match the Java side, where `BLOB_SPLIT_BY_FILE_SIZE` uses `noDefaultValue()` and the fallback is handled in `blobSplitByFileSize()`.

lxy-9602 (approved):
+1

PR #421, opened 2026-07-14
feat(blob): support blob-write-null-on-missing-file and blob-write-null-on-fetch-failure options

### LINKED ISSUE #420: [Feature] Support blob-write-null-on-missing-file and blob-write-null-on-fetch-failure options
### Search before asking

### Motivation

Java Paimon supports two table options for descriptor BLOB writes ([CoreOptions](https://github.com/apache/paimon/blob/master/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java)):

- `blob-write-null-on-missing-file`: write NULL for a descriptor BLOB value when the referenced file does not exist (e.g. FileNotFound / HTTP 404) at write time.
- `blob-write-null-on-fetch-failure`: write NULL when the referenced data cannot be fetched for other reasons (invalid descriptor, invalid offset, non-404 errors). A missing file is handled only by `blob-write-null-on-missing-file`.

paimon-cpp has neither the options nor the corresponding logic: when a blob descriptor references a file that does not exist or cannot be fetched, `BlobFormatWriter` fails the whole write. Pipelines that tolerate a small amount of missing upstream blob data cannot express that policy in C++, which diverges from the Java writer behavior in `BlobFormatWriter#addElement`.

### Solution

Align with the Java `BlobFormatWriter` semantics:

- Add the two option keys and thread them from the table options map through `BlobFileFormat` / `BlobWriterBuilder` into `BlobFormatWriter`.
- In `BlobFormatWriter::WriteBlob`, open the descriptor input stream before writing any bytes, so an open-time fetch failure can be converted into a NULL element (`kNullBinLength`) without leaving partial data in the output stream:
  - missing file (`Status::NotExist`) is converted only when `blob-write-null-on-missing-file` is enabled;
  - other open failures are converted only when `blob-write-null-on-fetch-failure` is enabled;
  - failures during the streaming copy keep failing the write, matching Java.
- Classification relies on `Status::NotExist`, but `JindoFileSystem::Open` currently maps every jindo error (including file-not-found) to `Status::IOError`, so on OSS the two options would gate the wrong cases. Map `JDO_FILE_NOT_FOUND_ERROR` to `Status::NotExist` at the jindo boundary (`Open`, and `JindoInputStream::Length` for lazily-surfaced not-found), consistent with `LocalFileSystem::Open`.

### Anything else?

_No response_

SteNicholas:
### Purpose

Linked issue: close #420

Align with the Java `BlobFormatWriter` semantics for descriptor BLOB writes:

- Add the `blob-write-null-on-missing-file` / `blob-write-null-on-fetch-failure` option keys and thread them from the table options map through `BlobFileFormat` / `BlobWriterBuilder` into `BlobFormatWriter`.
- `BlobFormatWriter::WriteBlob` now opens the descriptor input stream before writing any bytes, so an open-time fetch failure can be converted into a NULL element (`kNullBinLength`) without leaving partial data in the output stream. A missing file (`Status::NotExist`) is converted only by `blob-write-null-on-missing-file`; other open failures only by `blob-write-null-on-fetch-failure`; failures during the streaming copy keep failing the write, matching Java.
- Map `JDO_FILE_NOT_FOUND_ERROR` to `Status::NotExist` at the jindo boundary (`JindoFileSystem::Open`, and `JindoInputStream::Length` for lazily-surfaced not-found) so the missing-file classification also holds on OSS, consistent with `LocalFileSystem::Open`. Note: this makes a missing consumer file on OSS take the existing `IsNotExist` branch in `ConsumerManager::GetNextSnapshotId` (returning "no next snapshot" like the local file system and Java) instead of surfacing an `IOError`.

### Tests

- `BlobFormatWriterWriteNullTest.TestWriteNullOnMissingFile` / `TestWriteNullOnFetchFailure`: each option converts only its own failure class to NULL (with read-back verification), the other class still fails, and a rejected row leaves the writer usable.
- `BlobFormatWriterWriteNullTest.TestWriteNullOnBothOptionsEnabled`: both options together produce a mixed NULL/NULL/data file, verified by read-back.
- `BlobFormatWriterWriteNullTest.TestWriteNullClassifiesByStatusCode`: classification contract pinned on the open status code via a mock file system (`NotExist`/`IOError` × both option combinations), independent of the file system implementation.
- `BlobWriterBuilderTest.TestWriteNullOptions`: option parsing and wiring (absent keys default to false, `"true"`/`"false"` values, invalid value fails `Build`).
- `BlobFileFormatFactoryTest.TestWriteNullOptionPropagation`: end-to-end option propagation through `FileFormatFactory::Get("blob", options)`.
- `JindoUtilsTest`: unit tests for the jindo status conversion (`JDO_FILE_NOT_FOUND_ERROR` → `NotExist`, other codes → `IOError`, ok pass-through, and the plain macro keeping `IOError`); registered commented-out in `fs_test` alongside the other jindo tests since jindo is not built by default.

Local validation: `paimon-blob-format-test` 39/39 passed; `paimon-fs-test` matches the HEAD baseline. The jindo sources compile only with `-DPAIMON_ENABLE_JINDO=ON` (CI build entry).

### API and Format

- Two new public option constants on the already-exported `Options` struct in `include/paimon/defs.h` (`BLOB_WRITE_NULL_ON_MISSING_FILE`, `BLOB_WRITE_NULL_ON_FETCH_FAILURE`), both defaulting to false, so behavior is unchanged unless opted in.
- No storage format change: NULL elements reuse the existing `kNullBinLength` encoding that readers already understand.

### Documentation

Option semantics are documented on the new constants in `include/paimon/defs.h`.

### Generative AI tooling


src/paimon/fs/jindo/jindo_utils.h
```diff
@@ -28,4 +28,18 @@ namespace paimon::jindo {
         }                                             \
     } while (false)
 
+/// Like PAIMON_RETURN_NOT_OK_FROM_JINDO, but maps a jindo file-not-found error to
+/// Status::NotExist so that callers can distinguish a missing file from other IO errors,
+/// consistent with LocalFileSystem::Open.
```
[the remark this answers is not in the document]
SteNicholas: The macro was self-contained in practice: `jindo_utils.h` includes `JdoStatus.hpp`, which itself includes `jdo_error.h`, so `JDO_FILE_NOT_FOUND_ERROR` was always visible to any TU including this header. That said, relying on a transitive include from an SDK header is fragile, and `jindo_file_system.cpp` already includes `jdo_error.h` directly, so the direct include is the right hygiene. Addressed in 0871da3.

src/paimon/format/blob/blob_writer_builder_test.cpp
```diff
@@ -53,4 +57,81 @@ TEST_F(BlobWriterBuilderTest, TestSimple) {
...
+                                                                                 GetDefaultPool()));
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::ExportArray(*array, c_array));
+        return Status::OK();
+    };
+    auto build_writer = [&](const std::map<std::string, std::string>& options)
+        -> Result<std::unique_ptr<FormatWriter>> {
+        BlobWriterBuilder builder(struct_type_, options);
+        builder.WithFileSystem(file_system_);
+        return builder.Build(output_stream_, "none");
+    };
```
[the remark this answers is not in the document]
SteNicholas: There was no correctness issue in the previous shape — a rejected `AddBatch` returns before any byte is written (the stream open happens before the magic number since this PR), and no scenario reads the shared file back — but agreed that per-scenario isolation is more robust and keeps the test from masking a future writer-side stream-position check. Each `build_writer()` call now creates its own output file. Addressed in 0871da3.

src/paimon/format/blob/blob_format_writer.cpp
```diff
@@ -31,22 +31,33 @@
...
 
 BlobFormatWriter::BlobFormatWriter(const std::shared_ptr<OutputStream>& out, const std::string& uri,
                                    const std::shared_ptr<arrow::DataType>& data_type,
                                    const std::shared_ptr<FileSystem>& fs,
-                                   const std::shared_ptr<MemoryPool>& pool)
-    : out_(out), uri_(uri), data_type_(data_type), fs_(fs), pool_(pool) {
+                                   const std::shared_ptr<MemoryPool>& pool,
+                                   bool write_null_on_missing_file,
+                                   bool write_null_on_fetch_failure)
+    : out_(out),
```
lxy-9602, line 44: Please move the two parameters `write_null_on_missing_file` and `write_null_on_fetch_failure` before `pool` and `fs`.

src/paimon/CMakeLists.txt
```diff
@@ -773,6 +773,7 @@ if(PAIMON_BUILD_TESTS)
                     fs/local/local_file_test.cpp
                     # fs/jindo/jindo_file_system_factory_test.cpp
                     # fs/jindo/jindo_file_system_test.cpp
+                    # fs/jindo/jindo_utils_test.cpp
                     STATIC_LINK_LIBS
```
lxy-9602, line 810: It seems `fs/jindo/jindo_utils_test.cpp` could probably be enabled, since it only verifies in-memory state and does not actually access the OSS cluster, unlike the other tests.

src/paimon/fs/jindo/jindo_utils.h
```diff
@@ -28,4 +29,18 @@ namespace paimon::jindo {
...
+#define PAIMON_RETURN_NOT_OK_FROM_JINDO_WITH_NOT_EXIST(JINDO_STATUS) \
+    do {                                                             \
+        auto __s = (JINDO_STATUS);                                   \
+        if (PAIMON_UNLIKELY(!(__s).ok())) {                          \
+            if ((__s).getErrCode() == JDO_FILE_NOT_FOUND_ERROR) {    \
+                return Status::NotExist(__s.errMsg());               \
```
lxy-9602: Would it make sense to handle `JDO_FILE_NOT_FOUND_ERROR` directly like `PAIMON_RETURN_NOT_OK_FROM_JINDO`, so that we don’t need a separate `PAIMON_RETURN_NOT_OK_FROM_JINDO_WITH_NOT_EXIST` variant?

src/paimon/format/blob/blob_format_writer_test.cpp
```diff
@@ -394,6 +444,155 @@ TEST_P(BlobFormatWriterTest, TestAddBatchWithNullValues) {
...
+    }
+    {
+        ASSERT_OK_AND_ASSIGN(
+            std::shared_ptr<BlobFormatWriter> writer,
+            BlobFormatWriter::Create(output_stream_, struct_type_, not_exist_fs, pool_,
+                                     /*write_null_on_missing_file=*/false,
+                                     /*write_null_on_fetch_failure=*/true));
+        ASSERT_NOK_WITH_MSG(AddBatchOnce(writer, array), "mock not exist");
+    }
+}
```
lxy-9602, line 586: These tests already look very comprehensive. 👍 I wonder if we can add end-to-end write/read tests in `blob_table_inte_test.cpp`, including the merge with data files, so we can avoid missing any framework-level issues. The existing utils in `blob_table_inte_test.cpp` also provide some useful helpers.

src/paimon/format/blob/blob_format_writer.cpp
SteNicholas: Done in da5454a — moved the two parameters before `fs`/`pool` in both `Create()` and the private constructor, matching the Java `BlobFormatWriter` constructor order (the flags right after the type, infrastructure parameters last). All call sites are updated; the fixture now provides a `CreateDefaultWriter()` helper so the tests that don't care about the options avoid the repeated `false, false` arguments.

src/paimon/CMakeLists.txt
SteNicholas: You're right that it only verifies in-memory status conversion and never touches an OSS cluster. It can't be enabled unconditionally though: `jindo_utils.h` includes the jindo SDK headers (`JdoStatus.hpp` / `jdo_error.h`), which only exist when the SDK is fetched under `-DPAIMON_ENABLE_JINDO=ON`, so a default build would fail to compile `fs_test`. In da5454a the test is appended to `fs_test` under `if(PAIMON_ENABLE_JINDO)` (together with the jindo link libs and SDK include dir), which means it builds and actually runs in CI since `ci/scripts/build_paimon.sh` enables jindo. Verified locally with `-DPAIMON_ENABLE_JINDO=ON`: the test compiles, links and passes. The two tests that need real OSS access stay commented out.

src/paimon/fs/jindo/jindo_utils.h
SteNicholas: Agreed — done in da5454a. `PAIMON_RETURN_NOT_OK_FROM_JINDO` now maps `JDO_FILE_NOT_FOUND_ERROR` to `Status::NotExist` directly and the `_WITH_NOT_EXIST` variant is removed. Besides being simpler, this is actually more correct: it makes the remaining jindo call sites (e.g. `GetFileStatus`, which returns `NotExist` for a missing file on the local file system but returned `IOError` on jindo) consistent with `LocalFileSystem`, and matches Java where `FileIO` surfaces `FileNotFoundException` uniformly for every operation rather than per call site. I checked all callers — no production code branches on `IsIOError()`, so the broader mapping doesn't change any existing error-handling path.

src/paimon/format/blob/blob_format_writer_test.cpp
SteNicholas: Good idea — added three end-to-end tests in `blob_table_inte_test.cpp` (da5454a), reusing the existing descriptor helpers plus a new `DeleteDescriptorTarget` helper:

- `TestWriteNullOnMissingFile`: with the option enabled, a descriptor whose file is deleted before the write becomes a NULL element; `VerifyDataFileMetas` pins the layout (main data file with `write_cols={f0,f1}` + .blob file with `write_cols={blob}`, both keeping the full row count), and the read path merges them back into rows where that blob is NULL while the other rows/columns stay intact.
- `TestMissingFileFailsWriteWhenWriteNullDisabled`: with the options off, the write fails as before.
- `TestWriteNullOnFetchFailureKeepsMissingFileFailing`: enabling only `blob-write-null-on-fetch-failure` does not convert a missing file, pinning the option independence at the table level.

One note: the fetch-failure→NULL conversion itself can't be triggered end-to-end with the local file system (the Java ITCase uses an HTTP mock server for this); that classification stays covered by the mock-file-system unit test `TestWriteNullClassifiesByStatusCode`. All three tests pass for parquet/orc/avro.

test/inte/blob_table_inte_test.cpp
```diff
@@ -561,6 +576,131 @@ TEST_P(BlobTableInteTest, TestAppendTableWriteWithBlobAsDescriptorFalse) {
...
+    auto raw_array = std::dynamic_pointer_cast<arrow::StructArray>(
+        arrow::ipc::internal::json::ArrayFromJSON(arrow::struct_(fields), raw_json).ValueOrDie());
+    ASSERT_OK_AND_ASSIGN(auto desc_array, ConvertRawBlobToDescriptor(raw_array, {"blob"}));
+    ASSERT_OK(DeleteDescriptorTarget(desc_array, "blob", /*row=*/1));
+
+    auto schema = arrow::schema(fields);
+    ASSERT_NOK_WITH_MSG(WriteArray(table_path, {}, schema->field_names(), {desc_array}),
+                        "not exists");
+}
+
```
lxy-9602, line 699: Would it be possible to add one more test for `blob-write-null-on-fetch-failure = true`, covering a non-missing-file error such as a bad offset, to confirm that the data is read/written as null as expected?

SteNicholas: Done in 723e248 — added `TestWriteNullOnFetchFailure` in `blob_table_inte_test.cpp`. It rewrites row 1's descriptor to reference an offset beyond the end of the target file (via a new `CorruptDescriptorOffset` helper reusing the `TransformBlobFields` framework), i.e. a fetch failure that is not a missing file. With `blob-write-null-on-fetch-failure = true`, the write succeeds and converts that element to NULL: both the data file and the `.blob` file keep the full 3-row count, and the read path merges them back so row 1 comes back with a NULL blob while the other rows keep their data.

lxy-9602 (dismissed):
+1

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -1636,13 +1636,16 @@ macro(build_arrow)
...
+    # after libarrow.a on linkers that resolve symbols strictly left-to-right.
     target_link_libraries(arrow
                           INTERFACE zstd
                                     snappy
                                     lz4
                                     zlib
                                     re2::re2
-                                    arrow_bundled_dependencies)
+                                    arrow_bundled_dependencies
+                                    ${CMAKE_DL_LIBS})
```
zjw1111, line 1648: Thanks for fixing the link order here by moving `${CMAKE_DL_LIBS}` into arrow's INTERFACE — this is the right direction, since `-ldl` is really an implementation requirement of the libraries that call `dlopen`/`dlsym`/`dladdr` (arrow here; jindosdk already declares it PUBLIC).

As a possible follow-up (non-blocking for this PR): I wonder if it would be cleaner to keep declaring dl only as a usage requirement on each real provider like this, and gradually drop the manual `${CMAKE_DL_LIBS}` that is currently scattered across the individual `add_paimon_lib` targets — those targets don't use dl themselves (a grep for `dlopen`/`dlsym`/`dladdr` under `src/` and `include/` finds nothing), so their dl entries look like they were added on behalf of arrow. Declaring it on the provider gives a single source of truth and lets CMake place `-ldl` after each provider automatically, which is more robust than hand-placing it in the consumers.

SteNicholas: @zjw1111, I would like to create follow-up pull request for this.

lxy-9602 (approved):
+1

PR #429, opened 2026-07-16
fix(blob): support blob-only writes and reject nested blob fields

lxy-9602:
### Purpose

Linked issue: #283

This PR improves BLOB schema validation and data-evolution partial writes.

- Reject BLOB fields nested in ARRAY, MAP, or STRUCT types because C++ currently supports only top-level BLOB fields.
- Explicitly reject `ARRAY<BLOB>`, which is supported by Java Paimon but not yet supported by Paimon C++.
- Allow data-evolution writes whose write schema contains only BLOB fields.
- Keep the existing row-tracking requirement that a BLOB-only partial write must provide a matching `first_row_id`.

### Tests
ArrowSchemaValidatorTest.TestBlobFieldMustBeTopLevel
BlobUtilsTest.SeparateBlobArray
AppendOnlyWriterTest.TestWriteWithOnlyBlobField
BlobTableInteTest.TestDataEvolutionBlobOnlyWriteWithFirstRowId
BlobTableInteTest.TestDataEvolutionBlobOnlyFirstCommitFailsWithoutFirstRowId

### Generative AI tooling


lszskye (approved):
+1

PR #432, opened 2026-07-17
feat(commit): support TruncateTable/Abort/RollbackToAsLatest and add ut

lucasfang:
### Purpose

Linked issue: close #xxx

This change aligns the C++ `FileStoreCommit` with Java Paimon by adding three missing commit operations, plus unit-test coverage for the commit path.

- **`TruncateTable(commit_identifier)`**: truncates the whole table by overwriting all partitions with no new files. It reuses the existing `TryOverwrite` path with an empty partition list (which matches all partitions), producing an `OVERWRITE` snapshot that deletes every existing file. Mirrors Java `FileStoreCommit.truncateTable`.
- **`Abort(commit_messages)`**: cleans up after an unsuccessful commit by best-effort deleting the data and index files referenced by the given commit messages (new files, changelog files, compact-after files, and new index files from both the data and compact increments). Delete failures are ignored, matching Java `deleteQuietly`. A non-`CommitMessageImpl` message is rejected with an error. Mirrors Java `FileStoreCommit.abort`.
- **`RollbackToAsLatest(target_snapshot_id)`**: rolls the table back to a target snapshot and materializes it as a new latest snapshot. It reads the surviving `ADD` data files of both the current latest snapshot and the target snapshot, computes their difference (files only in latest become `DELETE`, files only in target become `ADD`), and writes the base and delta manifests. It then commits a single `OVERWRITE` snapshot whose visible state equals the target: the new snapshot inherits the target's schema id, index manifest, total record count, watermark, statistics, and properties, and its `nextRowId` is the larger of the latest and target `nextRowId` to keep row-id allocation monotonic. Mirrors Java `FileStoreCommit.rollbackToAsLatest`.

Unlike Java, the C++ commit path has no commit-callback mechanism, so for `RollbackToAsLatest` the Java pre/post callbacks and `rollbackIndexChanges` are intentionally omitted; the new snapshot's index manifest points directly to the target's, which is equivalent for the on-disk result. Two private helpers are added to `FileStoreCommitImpl`: `ReadAddManifestEntries` (mirrors `FileEntry.mergeEntries` followed by keeping only `ADD` entries) and `MaxNextRowId` (mirrors Java `maxNextRowId`).

The change also adds unit tests for `FileStoreCommitImpl` and introduces a dedicated `sequence_snapshot_properties_test.cpp` suite.

### Tests

Added GoogleTest cases (run via `paimon-core-test`):

- `FileStoreCommitImplTest.TestTruncateTable` — truncates a committed table and asserts a new `OVERWRITE` snapshot whose delta manifest deletes all files (0 added, 3 deleted for the `append_09` fixture).
- `FileStoreCommitImplTest.TestAbortDeletesDataAndIndexFiles` — materializes the data/index files referenced by a commit message and verifies `Abort` removes them.
- `FileStoreCommitImplTest.AbortIgnoresMissingFilesAndFailsForNonImplMessage` — aborting missing files is a no-op; a non-impl message returns an error.
- `FileStoreCommitImplTest.TestRollbackToAsLatest` — appends three snapshots, then rolls back to snapshot 1 (exercising the `DELETE`-only delta branch) and rolls forward to snapshot 3 (exercising the `ADD`-only delta branch). Asserts a new `OVERWRITE` snapshot with the target's schema id, total record count, index manifest, `nextRowId == max(latest, target)`, and a surviving file set equal to the target.
- `FileStoreCommitImplTest.TestRollbackToAsLatestDeletionVectorOnlyChange` — when the target and latest share identical data files and differ only in the deletion vector (index) manifest, the rollback produces an empty data delta (`DeltaRecordCount == 0`) and the new snapshot inherits the target's index manifest, dropping the deletion vector.
- `FileStoreCommitImplTest.TestRollbackToAsLatestNoLatestSnapshotReturnsError` — rolling back an empty table returns an error.
- `FileStoreCommitImplTest.TestRollbackToAsLatestTargetNotExistReturnsError` — rolling back to a non-existent target snapshot returns an error.
- Additional `FileStoreCommitImplTest` coverage (drop-partition validation, filter-and-commit, overwrite ownership/upgrade, conflict checks, etc.).
- New `SequenceSnapshotPropertiesTest` suite covering max-sequence-number parsing, extraction from files, and merge behavior.

### API and Format

Yes — public API additions. Three new pure-virtual methods on `paimon::FileStoreCommit` (`include/paimon/file_store_commit.h`):

- `virtual Status TruncateTable(int64_t commit_identifier) = 0;`
- `virtual Status Abort(const std::vector<std::shared_ptr<CommitMessage>>& commit_messages) = 0;`
- `virtual Result<bool> RollbackToAsLatest(int64_t target_snapshot_id) = 0;`

No storage format or protocol change. `TruncateTable` and `RollbackToAsLatest` produce standard `OVERWRITE` snapshots; there is no on-disk layout change.

### Documentation

New public API methods, documented via doxygen comments in the header. No new user-guide page is required.

### Generative AI tooling

Generated-by: Qoder


include/paimon/file_store_commit.h
```diff
@@ -135,6 +135,30 @@ class PAIMON_EXPORT FileStoreCommit {
...
+
+    /// Roll back to the target snapshot and materialize it as the latest snapshot.
+    ///
+    /// Reads the surviving files of both the current latest snapshot and the target
+    /// snapshot, then commits an OVERWRITE snapshot whose visible state equals the target.
+    ///
+    /// @param target_snapshot_id The snapshot id to roll back to.
+    /// @return Result<bool>; true if the atomic commit succeeded. Returns an error status if
+    ///     there is no latest snapshot or the target snapshot does not exist.
+    virtual Result<bool> RollbackToAsLatest(int64_t target_snapshot_id) = 0;
```
zjw1111, line 160: Scope mismatch: this PR also adds a third public pure-virtual API `RollbackToAsLatest` (plus its impl and the `ReadAddManifestEntries`/`MaxNextRowId` helpers and 4 tests), but the title and Purpose only mention `TruncateTable` and `Abort`. Since `FileStoreCommit` is `PAIMON_EXPORT`, adding pure-virtual methods affects ABI, and reviewers reading the Purpose won't expect this non-trivial rollback logic. Please either update the title/Purpose to cover `RollbackToAsLatest`, or split it into its own PR. The implementation itself looks correct and faithful to Java.

zjw1111 (commented):
Reviewed the three new commit operations against the Java `FileStoreCommitImpl`. No blocking correctness issues; the implementations track Java line-by-line and the tests are behavior-level (write files → operate → assert file/snapshot state).

Verified as correct:
- `TruncateTable`: empty `partitions={}` maps to "all partitions" in both the data scan (`FileStoreScan::CreatePartitionPredicate`) and the index filter (`CommitScanner::ReadAllIndexEntriesFromPartitions`), equivalent to Java `tryOverwritePartition(null, ...)`.
- `Abort`: delete set = `newFiles + changelogFiles + compactAfter + compact.changelogFiles` (data) and `newIndexFiles` from both increments; correctly excludes `compactBefore`. Matches Java exactly.
- `RollbackToAsLatest`: `Snapshot` ctor argument order matches the regular commit path; `total_record_count`/`index_manifest`/`watermark`/`statistics`/`properties` from target, `delta_record_count` from the diff, `commit_identifier=int64_max`, `next_row_id=max(latest,target)` — all consistent with Java. Skipping pre/post-commit callbacks is consistent with the C++ path not implementing callbacks yet.

One inline comment on scope. Test blind spots (non-blocking):
- `RollbackToAsLatest`: no coverage for the `success=false` (concurrent-conflict) branch and the temp-manifest-list leak on that path (leak matches Java).
- `Abort`: no coverage for a delete that fails on IO/permission and is silently ignored (only the missing-file no-op is covered).
- `TruncateTable`: only single-truncate covered; no empty-table case.

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -202,6 +202,184 @@ Status FileStoreCommitImpl::DropPartition(
...
+    // For row-tracking tables nextRowId must stay monotonic: a rollback to an older snapshot must
+    // not move it backwards, otherwise new appends would reuse row ids already assigned by the
+    // snapshots between the target and the previous latest, breaking the global uniqueness of
+    // _ROW_ID. Keep the larger of the previous latest and the target nextRowId.
+    std::optional<int64_t> next_row_id =
+        MaxNextRowId(latest.NextRowId(), target_snapshot.NextRowId());
+
+    int64_t delta_record_count =
+        ManifestEntry::RecordCountAdd(delta_files) - ManifestEntry::RecordCountDelete(delta_files);
+    Snapshot new_snapshot(
```
zjw1111, line 340: add TODO for changelog?

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -202,6 +202,184 @@ Status FileStoreCommitImpl::DropPartition(
...
+    std::vector<ManifestEntry> add_entries;
+    for (auto& entry : merged_entries) {
+        if (entry.Kind() == FileKind::Add()) {
+            add_entries.push_back(std::move(entry));
+        }
+    }
+    return add_entries;
+}
+
+std::optional<int64_t> FileStoreCommitImpl::MaxNextRowId(const std::optional<int64_t>& left,
```
zjw1111: use std::max

src/paimon/core/operation/commit/sequence_snapshot_properties_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+    ManifestEntry CreateEntry(const FileKind& kind, int64_t max_sequence_number) const {
+        return ManifestEntry(kind, BinaryRow(0), /*bucket=*/0, /*total_buckets=*/1,
+                             CreateDataFileMeta(max_sequence_number));
+    }
+};
+
+TEST_F(SequenceSnapshotPropertiesTest, MaxSequenceNumberEmptySnapshot) {
+    ASSERT_OK_AND_ASSIGN(std::optional<int64_t> result,
+                         SequenceSnapshotProperties::MaxSequenceNumber(std::nullopt));
+    EXPECT_FALSE(result.has_value());
```
zjw1111: ASSERT

src/paimon/core/operation/file_store_commit_impl.cpp
lucasfang: No need, this is aligned with Java implementation.

src/paimon/core/operation/file_store_commit_impl_test.cpp
```diff
@@ -763,6 +806,225 @@ TEST_F(FileStoreCommitImplTest, TestCommitMultipleTimes) {
...
+    ASSERT_OK_AND_ASSIGN(bool snapshot4_exist, file_system_->Exists(PathUtil::JoinPath(
+                                                   table_path_, "snapshot/snapshot-4")));
+    ASSERT_TRUE(snapshot4_exist);
+    ASSERT_OK_AND_ASSIGN(Snapshot snapshot4, commit_impl->snapshot_manager_->LoadSnapshot(4));
+    ASSERT_EQ(snapshot4.Id(), 4);
+    ASSERT_TRUE(snapshot4.GetCommitKind() == Snapshot::CommitKind::Overwrite());
+    ASSERT_EQ(snapshot4.SchemaId(), snapshot1.SchemaId());
+    ASSERT_EQ(snapshot4.TotalRecordCount(), snapshot1.TotalRecordCount());
+    ASSERT_EQ(snapshot4.NextRowId(),
+              FileStoreCommitImpl::MaxNextRowId(snapshot3.NextRowId(), snapshot1.NextRowId()));
```
zjw1111: Blocking compile error: `MaxNextRowId` was removed from production code (replaced by `std::max` at file_store_commit_impl.cpp:336), but this line still calls `FileStoreCommitImpl::MaxNextRowId`, so `paimon-core-test` will not build. Replace it with `std::max(snapshot3.NextRowId(), snapshot1.NextRowId())`, or keep the helper.

src/paimon/core/operation/commit/sequence_snapshot_properties_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+    ASSERT_FALSE(result.has_value());
+}
+
+TEST_F(SequenceSnapshotPropertiesTest, MaxSequenceNumberValid) {
+    std::map<std::string, std::string> properties{
+        {SequenceSnapshotProperties::kMaxSequenceNumberKey, "123"}};
+    ASSERT_OK_AND_ASSIGN(std::optional<int64_t> result,
+                         SequenceSnapshotProperties::MaxSequenceNumber(MakeSnapshot(properties)));
+    ASSERT_TRUE(result.has_value());
+    EXPECT_EQ(123, result.value());
```
zjw1111: These `EXPECT_*` are inside void `TEST_F` bodies (the helpers in this file return values and hold no assertions), so per our convention they should be `ASSERT_*`. The earlier change only converted the `has_value()` guards; please also convert the remaining ones: L108, L126, L131, L140, L147, L148, L154, L161, L168, L173.

zjw1111 (commented):
Two follow-ups after the latest fixes.

zjw1111:
Minor: the PR description has a lot of hard line breaks in the middle of sentences. Please drop them and let the text wrap naturally — the forced wrapping reads awkwardly in the rendered view.

zjw1111 (approved):
+1

Issue #357, opened 2026-06-10
[Feature] Support S3-compatible object storage (S3/MinIO) filesystem


Pandas886:
## Description

Add support for S3-compatible object storage as a filesystem backend in paimon-cpp, including:

- **AWS S3** — native S3 access
- **MinIO** — S3-compatible self-hosted object storage
- Other S3-compatible services (e.g. Alibaba OSS, Ceph RGW, etc.)

## Motivation

Many data lakehouse deployments use object storage (S3, MinIO, OSS) as the underlying storage layer. Currently paimon-cpp appears to support only local filesystem, with HDFS support also in progress (#33). S3/MinIO support would significantly broaden the adoption scenarios for paimon-cpp, especially:

1. Cloud-native deployments on AWS / Alibaba Cloud / other cloud providers
2. On-premise environments using MinIO as the storage backend
3. Hybrid setups where data lives on object storage

## Possible Implementation

- Implement a `S3FileSystem` class conforming to the existing `FileSystem` interface
- Use the AWS C++ SDK (`aws-sdk-cpp`) with S3 module, or a lightweight alternative like `aws-c-s3`
- Support configurable endpoint, region, access key, secret key, and path-style access (for MinIO compatibility)
- The Java version of Paimon already has S3 filesystem support, which can serve as a reference

## Additional Context

Related: #33 (HDFS filesystem support) — once the filesystem abstraction is mature, adding new backends like S3 should be straightforward.

lxy-9602:
Hi @Pandas886, thanks for raising this issue, and sorry for the delayed response.

For Aliyun OSS related scenarios, paimon-cpp currently supports access through Jindo. However, native support for S3-compatible filesystem backends such as AWS S3 and MinIO would still be valuable for broader object storage deployments.

Would you be interested in helping implement an S3 filesystem backend and contributing it to the community? We would be happy to discuss the design and review the contribution.

LinMingQiang:
Hi @Pandas886 , this is an exciting issue!  Could you also implement the HDFS ?  #33  It seems no one is working on it yet.

Pandas886:
> Hi [@Pandas886](https://github.com/Pandas886) , this is an exciting issue! Could you also implement the HDFS ? [#33](https://github.com/alibaba/paimon-cpp/issues/33) It seems no one is working on it yet.

I'm interested, but cpp is not a specialty for me

PR #433, opened 2026-07-19
feat(fs): add lightweight object-store and S3 filesystem support

mrdrivingduck:
Linked: #357

## Background

I started this work because I wanted duckdb-paimon to query Paimon data lakes stored on S3. That required adding S3 filesystem support to paimon-cpp first.

My first implementation used `Aws::S3::S3Client`. It worked, but the dependency cost was much larger than I expected. Reading a Paimon table only needs a fairly small part of S3: `HEAD`, paginated `LIST`, and ranged `GET`. Pulling in the complete AWS S3 client and its CRT dependency stack felt disproportionate for that job, and it also made the build integration considerably more complicated.

I then looked at how DuckDB's [`httpfs`](https://duckdb.org/docs/stable/core_extensions/httpfs/overview) and [Iceberg](https://duckdb.org/docs/current/core_extensions/iceberg/overview) extensions divide this work. DuckDB keeps the HTTP data path generic and limits the storage-specific code to things such as credentials, request signing, URL construction, and error handling. That seemed like a better direction for paimon-cpp as well.

Based on that, I rewrote the original S3 implementation around a reusable object-store filesystem layer and a small HTTP client abstraction.

## Design

The implementation is still being cleaned up, and some names and file boundaries may change before it is ready to merge. But this is the architectural split I expect to keep. I am sharing it now mainly to discuss whether this direction makes sense.

The implementation is organized as follows:

```text
FileSystem
└── ObjectStoreFileSystem
    ├── S3FileSystem
    └── [OSSFileSystem] (WIP)

Shared components
├── ObjectStoreClient
├── HttpClient
├── ObjectStoreInputStream
└── Read-ahead
```

`ObjectStoreFileSystem`, `ObjectStoreClient`, and `ObjectStoreInputStream` implement the filesystem semantics and reading behavior shared by object stores. `S3FileSystem` and `S3ObjectStoreClient` only handle S3-specific URLs, API requests, credentials, signing, and errors.

Object data is transferred by the libcurl-based `HttpClient`. The AWS C libraries are used only for the credential chain and SigV4 signing, so the implementation does not depend on the complete AWS S3 SDK. All of these interfaces remain internal to paimon-cpp, and the existing public `FileSystem` interface is unchanged.

## OSS experiment

To validate the abstraction against a second provider, I also built a WIP OSS filesystem using [Alibaba Cloud OSS C++ SDK v2](https://github.com/aliyun/alibabacloud-oss-cpp-sdk-v2). It reuses the common filesystem semantics, input stream, range reads, read-ahead, pagination, and error handling; the OSS-specific code mainly adapts credentials, endpoints, requests, and responses.

The OSS implementation is not part of this PR, but it could later replace or coexist with the current Jindo-based path. Jindo is a closed-source binary dependency and is no longer recommended in a number of environments, which makes it less suitable as the project's long-term OSS foundation than the open-source OSS SDK.

## Scope

The S3 filesystem in this PR is read-only. It implements the operations needed to read Paimon metadata, manifests, and data files. Write and mutation operations return an explicit unsupported error.

This PR does not add a general-purpose `http://` filesystem, and it does not attempt to cover HDFS. HDFS has different filesystem semantics and should remain a separate implementation if it is added later.

## Testing

Unit tests cover the shared object-store behavior, reads and read-ahead, error handling, and S3 configuration and signing. I also tested metadata, manifest, and ORC reads against a real S3 Paimon warehouse and verified the integration through duckdb-paimon.


src/paimon/fs/s3/s3_file_system_test.cpp
```diff
@@ -0,0 +1,326 @@
...
+    ScopedEnvironmentVariable(const char* name, std::optional<std::string> value) : name_(name) {
+        const char* previous = std::getenv(name);
+        if (previous != nullptr) {
+            previous_ = previous;
+        }
+        if (value) {
+            setenv(name, value->c_str(), 1);
+        } else {
+            unsetenv(name);
+        }
```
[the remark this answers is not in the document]
mrdrivingduck: Windows is not currently a supported paimon-cpp platform. Verified builds currently cover Linux x86_64 and macOS arm64, so this test intentionally remains POSIX-only.

lucasfang (commented):
Each path under src/paimon/fs should be an independent plugin. I suggest moving object_store to the framework side under src/paimon/common/fs/, so that future OSS FileSystem integrations can also reuse it.

mrdrivingduck:
> Each path under src/paimon/fs should be an independent plugin. I suggest moving object_store to the framework side under src/paimon/common/fs/, so that future OSS FileSystem integrations can also reuse it.

Agreed, done.

src/paimon/common/fs/http_client.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include <vector>
+
+#include "fmt/format.h"
+#include "paimon/common/utils/string_utils.h"
+
+namespace paimon::object_store {
+
+namespace {
+
+constexpr int kMaxAttempts = 3;
```
lucasfang: Please use `int32_t` instead of `int`

src/paimon/common/fs/http_client.cpp
```diff
@@ -0,0 +1,200 @@
...
+    mutable std::mutex mutex_;
+    mutable std::vector<CURL*> handles_;
+};
+
+CurlHttpClient::CurlHttpClient() : impl_(std::make_unique<Impl>()) {}
+CurlHttpClient::~CurlHttpClient() = default;
+
+Result<HttpResponse> CurlHttpClient::Execute(const HttpRequest& request,
+                                             const HttpBodyConsumer& consumer) const {
+    for (int attempt = 0; attempt < kMaxAttempts; ++attempt) {
```
lucasfang: Please use `int32_t` instead of `int`

src/paimon/common/fs/object_store_file_system.cpp
```diff
@@ -0,0 +1,532 @@
...
+#include <cassert>
+#include <cstring>
+#include <limits>
+#include <utility>
+
+#include "fmt/format.h"
+#include "paimon/common/utils/math.h"
+#include "paimon/common/utils/path_util.h"
+
+namespace paimon::object_store {
```
lucasfang: Please use a single-level namespace

lucasfang:
+1

src/paimon/common/fs/http_client.cpp
mrdrivingduck: done

src/paimon/common/fs/http_client.cpp
mrdrivingduck: done

src/paimon/common/fs/object_store_file_system.cpp
mrdrivingduck: done

mrdrivingduck:
> Please use int32_t instead of int

Might be good to have that in AGENTS.md 🤔

lucasfang:
> > Please use int32_t instead of int
> 
> Might be good to have that in AGENTS.md 🤔

There is [docs/code-style.md](https://github.com/alibaba/paimon-cpp/blob/main/docs/code-style.md) in AGENTS.md, seems that Agent miss that？

mrdrivingduck:
> There is [docs/code-style.md](https://github.com/alibaba/paimon-cpp/blob/main/docs/code-style.md) in AGENTS.md, seems that Agent miss that？

Yes, I see it now. Never mind.

cmake_modules/BuildAwsAuth.cmake
```diff
@@ -0,0 +1,187 @@
...
+    else()
+        set(S2N_URL
+            "${THIRDPARTY_MIRROR_URL}https://github.com/aws/s2n-tls/archive/refs/tags/${PAIMON_AWS_S2N_BUILD_VERSION}.zip"
+        )
+    endif()
+    externalproject_add(s2n_ep
+                        URL ${S2N_URL}
+                        URL_HASH SHA256=${PAIMON_AWS_S2N_BUILD_SHA256_CHECKSUM}
+                        DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+                        CMAKE_ARGS ${EP_COMMON_TOOLCHAIN}
```
zjw1111: The AWS external projects start from `EP_COMMON_TOOLCHAIN` instead of the project's shared `EP_COMMON_CMAKE_ARGS`, so they bypass the normalized C/C++ flags, fixed install layout, and CMake policy compatibility used by the rest of the third-party build. This can make these dependencies behave differently in sanitizer, customized-flag, and packaged builds. Could you start with `EP_COMMON_CMAKE_ARGS` here and then append the AWS-specific overrides? The same applies to `paimon_add_aws_c_project` below.

zjw1111 (commented):
Thanks for adding the lightweight object-store and S3 implementation. I found several configuration, endpoint, pagination, and build-consistency issues that should be addressed. Details are inline.

src/paimon/fs/s3/s3_file_system.h
```diff
@@ -0,0 +1,47 @@
...
+#include "paimon/common/fs/object_store_file_system.h"
+
+namespace paimon::s3 {
+
+inline constexpr char kS3RegionOption[] = "s3.region";
+inline constexpr char kS3EndpointOption[] = "s3.endpoint";
+inline constexpr char kS3PathStyleAccessOption[] = "s3.path-style-access";
+inline constexpr char kS3ProfileOption[] = "s3.profile";
+inline constexpr char kS3AccessKeyIdOption[] = "s3.access-key-id";
+inline constexpr char kS3SecretAccessKeyOption[] = "s3.secret-access-key";
```
zjw1111: Could we keep these option names compatible with the existing Paimon S3 configuration? Java and Python accept/document `s3.access-key` and `s3.secret-key`, while this implementation only looks for `s3.access-key-id` and `s3.secret-access-key`. Reusing a catalog configuration would silently ignore the explicit credentials and fall back to the ambient AWS provider chain, which can either fail or use an unintended principal. Please accept the established keys, optionally keeping these names as aliases, and add coverage for them.

src/paimon/fs/s3/s3_file_system.cpp
```diff
@@ -0,0 +1,759 @@
...
+    S3ObjectStoreClient(const std::map<std::string, std::string>& options,
+                        std::shared_ptr<HttpClient> http_client,
+                        std::shared_ptr<aws_credentials_provider> credentials)
+        : http_client_(std::move(http_client)), credentials_(std::move(credentials)) {
+        region_ = ResolveRegion(options);
+        auto endpoint = options.find(kS3EndpointOption);
+        use_default_endpoint_ = endpoint == options.end() || endpoint->second.empty();
+        endpoint_ = ParseEndpoint(use_default_endpoint_
+                                      ? fmt::format("https://s3.{}.amazonaws.com", region_)
+                                      : endpoint->second);
```
zjw1111: The default endpoint is hard-coded to `.amazonaws.com` for every region. That suffix is invalid for the `aws-cn` partition—for example, `cn-north-1` requires `s3.cn-north-1.amazonaws.com.cn`—so region-only configuration cannot access China buckets unless users also override the endpoint. Could you resolve the DNS suffix from the AWS partition (covering at least `aws`, `aws-cn`, and `aws-us-gov`), or explicitly require an endpoint for unsupported partitions?

src/paimon/fs/s3/s3_file_system.cpp
```diff
@@ -0,0 +1,759 @@
...
+        }
+        auto is_truncated = TagValue(body, "IsTruncated");
+        if (!is_truncated || (*is_truncated != "true" && *is_truncated != "false")) {
+            return Status::IOError("S3 ListObjectsV2 response has invalid IsTruncated");
+        }
+        result.is_truncated = *is_truncated == "true";
+        auto token = TagValue(body, "NextContinuationToken");
+        if (token) {
+            PAIMON_ASSIGN_OR_RAISE(result.continuation_token,
+                                   PercentDecode(*token, "NextContinuationToken"));
```
zjw1111: `NextContinuationToken` is opaque and is not one of the fields affected by `encoding-type=url` (only `Delimiter`, `Prefix`, `Key`, and `StartAfter` are). Decoding it changes tokens containing literal percent sequences: `%2F` becomes `/` and is encoded differently in the next request. Could you keep the XML-unescaped token verbatim and update `TestUrlEncodedListObjects` accordingly?

src/paimon/fs/s3/s3_file_system.cpp
```diff
@@ -0,0 +1,759 @@
...
+
+ private:
+    Status status_;
+};
+
+}  // namespace
+
+Result<bool> ParseBoolOption(const std::string& value) {
+    std::string lower = StringUtils::ToLowerCase(value);
+    if (lower == "true" || lower == "1" || lower == "yes" || lower == "on") {
```
zjw1111: Could we reuse `OptionsUtils::GetValueFromMap<bool>` / `StringUtils::StringToValue<bool>` here instead of maintaining a second boolean parser? The accepted spellings currently diverge: the shared parser supports `t/y/f/n`, while this one adds `on/off`. Is accepting `on/off` required by an actual S3 configuration scenario or compatibility target? If so, could we extend the shared parser with those spellings and tests, then reuse it here so boolean option semantics remain consistent across the project?

src/paimon/fs/s3/s3_file_system.cpp
```diff
@@ -0,0 +1,759 @@
...
+            authority = object.bucket + "." + authority;
+        }
+        PAIMON_ASSIGN_OR_RAISE(std::string encoded_key, PercentEncode(object.key, true));
+        request_path += encoded_key;
+        if (!query.empty()) {
+            request_path += "?" + query;
+        }
+
+        AwsAuthRuntime& runtime = GetAwsAuthRuntime();
+        aws_http_message* message = aws_http_message_new_request(runtime.allocator());
```
zjw1111: Could we use the existing `ScopeGuard` (or another scoped owner) for `message` and `signable` here? The same manual acquire/free pattern also appears in `PercentEncode`/`PercentDecode` for the curl-allocated buffers. Reusing the common RAII utility would keep cleanup correct across all error returns and align this code with the project's resource-management convention.

zjw1111 (commented):
I found two places where existing project utilities could be reused; details are inline.

mrdrivingduck:
> Thanks for adding the S3 integration. Have you tested it end to end against a real S3 bucket, in addition to the mocked HTTP tests? Could you share which operations were verified and whether the supported read path (HEAD, LIST, and range reads) worked as expected? The current filesystem implementation appears intentionally read-only, so could you also confirm that writes are out of scope, or, if write support is expected, whether it has been validated against real S3 as well?

Thanks for looking. 

Yes, I have a WIP duckdb-paimon branch that uses this branch, and I have tested the real S3 read path end to end through DuckDB SQL. The validation covered ATTACH, paimon_scan, and paimon_snapshots against a real Paimon table on S3, including metadata access, object listing, and range reads of the data files.

The S3 filesystem is intentionally read-only in this PR; write support is out of scope. I want to make the commit small.

Review comments will be addressed later. Thanks.

cmake_modules/BuildAwsAuth.cmake
mrdrivingduck: Done.

src/paimon/fs/s3/s3_file_system.h
mrdrivingduck: Done. It now uses the existing Paimon option names.

src/paimon/fs/s3/s3_file_system.cpp
mrdrivingduck: Done.

src/paimon/fs/s3/s3_file_system.cpp
mrdrivingduck: Done.

src/paimon/fs/s3/s3_file_system.cpp
mrdrivingduck: Done.

src/paimon/fs/s3/s3_file_system.cpp
mrdrivingduck: Done.

cmake_modules/BuildAwsAuth.cmake
```diff
@@ -0,0 +1,193 @@
...
+    # when it was built with Apple's native TLS backend.
+    if(APPLE AND (NAME STREQUAL "aws-c-http" OR NAME STREQUAL "aws-c-auth"))
+        list(APPEND AWS_PLATFORM_CMAKE_ARGS -DBYO_CRYPTO=ON)
+    endif()
+    externalproject_add(${NAME}_ep
+                        URL ${URL}
+                        URL_HASH SHA256=${CHECKSUM}
+                        DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+                        CMAKE_ARGS ${EP_COMMON_CMAKE_ARGS}
+                                   -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
```
zjw1111: some args such as CMAKE_BUILD_TYPE, DCMAKE_INSTALL_PREFIX, BUILD_SHARED_LIBS, BUILD_TESTING can be remove, already included in EP_COMMON_CMAKE_ARGS

src/paimon/fs/s3/s3_file_system_factory.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "paimon/factories/factory.h"
+#include "paimon/fs/s3/s3_file_system.h"
+
+namespace paimon::s3 {
+
+const char S3FileSystemFactory::IDENTIFIER[] = "s3";
+
+Result<std::unique_ptr<FileSystem>> S3FileSystemFactory::Create(
+    const std::string&, const std::map<std::string, std::string>& options) const {
+    PAIMON_RETURN_NOT_OK(ValidateS3Options(options));
```
zjw1111: I wonder why here `ValidateS3Options` first, and validate in `MakeS3ObjectStoreClient` again?

src/paimon/fs/s3/s3_file_system.cpp
```diff
@@ -0,0 +1,779 @@
...
+    if (!parsed.ok()) {
+        return Status::Invalid(
+            fmt::format("{} {}", kS3PathStyleAccessOption, parsed.status().message()));
+    }
+    return Status::OK();
+}
+
+S3FileSystem::S3FileSystem(const std::map<std::string, std::string>& options)
+    : ObjectStoreFileSystem("s3",
+                            MakeS3ObjectStoreClient(options, std::make_shared<CurlHttpClient>())) {}
```
zjw1111: maybe can use `S3FileSystem::Create()` to validate, and constructor does not return any exception value, and remove `InvalidObjectStoreClient`

src/paimon/fs/s3/s3_file_system.cpp
```diff
@@ -0,0 +1,779 @@
...
+
+    Result<HttpResponse> Execute(const ObjectStorePath& object, HttpMethod method,
+                                 const std::string& query, const HttpHeaders& headers,
+                                 const HttpBodyConsumer& consumer) const {
+        std::string authority = endpoint_.authority;
+        std::string request_path = endpoint_.base_path;
+        if (request_path.empty() || request_path.back() != '/') {
+            request_path += '/';
+        }
+        bool use_path_style = path_style_ || (!path_style_configured_ && use_default_endpoint_ &&
```
zjw1111: There is a different with aws sdk when determine path-style. Aws sdk check dotted-bucket as first, `s3.path-style-access` as second. But your current implementation is the opposite.

src/paimon/fs/s3/s3_file_system.cpp
```diff
@@ -0,0 +1,779 @@
...
+                fmt::format("GetObject read {} bytes for s3://{}/{}, expected {}", copied,
+                            path.bucket, path.key, size));
+        }
+        return copied;
+    }
+
+    void GetObjectRangeAsync(const ObjectStorePath& path, int64_t offset, int64_t size,
+                             char* buffer, std::function<void(Status)>&& callback) const override {
+        auto self = shared_from_this();
+        GetGlobalDefaultExecutor()->Add([self = std::move(self), path, offset, size, buffer,
```
zjw1111: I think use GlobalDefaultExecutor is dangerous, slow I/O will block the global thread pool (although no one else uses the global thread pool yet). You can consider using `CreateDefaultExecutor` to create a thread pool for s3 internal use.

cmake_modules/SetupCxxFlags.cmake
```diff
@@ -75,7 +75,6 @@ if("${BUILD_WARNING_LEVEL}" STREQUAL "CHECKIN")
         set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} -Wall")
         set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} -Wextra")
         set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} -Wdocumentation")
-        set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} -Wglobal-constructors")
```
zjw1111, line 78: I wonder why remove this line

cmake_modules/BuildUtils.cmake
```diff
@@ -180,7 +180,14 @@ function(add_paimon_lib LIB_NAME)
...
                               PUBLIC "$<BUILD_INTERFACE:paimon_sanitizer_flags>")
 
-        if(NOT APPLE)
+        if(APPLE)
+            # Arrow's checked_cast uses dynamic_cast in Debug builds, which
+            # fails across dylib boundaries on macOS due to duplicate RTTI
+            # from static Arrow archives.  Define NDEBUG so checked_cast
+            # falls back to static_cast (same as Release mode).
+            target_compile_definitions(${LIB_NAME}_objlib
+                                       PRIVATE $<$<CONFIG:Debug>:NDEBUG>)
```
zjw1111: I think this change is too broad in scope. It would cause all `assert` checks to become ineffective on macOS. Let's revert this change for now, and address the macOS build issues in a unified way later.

cmake_modules/SetupCxxFlags.cmake
mrdrivingduck: Sorry, it's a temporary fix when hacking on MacOS, I mistakenly add it into the commit. Will revert it.

cmake_modules/BuildAwsAuth.cmake
mrdrivingduck: Done.

src/paimon/fs/s3/s3_file_system_factory.cpp
mrdrivingduck: Done.

src/paimon/fs/s3/s3_file_system.cpp
mrdrivingduck: Done.

src/paimon/fs/s3/s3_file_system.cpp
mrdrivingduck: Done.

src/paimon/fs/s3/s3_file_system.cpp
mrdrivingduck: Done.

cmake_modules/BuildUtils.cmake
mrdrivingduck: Done.

zjw1111 (approved):
+1

PR #435, opened 2026-07-20
fix(blob): identify a missing blob file with FileSystem::Exists

SteNicholas:
### Purpose

Follow-up to #421.

`BlobFormatWriter` told a missing file apart from a fetch failure by testing `Status::NotExist` on a failed descriptor open. That makes the two write-null options depend on every file system plugin agreeing on which status code a missing file maps to, which is hard to keep uniform across all of the plugin interfaces: supporting it in #421 already required mapping `JDO_FILE_NOT_FOUND_ERROR` to `Status::NotExist` at the jindo boundary, and any plugin that reports a missing file as a plain IO error silently disables `blob-write-null-on-missing-file`.

- Decide whether the referenced file is missing with `FileSystem::Exists` instead of the status of a failed open. Java classifies the same way, through `UriReader.FileUriReader.exists` -> `FileIO.exists`; the only error code its writer tests is HTTP 404, which is protocol-defined rather than plugin-specific. The check runs only when `blob-write-null-on-missing-file` needs the classification, matching the gating of `FlinkRowWrapper.isNullAt` via `checkBlobDescriptorExists`: with only `blob-write-null-on-fetch-failure` enabled there is no check, and a missing file is converted (and counted) as a fetch failure like any other failed open. The default path (both options off) pays nothing.
- A failed open re-checks `FileSystem::Exists` once, so a file deleted between the up-front check and the open is still classified as missing rather than silently defeating `blob-write-null-on-missing-file`, whose writer does not convert a fetch failure. This narrows the race window rather than closing it; classifying from the open status instead would reintroduce the plugin-specific codes this PR removes.
- A failed existence check leaves it unknown whether the file is there. With `blob-write-null-on-fetch-failure` enabled it is deferred to the open, as `FlinkRowWrapper.deferExistsCheckFailure` does: a blob that opens fine is written as data, and only an open that also fails is handled as a fetch failure. Without fetch-failure handling to defer to, the write fails with a status naming the check and the blob URI.
- A descriptor that cannot be deserialized is now a fetch failure as well, which is what `BLOB_WRITE_NULL_ON_FETCH_FAILURE` already documented ("e.g. invalid descriptor") but the code did not do — it propagated unconditionally. Descriptor detection only inspects the version and magic header, so a descriptor truncated after those passes detection and then fails to deserialize. The status keeps its original code and detail (via `Status::WithMessage`) and gains an `invalid blob descriptor` prefix, because the underlying status comes from the byte reader and never mentions blobs.
- `BlobFormatWriter::Create` now rejects a null file system, so that configuration error surfaces at construction instead of being converted into silently NULL rows by `blob-write-null-on-fetch-failure`.
- Report both conversions as `blob.write.null-on-missing-file.count` and `blob.write.null-on-fetch-failure.count`, and name the blob URI, the BLOB field and the output file in the write-null warnings.

This PR also follows up the review comment on #421 (https://github.com/alibaba/paimon-cpp/pull/421#discussion_r3593647786): `${CMAKE_DL_LIBS}` was listed by hand on individual targets that never call `dlopen`/`dlsym`/`dladdr` themselves — a grep over `src/` and `include/` finds no such call. Those entries were carried on behalf of arrow, which declares dl as a usage requirement on its own interface, so CMake places `-ldl` after `libarrow.a` automatically on linkers that resolve symbols strictly left-to-right; every target that dropped an entry links arrow and inherits it transitively. dl is now declared only on the real providers (`arrow`, `lucene`, `jindosdk::nextarch`), with the convention recorded next to arrow's declaration so consumers do not add it back.

### Tests

- `BlobFormatWriterWriteNullTest.TestWriteNullClassifiesByExistence` (rewritten from `TestWriteNullClassifiesByStatusCode`): a mock file system whose `Open` always fails with `IOError` stands in for a plugin that reports a missing file as something other than `Status::NotExist`. The cases pin that classification follows `Exists` only when `blob-write-null-on-missing-file` is enabled: a file classified as missing is never opened under that option, while with only `blob-write-null-on-fetch-failure` the file is opened and its failure converted as a fetch failure. A `VanishingFileSystem` (present on the first `Exists`, absent afterwards) pins the deletion race: re-checked and classified as missing under `blob-write-null-on-missing-file` — alone and with both options enabled, attributed to the missing-file counter — converted as a fetch failure with only `blob-write-null-on-fetch-failure`, and existence never consulted when the missing-file option is off.
- `BlobFormatWriterWriteNullTest.TestWriteNullOnExistsCheckFailure` (reworked): a failed existence check defers to the open when `blob-write-null-on-fetch-failure` is enabled — a succeeding open writes the blob as data (verified by read-back), a failing open is handled as a fetch failure with one re-check — and fails the write with a status naming the check and keeping the underlying message when it is not; with only fetch-failure enabled the check never runs. Its mock delegates `Open` to a separate `LocalFileSystem` so the open still succeeds — `LocalFileSystem::Open` calls `Exists` on itself, and without that delegation the test could not tell whether the writer acted on the check at all — and optionally fails `Open` instead.
- `BlobFormatWriterWriteNullTest.TestWriteNullOnInvalidDescriptor`: a descriptor truncated after the magic header passes detection, fails to deserialize, and is handled as a fetch failure under both option settings, with read-back verification.
- `BlobFormatWriterWriteNullTest.TestWriteNullOnMissingFile` / `TestWriteNullOnBothOptionsEnabled`: unchanged in intent; the both-options case additionally asserts the two counters. `TestWriteNullOnFetchFailure`: a missing file is now converted (and counted) as a fetch failure, since no existence check runs without `blob-write-null-on-missing-file`.
- `BlobFormatWriterTest.TestCreateWithInvalidParameters`: extended with the null file system rejection.
- `BlobTableInteTest.TestWriteNullOnFetchFailureCoversMissingFile` (renamed from `TestWriteNullOnFetchFailureKeepsMissingFileFailing`): the end-to-end counterpart of the same change. With only `blob-write-null-on-fetch-failure` enabled, writing a descriptor whose file was deleted now succeeds; both the data file and the `.blob` file keep the full row count, and the read merges them back into a row whose blob is NULL. The test previously asserted that such a write fails, which is exactly the behavior this PR removes, so it was updated rather than kept.

Local validation: `paimon-blob-format-test` 41/41 passed; `pre-commit` (clang-format, cmake-format, cpplint, codespell) and `git diff --check` clean. Each regression test was checked for discriminating power by reverting the corresponding implementation change and confirming the test fails.

The `test/inte` binaries cannot be run on the development machine — they abort in Arrow's `json_simple.cc` `DCHECK(*out)` during `ArrayFromJSON`, including on blob tests this PR does not touch — so `paimon-blob-table-inte-test` was only compiled locally and its result relies on the Linux CI build.

The CMake change could not be validated meaningfully on the development machine: `CMAKE_DL_LIBS` expands to empty on macOS, so the removals are a no-op there; `format/lance`, `global_index/lucene` and `global_index/lumina` are parsed but their targets are guarded off, and `benchmark/CMakeLists.txt` is not parsed at all with `PAIMON_BUILD_BENCHMARKS=OFF`. It relies on the Linux CI build, and on CI covering those optional component flags.

### API and Format

- Header changes under `include/` are documentation-only: the option docs in `include/paimon/defs.h` now state that the missing-file classification applies only when `blob-write-null-on-missing-file` is enabled, and that a missing file otherwise follows `blob-write-null-on-fetch-failure`. The option constants are unchanged.
- No storage format change: NULL elements keep the existing `kNullBinLength` encoding.
- Behavior changes visible to callers: with only `blob-write-null-on-fetch-failure` enabled, a missing file is now converted to NULL as a fetch failure instead of failing the write, and no existence check runs; a failed existence check with fetch-failure enabled defers to the open instead of failing; an undeserializable descriptor is governed by `blob-write-null-on-fetch-failure` instead of always failing the write; `BlobFormatWriter::Create` rejects a null file system; and a failed existence check without fetch-failure handling reports `failed to check existence of blob file '<uri>'`.
- New internal metric name constants (`BlobMetrics`) in `src/paimon/format/blob/blob_format_writer.h`, reported through the existing `FormatWriter::GetWriterMetrics`.

### Documentation

No user-facing documentation change beyond the option semantics in `include/paimon/defs.h`; the class documentation in `blob_format_writer.h` states when the existence check runs and how the two options interact.

### Generative AI tooling

### DIFF
--- a/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cmake_modules/ThirdpartyToolchain.cmake
@@ -1643,6 +1643,9 @@ macro(build_arrow)
 
     # libarrow.a calls dlsym; keep ${CMAKE_DL_LIBS} in the interface so -ldl is placed
     # after libarrow.a on linkers that resolve symbols strictly left-to-right.
+    # Every library that uses dl itself (arrow here; also lucene and jindosdk::nextarch)
+    # declares it on its own interface the same way; consumers inherit it transitively
+    # and must not list ${CMAKE_DL_LIBS} again themselves.
     target_link_libraries(arrow
                           INTERFACE zstd
                                     snappy
--- a/src/paimon/format/blob/blob_format_writer.cpp
+++ b/src/paimon/format/blob/blob_format_writer.cpp
@@ -30,6 +30,7 @@
 #include "paimon/common/utils/arrow/status_utils.h"
 #include "paimon/common/utils/delta_varint_compressor.h"
 #include "paimon/data/blob.h"
+#include "paimon/fs/file_system.h"
 #include "paimon/io/byte_array_input_stream.h"
 #include "paimon/logging.h"
 
@@ -48,6 +49,8 @@ BlobFormatWriter::BlobFormatWriter(const std::shared_ptr<OutputStream>& out, con
       pool_(pool),
       write_null_on_missing_file_(write_null_on_missing_file),
       write_null_on_fetch_failure_(write_null_on_fetch_failure) {
+    // Create() has already checked that data_type has exactly one BLOB field.
+    blob_field_name_ = data_type_->field(0)->name();
     metrics_ = std::make_shared<MetricsImpl>();
     tmp_buffer_ = Bytes::AllocateBytes(kTmpBufferSize, pool_.get());
     magic_number_bytes_ = IntegerToLittleEndian<int32_t>(BlobDefs::kMagicNumber, pool_);
@@ -66,6 +69,9 @@ Result<std::unique_ptr<BlobFormatWriter>> BlobFormatWriter::Create(
     }
     if (pool == nullptr) {
         return Status::Invalid("blob format writer create failed. pool is nullptr");
+    }
+    if (fs == nullptr) {
+        return Status::Invalid("blob format writer create failed. fs is nullptr");
     }
     if (data_type->num_fields() != 1) {
         return Status::Invalid(
@@ -117,6 +123,10 @@ Status BlobFormatWriter::AddBatch(ArrowArray* batch) {
 }
 
 Status BlobFormatWriter::Flush() {
+    metrics_->SetCounter(BlobMetrics::WRITE_NULL_ON_MISSING_FILE_COUNT,
+                         null_on_missing_file_count_);
+    metrics_->SetCounter(BlobMetrics::WRITE_NULL_ON_FETCH_FAILURE_COUNT,
+                         null_on_fetch_failure_count_);
     return out_->Flush();
 }
 
@@ -140,29 +150,20 @@ Status BlobFormatWriter::Finish() {
 Status BlobFormatWriter::WriteBlob(std::string_view blob_data) {
     // Open the blob input stream before writing any bytes, so that a failed fetch can be
     // converted to a NULL element without leaving partial data in the output stream.
-    // Dynamically check whether blob_data is a serialized BlobDescriptor (by magic header)
-    // rather than relying on blob_as_descriptor_ config. This is consistent with Java behavior:
-    // at write time, the input bytes are auto-detected as descriptor or raw data.
+    // Whether blob_data is a serialized BlobDescriptor is detected by its magic header rather
+    // than taken from a blob_as_descriptor option, so each row may hold either form.
     std::unique_ptr<InputStream> in;
     PAIMON_ASSIGN_OR_RAISE(bool is_descriptor,
                            BlobDescriptor::IsBlobDescriptor(blob_data.data(), blob_data.size()));
     if (is_descriptor) {
-        Result<std::unique_ptr<InputStream>> opened = OpenDescriptorInputStream(blob_data);
-        if (!opened.ok()) {
-            const Status& status = opened.status();
-            // A missing file is only handled by 'blob-write-null-on-missing-file'; other fetch
-            // failures are only handled by 'blob-write-null-on-fetch-failure' (aligned with Java).
-            bool write_null =
-                status.IsNotExist() ? write_null_on_missing_file_ : write_null_on_fetch_failure_;
-            if (write_null) {
-                PAIMON_LOG_WARN(logger_, "Failed to open blob, writing NULL for BLOB field: %s",
-                                status.ToString().c_str());
-                bin_lengths_.push_back(BlobDefs::kNullBinLength);
-                return Status::OK();
-            }
-            return status;
+        PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<InputStream> descriptor_in,
+                               OpenDescriptorInputStream(blob_data));
+        // A null stream means a write-null option already converted the failure.
+        if (descriptor_in == nullptr) {
+            bin_lengths_.push_back(BlobDefs::kNullBinLength);
+            return Status::OK();
         }
-        in = std::move(opened).value();
+        in = std::move(descriptor_in);
     } else {
         in = std::make_unique<ByteArrayInputStream>(blob_data.data(), blob_data.size());
     }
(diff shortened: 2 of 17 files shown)


src/paimon/format/blob/blob_format_writer.cpp
```diff
@@ -206,10 +208,60 @@ Status BlobFormatWriter::WriteBlob(std::string_view blob_data) {
...
+            ++null_on_missing_file_count_;
+            return std::unique_ptr<InputStream>();
+        }
+    }
+
+    Result<std::unique_ptr<InputStream>> opened = blob->NewInputStream(fs_);
+    if (!opened.ok()) {
+        return HandleFetchFailure(blob->Uri(), opened.status());
+    }
+    return std::move(opened).value();
```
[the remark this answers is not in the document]
SteNicholas: Good catch, fixed in 3d4b9a2: a failed `NewInputStream()` now asks `FileSystem::Exists` once more and applies the missing-file handling when the file is gone.

Two corrections to the reasoning, though.

The combination you cite is not the damaging one. With only `write_null_on_fetch_failure_` enabled, converting the row to NULL is at least the same outcome the option produces for every other open failure. The worse case is the reverse: with only `write_null_on_missing_file_` enabled, a deletion inside the window made the write fail hard, because that writer does not convert a fetch failure — the option the user explicitly enabled was silently defeated by the race. Both are now handled.

The second half of the concern is not addressable by a re-check: if a file system reports `Exists=true` while `Open` fails with a not-exist-like error, asking `Exists` again returns `true` again. Only the deletion race is fixable this way.

Worth stating the limit explicitly: this narrows the window, it does not close it. No existence check outside the open is atomic with it, so a deletion between the re-check and its answer is still misclassified. Closing it properly would require reading the answer off the open's own status, which is precisely the plugin-specific coupling this PR removes — that trade is deliberate. A re-check that itself fails falls through to the open failure, since the file's state is then unknown.

Covered by three new cases in `BlobFormatWriterWriteNullTest.TestWriteNullClassifiesByExistence`, using a mock file system that reports the file present on the first `Exists()` and absent afterwards, with `Open()` failing as a plain `IOError` so the classification cannot come from the open status.

src/paimon/format/blob/blob_format_writer.cpp
```diff
@@ -206,10 +208,73 @@ Status BlobFormatWriter::WriteBlob(std::string_view blob_data) {
...
+    // open: file system implementations disagree on which status a missing file maps to. Checking
+    // up front, as Java does, keeps a file already known to be missing from being opened at all.
+    // Skipped when neither option is enabled, since the two cases then get the same treatment.
+    bool classify_by_existence = write_null_on_missing_file_ || write_null_on_fetch_failure_;
+    if (classify_by_existence) {
+        Result<bool> exists = fs_->Exists(blob->Uri());
+        if (!exists.ok()) {
+            // Unknown whether the file is there, so do not assume either answer.
+            return HandleFetchFailure(blob->Uri(), exists.status());
+        }
```
lxy-9602: Could you help confirm whether this path is fully aligned with Java?

In Java, when an existence pre-check fails with a non-404 error and fetch-failure handling is enabled, `FlinkRowWrapper` explicitly defers the error to the writer fetch path, allowing `newInputStream()` to be attempted. 
Also, when only `blob-write-null-on-fetch-failure` is enabled, Java does not perform the existence pre-check.

Here, an `Exists` failure is handled immediately by `HandleFetchFailure`, so a valid blob may be written as NULL even if `Open` would succeed. Would it make sense to continue with `Open` in this case and only write NULL if the actual open/fetch also fails?

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -1597,6 +1597,8 @@ macro(build_arrow)
 
     # libarrow.a calls dlsym; keep ${CMAKE_DL_LIBS} in the interface so -ldl is placed
     # after libarrow.a on linkers that resolve symbols strictly left-to-right.
+    # This is the single source of truth for dl: targets linking arrow inherit it
+    # transitively and must not list ${CMAKE_DL_LIBS} again themselves.
```
zjw1111: This comment is overly absolute. DL is also required in Jindo and Lucene, so it's not unique to Arrow.

src/paimon/format/blob/blob_format_writer.cpp
SteNicholas: You're right on both points — the path was not fully aligned. Fixed in 3d4b9a2.

- The existence pre-check now runs only when `blob-write-null-on-missing-file` is enabled, matching the gating of `FlinkRowWrapper.isNullAt` via `checkBlobDescriptorExists` (wired from `blobWriteNullOnMissingFile()` in `FlinkSinkBuilder`). With only `blob-write-null-on-fetch-failure` enabled there is no check, so a missing file is not told apart from any other failed open and is converted — and counted — as a fetch failure, the outcome Java produces through `openBlobInputStream` on a file system whose errors carry no HTTP status. The open-failure re-check from the thread above is gated the same way. This supersedes one sentence of my reply there: with only fetch-failure enabled, a deletion race is no longer reported as missing — in that configuration there is no existence check to consult, and the row is converted as a fetch failure.
- A failed existence check with fetch-failure handling enabled is now deferred to the open, as `FlinkRowWrapper.deferExistsCheckFailure` does: `NewInputStream` is attempted, a blob that opens fine is written as data, and only an open that also fails is handled as a fetch failure. Without fetch-failure handling the check failure still fails the write, matching the rethrow in `descriptorFileExists`; the status now names the check and the blob URI and keeps the underlying message.

One part of Java's defer condition is deliberately not ported: `deferExistsCheckFailure` also tests `!isNotFoundError(failure)`, so an exists check that fails with an HTTP 404 error is not deferred. That reads the HTTP status off the exception, which is the protocol-specific classification this writer cannot do across file system plugins, so a failed check here is always deferred and the open decides. In practice an HTTP reader reports 404 by returning `exists=false` rather than by failing the check, so the gap is narrow.

Covered by the reworked `TestWriteNullOnExistsCheckFailure` (deferral with a succeeding open, verified by read-back; with a failing open; the no-defer hard failure; and no check at all with only fetch-failure enabled) and the updated `TestWriteNullOnFetchFailure` / `TestWriteNullClassifiesByExistence` (a missing file under fetch-failure alone is opened and converted, existence never consulted). The option docs in `defs.h` and the writer docs now state the check is conditional on `blob-write-null-on-missing-file`.

End to end, `BlobTableInteTest.TestWriteNullOnFetchFailureCoversMissingFile` covers the same path through a real table write. It is renamed from `TestWriteNullOnFetchFailureKeepsMissingFileFailing`, which asserted the behavior this change removes: with only `blob-write-null-on-fetch-failure` enabled, writing a descriptor whose file was deleted now succeeds, the data file and the `.blob` file both keep the full row count, and the read merges them back into a row whose blob is NULL.

cmake_modules/ThirdpartyToolchain.cmake
SteNicholas: Right, that was overstated — reworded in 3d4b9a2: every library that itself uses dl (arrow, lucene, jindosdk::nextarch) declares ${CMAKE_DL_LIBS} on its own interface, and consumers must not list it again; arrow is one provider among them, not the single source.

zjw1111 (approved):
+1

PR #439, opened 2026-07-21
feat: add metrics storage-read-bytes for parquet and fix clean inte test

lucasfang:
### Purpose

Linked issue: close #xxx

This change bundles two independent items:

**1. Parquet byte-level read metrics (feature)**

Add two counters so callers can observe how many bytes a Parquet read actually consumes:

- `parquet.read.storage-read-bytes` — physical bytes read from storage.

**2. Revert #397 commit internal-retry / dedup (fix)**

`#397` changed the commit-exception branch in `TryCommitOnce` from `return Status::Invalid(...)` to
`return false`, driving an internal retry loop plus a `CheckCommitted`/`retry_start_snapshot_id`
dedup. This introduced flakiness in `CleanInteTest.TestDropPartitionAndExpireSnapshotWithIOException`
and diverged from Java, which has no equivalent `RetryCommitResult` type system in this C++ codebase.

This reverts to the pre-#397 behavior:

- Remove `retry_start_snapshot_id` from the `TryCommit` loop and the `TryCommitOnce` signature.
- Remove the `CheckCommitted` method (definition and declaration).
- Restore the commit-exception branch to
  `return Status::Invalid("You need call FilterAndCommit to retry commit for exception. ", ...)`.

Rationale: on a commit exception the outcome is uncertain (an atomic write may have timed out but
actually landed). Retrying internally can produce a duplicate snapshot. Safe retry belongs to the
upper layer via `FilterAndCommit`, which dedups by `CommitIdentifier()`; a bare `Commit` retry has no
such dedup and would create a duplicate. Returning `Status::Invalid` makes that contract explicit and
keeps `guard.Release()` so freshly written meta files are not cleaned up.

### Tests

- `TestCommitWithAtomicWriteSnapshotTimeoutAndActuallySucceed` restored to its pre-#397 assertions:
    the initial `Commit` returns an error (`ASSERT_NOK`) while the snapshot has physically landed, and
    recovery goes through `FilterAndCommit`, which idempotently filters the already-committed
    identifier (`num_committed == 0`).
- `CleanInteTest.TestDropPartitionAndExpireSnapshotWithIOException` — re-enabled (removed the
  `DISABLED_` prefix); the CI-friendly randomized loop step is kept. Verified deterministically by
  running the loop exhaustively (`i += 1`, 0..1000) locally: PASSED (~14 min), confirming the revert
  root-causes and eliminates the flakiness rather than lowering its probability.
- Parquet metric wiring covered by updated unit tests: `parquet_file_batch_reader_test.cpp`,
  `page_filtered_row_group_reader_test.cpp`, `predicate_pushdown_test.cpp` (Create call sites now pass
  the `storage_read_bytes` counter).

### API and Format

- No public API change: all touched declarations are internal (`src/paimon/...`); no headers under
  `include/paimon/` are modified.
- No storage format or protocol change.
- New observable metric counter key: `parquet.read.storage-read-bytes`.
- Internal signature changes: `ParquetFileBatchReader::Create`/constructor gain a required
  `storage_read_bytes` parameter; `FileStoreCommitImpl::TryCommitOnce` drops `retry_start_snapshot_id`
  and `CheckCommitted` is removed. All call sites are updated in this change.

### Documentation

No new user-facing feature documentation. The change adds observability counters and reverts an
internal retry behavior; no doc pages are introduced.


src/paimon/format/parquet/parquet_file_batch_reader.h
```diff
@@ -71,7 +73,8 @@ class ParquetFileBatchReader : public PrefetchFileBatchReader {
         std::shared_ptr<arrow::io::RandomAccessFile>&& input_stream,
         const std::map<std::string, std::string>& options, int32_t batch_size,
         std::shared_ptr<::parquet::FileMetaData> file_metadata,
-        const std::shared_ptr<arrow::MemoryPool>& pool);
+        const std::shared_ptr<arrow::MemoryPool>& pool,
```
zjw1111: make pool at last

src/paimon/common/utils/arrow/arrow_input_stream_adapter.cpp
```diff
@@ -45,7 +45,10 @@ arrow::Status ValidateArrowIoRange(int64_t value, const char* name) {
 ArrowInputStreamAdapter::ArrowInputStreamAdapter(
     const std::shared_ptr<paimon::InputStream>& input_stream,
     const std::shared_ptr<arrow::MemoryPool>& pool, int64_t file_size)
-    : input_stream_(input_stream), pool_(pool), file_size_(file_size) {
+    : input_stream_(input_stream),
```
lxy-9602, line 48: Please put `pool` to the last param.

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -65,18 +65,21 @@ namespace paimon::parquet {
 ParquetFileBatchReader::ParquetFileBatchReader(
     std::shared_ptr<arrow::io::RandomAccessFile>&& input_stream,
     std::unique_ptr<FileReaderWrapper>&& reader, const std::map<std::string, std::string>& options,
-    const std::shared_ptr<arrow::MemoryPool>& arrow_pool)
+    const std::shared_ptr<arrow::MemoryPool>& arrow_pool,
+    std::shared_ptr<std::atomic<uint64_t>> raw_input_bytes)
     : options_(options),
```
lxy-9602, line 70: Also place `arrow_pool` to the last pos.

lxy-9602 (commented):
+1

src/paimon/format/parquet/parquet_file_batch_reader.h
```diff
@@ -130,6 +133,8 @@ class ParquetFileBatchReader : public PrefetchFileBatchReader {
     }
 
     std::shared_ptr<Metrics> GetReaderMetrics() const override {
+        uint64_t storage = storage_read_bytes_ ? storage_read_bytes_->load() : 0;
+        metrics_->SetCounter(ParquetMetrics::READ_STORAGE_BYTES, storage);
```
zjw1111, line 137: `storage_read_bytes_` is populated by `ArrowInputStreamAdapter`, but in the prefetch path that adapter wraps `CacheInputStream`. Cache hits increment this counter without storage I/O, while `ReadAheadCache` performs physical reads through a separate stream. This counter therefore does not represent `parquet.read.storage-read-bytes`. Please measure below the cache or at the filesystem stream, and cover cache-hit/prefetch cases.

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -1250,11 +1220,21 @@ Result<bool> FileStoreCommitImpl::TryCommitOnce(
...
+        // To prevent the case where an atomic write times out but actually succeeds,
+        // retrying the commit could lead to the snapshot file being committed multiple times.
+        // Therefore, retries should be handled by the upper layer,
+        // which should call FilterAndCommit to avoid duplicate commits.
+        // Therefore, we should not trigger cleanup here,
+        // as it may delete meta files from a snapshot that was just written by ourselves,
+        // leading to an incomplete or corrupted snapshot.
         guard.Release();
-        return false;
+        return Status::Invalid("You need to call FilterAndCommit to retry commit for exception. ",
```
zjw1111, line 1236: Returning an error here bypasses the `TryCommit` retry loop via `PAIMON_ASSIGN_OR_RAISE`, so `commit.timeout`, `commit.max-retries`, and the removed `CheckCommitted` dedup are no longer applied to uncertain atomic outcomes. Current Java still returns `RetryCommitResult` and scans `commitUser + identifier + kind` before retrying. Please preserve equivalent retry/dedup semantics rather than shifting recovery to every caller.

src/paimon/core/operation/file_store_commit_impl_test.cpp
```diff
@@ -638,12 +638,7 @@ TEST_F(FileStoreCommitImplTest, TestCommitWithAtomicWriteSnapshotTimeoutAndActua
                               "/orc/append_09.db/append_09/commit_messages/commit_messages-01",
                           /*version=*/3);
     ASSERT_GT(msgs.size(), 0);
-    ASSERT_OK(commit->Commit(msgs, /*commit_identifier=*/1));
-    std::shared_ptr<Metrics> metrics = commit->GetCommitMetrics();
-    ASSERT_TRUE(metrics);
-    ASSERT_OK_AND_ASSIGN(uint64_t counter,
-                         metrics->GetCounter(CommitMetrics::LAST_COMMIT_ATTEMPTS));
-    ASSERT_EQ(2u, counter);
+    ASSERT_NOK(commit->Commit(msgs, /*commit_identifier=*/1));
```
zjw1111, line 641: This path now returns before `attempt += cnt`, while the scope guard still reports commit metrics. `LAST_COMMIT_ATTEMPTS` becomes `0` even though `AtomicStore` was attempted once. Please keep an assertion for this metric and count the failed attempt as `1`.

zjw1111 (changes requested):
Requesting changes for three remaining correctness and observability issues. The previously reported missing `ParquetFileBatchReader::Create` argument has been fixed.

zjw1111 (approved):
+1

Issue #92, opened 2026-01-28
[Feature] Support REST catalog implementation


SteNicholas:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/paimon-cpp/issues) and found nothing similar.

### Motivation

Paimon has supported Rest catalog in https://github.com/apache/paimon/pull/4553, which should also support for Paimon Cpp.

### Solution

Introduce REST catalog implementation.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #441, opened 2026-07-22
feat(rest): support rest catalog for databases, tables and snapshots

SteNicholas:
### Purpose

Linked issue: #92

Add a REST catalog implementation, selected via the `metastore=rest` option, that talks the Paimon REST catalog open API.

**New components**

- `RestHttpClient` — blocking libcurl client with exponential-backoff retries:
  - 429/503 responses are retried for all methods, transport errors only for idempotent methods via an explicit allowlist of transient failures (name-resolution and connect failures, timeouts, TLS failures and a server closing the connection without responding are never retried).
  - A `Retry-After` response header — delta-seconds or HTTP-date form, parsed locale-independently — takes precedence over the backoff when it yields a positive delay. Every retry sleep is bounded by a per-sleep cap and an overall retry budget; a `Retry-After` beyond either bound stops retrying rather than sleeping less than the server requested.
  - Redirects keep the method and body of POST/DELETE requests (`CURLOPT_POSTREDIR`) and are restricted to http(s) targets.
  - Header names must be valid HTTP tokens and header values must not contain CR/LF/NUL; a violating header fails the request before anything is sent.
  - Easy handles are pooled and reset between requests, so a request reuses the connection opened by the previous one instead of paying a TCP and TLS handshake every time; TLS below 1.2 is refused and requests carry a default `paimon-cpp` user agent that a `header.User-Agent` option overrides.
  - Per-request logging carries the request id (debug on success, warn on a retry or a final failure).
- `RestApi` — HTTP + JSON layer:
  - `/v1/config` option merging (overrides > client options > defaults), `header.` options sent as request headers, and paged listing.
  - Error mapping to `Status` (404 → `NotExist`, 409 → `Exist`, 400 → `Invalid`, 501 → `NotImplemented`) that prefers the code of the parsed error body over the http status.
  - Sensitive server messages are redacted, and response bodies are never echoed into error messages (they may carry credentials), so a body that is not an error object is reported as unparsable, distinct from an error object carrying no message.
  - Errors carry the request id (`x-request-id`, falling back to any `request-id` header) and attach a `RestErrorDetail` with the mapped code, so callers can distinguish e.g. authentication failures programmatically.
- `RestCatalog` — database/table create/drop/rename/list/get, snapshot listing, `table-default.` option defaults on table creation, and system/branch table checks:
  - The server table response is converted into a `TableSchema` with the computed `highestFieldId`: ids at or above `SpecialFieldIds::SYSTEM_FIELD_ID_START` are excluded, a duplicated field id at any nesting level is rejected, and a missing or wrong-typed `partitionKeys`/`primaryKeys`/`options` member fails instead of silently defaulting.
  - Branch identifiers are passed through to the server, which resolves the branch and returns the branch's own schema (the default branch `main` maps case-insensitively to the bare table); snapshots of a branch are listed under the branch object name.
  - The `sys` database serves the local global system tables like `FileSystemCatalog`.
- Bear token authentication provider (`token.provider=bear`, the protocol's historical spelling of "bearer").

**Shared code**

- The pieces overlapping with the object-store file systems are unified under `common/utils`: the generic HTTP client moved there from `common/fs` and is shared by the S3 file system and the REST catalog, `UrlUtils` carries both URL-encoding flavors (form-urlencoded for the REST api, RFC 3986 for S3) with the S3 client reusing it, and libcurl detection in CMake is a single `find_package(CURL)` gated on `PAIMON_ENABLE_S3 OR PAIMON_ENABLE_REST`.
- `CatalogUtils` holds the system-database and system/branch table checks shared by `FileSystemCatalog` and `RestCatalog`, with the check order and messages aligned with the Java `CatalogUtils` (`Cannot 'createTable' for system table ...`). `FileSystemCatalog` now also rejects branch identifiers on create/drop/rename — a branch identifier resolves to the main table directory, so dropping `t$branch_b` would otherwise delete the whole table.
- The new `SensitiveConfigUtils` centralizes credential redaction with the Java `SensitiveConfigUtils` key markers and masking rule (matched on the key lower-cased with separators removed; a key naming a true secret is masked as a whole, while an identifier-like key such as `dlf.access-key-id` keeps at most a four-character tail of a long enough value). `sys.catalog_options` masks every row through it so credentials such as the bearer token never surface to users who can query that table, and a server error message carrying a secret marker is redacted as a whole.
- `RestUtil::ExtractRequestId` and `RestAuthParameter::Create` likewise hold the request-id lookup and the query-parameter encoding in one place, so the log line and the error message report the same id and no call site can sign an unencoded parameter.

**Build**

libcurl is used from the system rather than bundled, so the new CMake option `PAIMON_ENABLE_REST` defaults to `OFF`, consistent with the other optional components that need something outside the bundled third-party toolchain (jindo, lance, lumina, lucene, tantivy). CI enables it explicitly in `ci/scripts/build_paimon.sh` and installs the libcurl development package, so the REST code and its tests are still built and run on every job. With the option off, `Catalog::Create` reports that `metastore=rest` requires a build with `PAIMON_ENABLE_REST=ON`.

Follow-ups (not in this PR): DLF signing, data-token file IO, paged listing parameters (`maxResults`/name patterns), and the remaining endpoints (alter/commit/views/partitions/tags/functions).

### Tests

New `rest_test` binary (62 cases, registered under the `unittest` label so it runs in CI), all against an in-process mock HTTP/REST server:

- `RestHttpClientTest`: uri normalization, query encoding, a pooled handle carrying no method or body over from the previous request, retry classification (429/503 retried, 404 and non-allowlisted transport errors such as a refused connection or a redirect loop not, truncated responses retried only for idempotent methods), `Retry-After` precedence in both forms plus the per-sleep and overall budget bounds, retry exhaustion, redirect following with POST keeping its method and body, rejection of invalid header names/values, and transport errors omitting the url and query.
- `RestUtilTest` / `ResourcePathsTest` / `RestMessagesTest`, each next to its implementation: prefix extraction and the request-id lookup including the gateway-header fallback, resource path building with url-encoded segments, JSON round trips of all request/response messages, and config merge with null-value filtering (a key present in defaults, client options and overrides at once, plus the null-override and null-default paths).
- `RestCatalogTest`: client-side option validation (missing uri/token, unsupported token provider), end-to-end catalog operations including pagination, `ignore_if_exists` / `ignore_if_not_exists` paths, client/server headers and the json content type, `table-default.` defaults, system tables, branch tables resolved by the server, broken-schema rejection (missing and wrong-typed `partitionKeys`/`primaryKeys`/`options` members), a non-empty partition-key round trip, server errors surfacing as errors instead of "does not exist", and snapshot listing with sorting.
- `RestApiErrorTest`: http-status-to-`Status` mapping including the request-id fallback, redaction and the `RestErrorDetail` code.

The shared pieces are covered next to their implementations:

- `UrlUtilsTest`: both encoding flavors, including that a literal `%` is encoded even when `/` is preserved, which is what the S3 client relies on.
- `SensitiveConfigUtilsTest`: key classification, whole-value and tail-preserving masking, and free-form text redaction.
- `HttpClientUtilTest` in the common utils suite.
- `TableSchemaTest.TestComputeHighestFieldId`: the highest-field-id computation.
- `SnapshotTest.TestToSnapshotInfo`: the snapshot-to-info conversion.
- `FileSystemCatalogTest.TestBranchIdentifierRejectedForTableOperations`: the branch rejection.
- The `sys.catalog_options` integration test: credential-carrying options (token, access-key secret, account key, credential, SAS) are masked in every emitted row, and an access-key id keeps only its four-character tail.

### API and Format

- New public header `include/paimon/catalog_options.h` with a `PAIMON_EXPORT`ed `CatalogOptions` struct holding the catalog-level option keys `METASTORE`, `URI`, `TOKEN`, `TOKEN_PROVIDER` and the `TABLE_DEFAULT_OPTION_PREFIX` (`table-default.`) key prefix; table-level keys stay in `Options` (`defs.h`), mirroring the CatalogOptions/CoreOptions separation of the Java implementation.
- `RestCatalog` implements the `Catalog` interface including `GetOptions()`, which returns the client options merged with the server-side `/v1/config`.
- `Catalog::Create` now dispatches on the `metastore` option (`filesystem` remains the default; unknown values are rejected). For `metastore=rest`, `root_path` is not a filesystem path but the warehouse (instance) name registered on the REST server, documented on the API.
- No storage format changes. The wire protocol is the Paimon REST catalog open API.

### Documentation

- `docs/source/user_guide/catalog.rst` now documents both metastores: the existing filesystem metastore section and a new REST Catalog section covering the `PAIMON_ENABLE_REST` build requirement, the `root_path` semantics (warehouse/instance name), the `CatalogOptions` keys (`metastore`, `uri`, `token.provider`, `token`, `table-default.<key>`) and a configuration example, replacing the note that claimed REST catalog support is future work. It also states which parts of the REST catalog are not implemented yet (altering a database or a table, views, functions, partitions, tags, branch management, consumers and the `dlf` token provider), so the guide is not read as promising them.
- `docs/source/building.rst` lists `-DPAIMON_ENABLE_REST=ON` under the optional components, including its libcurl requirement.

### Generative AI tooling

Generated-by: Claude Code (claude-fable-5, claude-opus-5)


src/paimon/rest/rest_util.cpp
```diff
@@ -0,0 +1,117 @@
...
+                                        rapidjson::Document::AllocatorType* allocator) {
+    rapidjson::Document doc;
+    doc.Parse(json.c_str());
+    if (doc.HasParseError()) {
+        throw std::invalid_argument("invalid json: " + json);
+    }
+    rapidjson::Value value;
+    value.CopyFrom(doc, *allocator);
+    return value;
+}
```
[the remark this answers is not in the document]
SteNicholas: Good catch, fixed. `ParseToValue` now reports only the rapidjson error description and offset (`GetParseError_En` + `GetErrorOffset`), matching the existing pattern in `variant_builder.cpp`, and never echoes the payload:

```cpp
throw std::invalid_argument(fmt::format("invalid json: {} (at offset {})",
                                        rapidjson::GetParseError_En(doc.GetParseError()),
                                        doc.GetErrorOffset()));
```

This was inconsistent with `http_client.cpp`, which already avoids putting response bodies into errors for the same reason. The concern is concrete here: the only callers parse the table schema JSON, and for `CreateTableRequest` that JSON is built from the table options, which can carry credentials. Added `RestMessagesTest.InvalidSchemaJsonErrorOmitsPayload` as a regression test asserting the secret value does not appear in the resulting `Status`.

CMakeLists.txt
```diff
@@ -63,9 +63,14 @@ option(PAIMON_ENABLE_LUMINA "Whether to enable lumina vector index" OFF)
 option(PAIMON_ENABLE_LUCENE "Whether to enable lucene index" OFF)
 option(PAIMON_ENABLE_TANTIVY
        "Whether to enable tantivy-fulltext global index (Rust FFI, experimental)" OFF)
+option(PAIMON_ENABLE_REST "Whether to enable the rest catalog (requires libcurl)" ON)
```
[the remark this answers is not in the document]
SteNicholas: This is out of date — the PR description was updated before this review was posted and now states that `PAIMON_ENABLE_REST` defaults to `OFF`, which matches the CMake option.

`OFF` is the intended default: libcurl is used from the system rather than built by the bundled third-party toolchain, so this follows the convention of the other optional components with external requirements (jindo, lance, lumina, lucene, tantivy), and avoids breaking existing builds on machines without the libcurl development package. CI enables it explicitly in `ci/scripts/build_paimon.sh` and installs libcurl, so the REST code and its tests still build and run on every job.

CMakeLists.txt
```diff
@@ -63,9 +63,14 @@ option(PAIMON_ENABLE_LUMINA "Whether to enable lumina vector index" OFF)
...
 if(PAIMON_ENABLE_ORC)
     add_definitions(-DPAIMON_ENABLE_ORC)
 endif()
+if(PAIMON_ENABLE_REST)
+    find_package(CURL REQUIRED)
+    add_definitions(-DPAIMON_ENABLE_REST)
+endif()
```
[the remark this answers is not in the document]
SteNicholas: Same as the comment above on line 66: the PR description has already been updated to state the `OFF` default, so the description and the CMake option agree. `OFF` is intentional because libcurl comes from the system rather than the bundled third-party toolchain; CI turns the option on explicitly and installs the development package.

src/paimon/rest/http_client.cpp
```diff
@@ -0,0 +1,263 @@
...
+        if (!retriable || execution_count > config_.max_retries) {
+            return result;
+        }
+        int64_t delay_ms =
+            GetRetryDelayMs(execution_count, result.ok() ? &result.value() : nullptr);
+        std::string reason = result.ok() ? fmt::format("http status {}", result.value().code)
+                                         : result.status().ToString();
+        PAIMON_LOG_WARN(logger_, "retrying %s %s in %lld ms (attempt %d/%d): %s", method.c_str(),
+                        path.c_str(), static_cast<long long>(delay_ms),  // NOLINT(runtime/int)
+                        execution_count, config_.max_retries, reason.c_str());
```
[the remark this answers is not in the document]
SteNicholas: The `attempt 3/2` case cannot occur. The early return precedes the log statement:

```cpp
if (!retriable || execution_count > config_.max_retries) {
    return result;
}
...
PAIMON_LOG_WARN(logger_, "retrying ... (retry %d/%d) ...", execution_count, config_.max_retries);
```

so the log is only reached while `execution_count <= max_retries`; with `max_retries = 2` the highest value printed is `2/2`, and `HttpClientTest.RetriesExhausted` pins the resulting 3 total attempts (initial + 2 retries).

The wording was ambiguous though, since the numerator is the retry about to be issued while the denominator is the retry budget, so I renamed it to `retry %d/%d`, which reads unambiguously as "retry N of at most max_retries".

src/paimon/rest/http_client.cpp
```diff
@@ -0,0 +1,263 @@
...
+    body->append(data, size * nmemb);
+    return size * nmemb;
+}
+
+// `StringUtils::Trim` only strips spaces; header lines end with "\r\n".
+void TrimWhitespace(std::string* str) {
+    const char* whitespace = " \t\r\n";
+    str->erase(str->find_last_not_of(whitespace) + 1);
+    str->erase(0, str->find_first_not_of(whitespace));
+}
```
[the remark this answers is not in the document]
SteNicholas: Leaving this as is. `find_last_not_of` returning `npos` followed by `+ 1` evaluating to `0` is a well-established idiom for exactly this trim pattern, and the behavior is correct for the empty and all-whitespace inputs: the first `erase(0)` clears the string, and the second `erase(0, npos)` is then a no-op. Adding an explicit `npos` branch would add a branch to two lines without changing behavior.

For context, this local helper exists only because `StringUtils::Trim` strips spaces alone and does not take a character set, which the comment above it records; there is no existing project helper to reuse here.

src/paimon/rest/http_client.h
```diff
@@ -0,0 +1,98 @@
...
+#include "paimon/status.h"
+
+namespace paimon {
+
+/// A blocking HTTP client for the REST catalog based on libcurl, mirroring the retry
+/// behavior of `HttpClient` plus `ExponentialHttpRequestRetryStrategy` in the Java
+/// implementation: HTTP 429/503 responses are retried for all methods, transport errors
+/// are retried only for idempotent methods, with exponential backoff (honoring a numeric
+/// `Retry-After` response header).
+class HttpClient {
```
zjw1111: Maybe need some refactor(such as http client, url encode/decode, cmake system), these classes have some overlap with #433 .

include/paimon/defs.h
```diff
@@ -112,6 +112,20 @@ struct PAIMON_EXPORT Options {
     /// Default value is local.
     static const char FILE_SYSTEM[];
 
+    /// "metastore" - Metastore of the paimon catalog.
```
lucasfang: Currently, defs.h only holds the configuration items for CoreOptions. Can we extract the Catalog-related configurations into a separate file? similar to how it's done in Java, and name it something like catalog_defs.h or catalog_options.h.

src/paimon/rest/rest_http_client.h
```diff
@@ -0,0 +1,98 @@
...
+        int64_t retry_base_delay_ms = 1000;
+    };
+
+    struct Response {
+        int64_t code = 0;
+        std::string body;
+        /// Response headers with lower-cased names.
+        std::map<std::string, std::string> headers;
+
+        bool IsSuccessful() const {
```
lucasfang, line 77: Please replace the magic numbers with named constants or enums (e.g., HTTP_OK, HTTP_ACCEPTED)

src/paimon/rest/rest_api.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+    // Mirrors DefaultErrorHandler: carry the request id through, it is what the server
+    // operators need to trace a failed call.
+    auto request_id_iter = response.headers.find("x-request-id");
+    if (request_id_iter != response.headers.end() && !request_id_iter->second.empty() &&
+        request_id_iter->second != "unknown") {
+        message += fmt::format(" requestId:{}", request_id_iter->second);
+    }
+    switch (response.code) {
+        case 400:
```
lucasfang: Please replace the magic numbers with named constants or enums

include/paimon/defs.h
SteNicholas: Done — added `include/paimon/catalog_options.h` with a `CatalogOptions` struct holding the catalog-level keys (`metastore`, `uri`, `token`, `token.provider`), defined in `src/paimon/common/catalog_options.cpp`; `Options` in `defs.h` keeps table-level keys only.

One deliberate difference from Java: `token`/`token.provider` live in `RESTCatalogOptions` there, but the C++ client only supports the bear provider and has no `dlf.*` options yet, so they stay in the single `CatalogOptions` header for now and can be split out once DLF authentication lands.

src/paimon/rest/rest_http_client.h
SteNicholas: Fixed — added an `HttpStatus` struct with named `constexpr` constants (`kOk`, `kAccepted`, `kNoContent`, …) following the project's `k`-prefix constant convention; `Response::IsSuccessful` and the 429/503 retry check in `http_client.cpp` now use them.

src/paimon/rest/rest_api.cpp
SteNicholas: Fixed — `ErrorToStatus` now switches over the same `HttpStatus` constants (`kBadRequest`, `kUnauthorized`, …, `kServiceUnavailable`) instead of numeric literals.

.github/workflows/build_and_test.yaml
```diff
@@ -67,6 +67,11 @@ jobs:
         uses: ./.github/actions/setup-ccache
         with:
           cache-key-prefix: ccache-${{ matrix.name }}
+      - name: Install dependencies
```
mrdrivingduck: Added same deps in #433, a clean rebase with main could be better? I think it bypass the conflict detection.

mrdrivingduck (changes requested):

.github/workflows/build_and_test.yaml
SteNicholas: Good catch, thanks. Rebased onto the latest main and dropped the duplicated install step from this PR — the step added in #433 already installs `libcurl4-openssl-dev`, so the CI change here is now only enabling `PAIMON_ENABLE_REST` in `build_paimon.sh`.

src/paimon/rest/rest_catalog.cpp
```diff
@@ -0,0 +1,447 @@
...
+    return fs_;
+}
+
+Result<std::vector<SnapshotInfo>> RestCatalog::ListSnapshots(const Identifier& identifier,
+                                                             const std::string& branch) const {
+    if (!branch.empty()) {
+        return Status::NotImplemented(
+            "do not support listing snapshots of a branch for the rest catalog.");
+    }
+    PAIMON_RETURN_NOT_OK(CheckNotSystemTable(identifier, "list snapshots"));
```
zjw1111: I wonder if this path should also reject a branch that is encoded in the identifier itself. Every other table-level read path in this class normalizes the identifier through `GetDataTableName()` before hitting the api (`TableExists`, `GetTableLocation`, `LoadDataTableSchema`), but `RestApi::ListSnapshots` uses `Identifier::GetTableName()`, which keeps the raw object name. So the same request expressed two ways behaves differently:

- `ListSnapshots(Identifier("db", "t1"), "b1")` returns `NotImplemented`
- `ListSnapshots(Identifier("db", "t1$branch_b1"), "")` is sent to the server as `/tables/t1%24branch_b1/snapshots`

Note that `CheckNotSystemTable` does not catch the second form, since `$branch_` is parsed into the branch component rather than the system-table component.

Would it be possible to add `PAIMON_RETURN_NOT_OK(CheckNotBranch(identifier, "list snapshots"));` here, so both spellings behave the same until branch snapshot listing lands? A test with a branch-suffixed identifier would be a nice addition too.

For reference, Java resolves branches on the server side instead (`RESTApi.listSnapshotsPaged` and `RESTApi.getTable` both pass `Identifier.getObjectName()` through). Adopting that model here would also require changing the other read paths, so it seems out of scope for this PR — just flagging it in case you would rather keep the identifier untouched and follow up later.

zjw1111 (commented):
Thanks for putting this together — the alignment with the Java implementation is impressively thorough (config merge precedence, retry classification, `isSuccessful` status set, paging termination, audit option keys all match).

One small consistency point on snapshot listing, left inline.

src/paimon/rest/http_client.h
SteNicholas: Done in the latest revision — the overlapping pieces are unified under `common/utils`: the generic HTTP client moved from `common/fs` to `common/utils/http_client` and is shared by the S3 file system and the REST catalog; `UrlUtils` now carries both URL-encoding flavors (form-urlencoded for the REST api, RFC 3986 for S3) and the S3 client reuses it instead of its curl-based helpers; the REST client reuses the shared curl global-init guard, `ScopeGuard` cleanup and header-line parsing; and libcurl detection in CMake is a single `find_package(CURL)` gated on `PAIMON_ENABLE_S3 OR PAIMON_ENABLE_REST`.

src/paimon/rest/rest_catalog.cpp
SteNicholas: Good catch, thanks — addressed in the latest revision, which goes with the server-side resolution model you referenced rather than keeping the two spellings apart:

- `ListSnapshots` now rejects a branch encoded in the identifier via `CatalogUtils::CheckNotBranch`, so `ListSnapshots(Identifier("db", "t1$branch_b1"), "")` and the `branch` argument form no longer behave differently.
- The branch is passed through the `branch` argument and sent under the branch object name (`t1$branch_b1`), so the server resolves the branch and lists its own snapshots; the default branch `main` maps case-insensitively to the bare table.
- The other read paths (`TableExists`, `GetTableLocation`, `LoadTableSchema`) also keep the branch in the object name sent to the server now, so branch reads resolve server-side consistently instead of silently falling back to the main table.

The `ListSnapshots` test covers the branch-suffixed identifier rejection, listing through the `branch` argument (including `"MAIN"`), and a branch unknown to the server surfacing as `NotExist`.

src/paimon/rest/rest_http_client.cpp
```diff
@@ -0,0 +1,362 @@
...
+    curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT_MS,
+                     static_cast<long>(config_.connect_timeout_ms));  // NOLINT(runtime/int)
+    curl_easy_setopt(curl, CURLOPT_TIMEOUT_MS,
+                     static_cast<long>(config_.request_timeout_ms));  // NOLINT(runtime/int)
+    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, WriteBodyCallback);
+    curl_easy_setopt(curl, CURLOPT_WRITEDATA, &response.body);
+    curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, WriteHeaderCallback);
+    curl_easy_setopt(curl, CURLOPT_HEADERDATA, &response.headers);
+    // Follow redirects transparently, restricted to http(s) targets.
+    curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
```
zjw1111, line 275: Optional hardening, non-blocking for this PR.

`CURLOPT_CUSTOMREQUEST` only changes the verb in the request line; `CURLOPT_POSTFIELDS` changes libcurl's internal method to `HTTPREQ_POST`, and that is the only thing libcurl looks at on a redirect. With `CURLOPT_POSTREDIR` unset, 301/302 rewrite POST to GET and 303 rewrites any non-GET method, dropping the body — and a custom verb is discarded too, since `http_switch_to_get()` sets `http_ignorecustom`. 307/308 are unaffected.

The DELETE path is not reachable today: `DropDatabase` and `DropTable` pass an empty body, so the internal method stays `HTTPREQ_GET` and no rewrite applies. It becomes reachable if a body-carrying delete (a `purge`/`cascade` parameter) is added later, since the non-empty-body branch below sets `CURLOPT_POSTFIELDS`.

POST is reachable, though only against a redirecting endpoint. `CreateDatabase` and `CreateTable` would be replayed as GET on `/v1/{prefix}/databases` and `.../tables`, which are the list endpoints and return 200; since both call sites only inspect `Execute(...).status()`, the client reports `Status::OK()` while nothing was created. A conforming server will not redirect these paths, so I do not think it blocks the PR.

Would it be possible to cover both whenever convenient — either `CURLOPT_POSTREDIR, CURL_REDIR_POST_ALL`, or restricting `CURLOPT_FOLLOWLOCATION` to idempotent methods and surfacing 3xx to the caller for POST/DELETE? Happy for this to be a follow-up.

src/paimon/rest/rest_http_client.cpp
```diff
@@ -0,0 +1,362 @@
...
+                    std::clamp(retry_after_seconds.value(), -kMaxSeconds, kMaxSeconds) * 1000;
+            } else {
+                std::optional<int64_t> date_seconds = ParseHttpDateSeconds(iter->second);
+                if (date_seconds) {
+                    retry_after_ms = (date_seconds.value() - now_epoch_seconds) * 1000;
+                }
+            }
+            // Non-positive values fall through to the exponential backoff.
+            if (retry_after_ms && retry_after_ms.value() > 0) {
+                return retry_after_ms.value();
```
zjw1111, line 371: Could you bound this delay by a client retry budget? The clamp above only prevents overflow during the seconds-to-milliseconds conversion; every positive `Retry-After` value is returned verbatim and later passed to `sleep_for`. A malformed or overly large server value can therefore block the calling thread for hours or longer. Please add a maximum per-retry delay and an overall retry deadline; when `Retry-After` exceeds the remaining budget, stop retrying rather than retrying earlier than requested.

src/paimon/rest/rest_http_client.cpp
```diff
@@ -0,0 +1,362 @@
...
+        case CURLE_GOT_NOTHING:
+        case CURLE_SSL_CONNECT_ERROR:
+        case CURLE_SSL_CERTPROBLEM:
+        case CURLE_SSL_CIPHER:
+        case CURLE_PEER_FAILED_VERIFICATION:
+        case CURLE_SSL_CACERT_BADFILE:
+        case CURLE_SSL_ISSUER_ERROR:
+            return false;
+        default:
+            return true;
```
zjw1111, line 127: Could you make retriable transport errors an explicit allowlist instead of returning true for every unlisted CURLcode? Permanent failures such as `CURLE_URL_MALFORMAT`, `CURLE_UNSUPPORTED_PROTOCOL`, `CURLE_TOO_MANY_REDIRECTS` and `CURLE_OUT_OF_MEMORY` currently run through all retries for idempotent methods without any chance of recovering — a misconfigured `uri` costs six attempts and roughly 31s of backoff before returning the same error, and a redirect loop costs 6 x `CURLOPT_MAXREDIRS` real requests. The comment above also documents the opposite guarantee ("Returns false for the transport failure kinds that do not become healthy by retrying"), which a blacklist cannot provide.
`IsRetryable` in `common/utils/http_client.cpp` already uses the allowlist shape and can be aligned with. Please enumerate only transient failures that are known to benefit from retrying and add a negative test for a permanent error.

src/paimon/core/catalog/file_system_catalog.cpp
```diff
@@ -425,10 +421,7 @@ Status FileSystemCatalog::DropTableImpl(const Identifier& identifier,
 }
 
 Status FileSystemCatalog::DropTable(const Identifier& identifier, bool ignore_if_not_exists) {
-    PAIMON_ASSIGN_OR_RAISE(bool is_system_table, IsSystemTable(identifier));
-    if (is_system_table) {
-        return Status::Invalid(fmt::format("Cannot drop system table {}.", identifier.ToString()));
-    }
+    PAIMON_RETURN_NOT_OK(CatalogUtils::CheckNotSystemTable(identifier, "drop table"));
```
zjw1111: Could you also call `CheckNotBranch` from `FileSystemCatalog::CreateTable`, `DropTable`, and `RenameTable` (for both rename identifiers)? `IsSystemTable()` is false for a branch identifier, while `GetTableLocation()` ultimately uses `GetDataTableName()` and strips the `$branch_*` suffix. As a result, dropping `t$branch_b` resolves to the main `t` directory and can delete the entire table; rename has the same base-path problem. Java `AbstractCatalog` rejects branch identifiers for all three operations.

src/paimon/rest/rest_http_client.cpp
```diff
@@ -0,0 +1,362 @@
...
+        if (updated_list == nullptr) {
+            return false;
+        }
+        header_list = updated_list;
+        return true;
+    };
+    // Disable the automatic "Expect: 100-continue" behavior of libcurl on POST.
+    bool headers_ok = append_header("Expect:");
+    for (const auto& [name, value] : headers) {
+        headers_ok = headers_ok && append_header(name + ": " + value);
```
zjw1111, line 314: Could you validate every header name and value before appending it to the curl list? libcurl sends custom header strings verbatim without filtering control characters. Since these values can come from both client options and the server-provided defaults/overrides, a value containing CRLF can inject additional headers into subsequent requests. Please require a valid HTTP token for the name and reject CR, LF, and NUL in values before storing `base_headers` or before constructing the curl list.

include/paimon/utils/special_field_ids.h
```diff
@@ -39,6 +39,10 @@ class SpecialFieldIds {
 
     /// Special field ID reserved for index score. Value: CPP_FIELD_ID_END - 1
     static const int32_t INDEX_SCORE = CPP_FIELD_ID_END - 1;
+
+    /// The lowest field ID reserved for system fields; IDs at or above this bound are
+    /// excluded when computing the highest field ID of a schema. Value: INT32_MAX / 2
+    static const int32_t SYSTEM_FIELD_ID_START = std::numeric_limits<int32_t>::max() / 2;
```
zjw1111, line 45: Could you make this member `static constexpr int32_t`, or add the corresponding out-of-line definition in `special_field_ids.cpp`? A plain in-class `static const` integral member is not implicitly inline in C++17. The current comparison is constant-folded, but any external address-taking or reference-binding use of this public-header constant will ODR-use it and fail to link because this is the only `SpecialFieldIds` constant without a definition.

docs/source/building.rst
```diff
@@ -145,6 +145,7 @@ boolean flags to ``cmake``.
 * ``-DPAIMON_ENABLE_AVRO=ON``: Apache Avro libraries and Paimon integration
 * ``-DPAIMON_ENABLE_JINDO=ON``: Support for Alibaba Jindo filesystems
 * ``-DPAIMON_ENABLE_LUMINA=ON``: Support for Lumina vector index, lumina is only supported on gcc9 or higher.
+* ``-DPAIMON_ENABLE_REST=ON``: Support for the REST catalog (``metastore=rest``), requires the libcurl development package.
```
zjw1111, line 148: Could you also update `docs/source/user_guide/catalog.rst`? It still says that C++ Paimon only supports the filesystem catalog and that REST catalog support is future work. The `Catalog::Create` documentation should also explain that `root_path` is the warehouse/instance name for `metastore=rest` and point users to the new `CatalogOptions` keys (`metastore`, `uri`, and token settings), otherwise the newly added feature is difficult to discover and the user guide directly contradicts this PR.

src/paimon/rest/rest_catalog.cpp
```diff
@@ -0,0 +1,431 @@
...
+                               GlobalSystemTableLoader::IsSupported(identifier.GetTableName(),
+                                                                    api_->GetMergedOptions()));
+        if (!supported) {
+            return Status::NotExist(fmt::format("{} not exist", identifier.ToString()));
+        }
+        GlobalSystemTableContext context;
+        context.catalog = this;
+        context.fs = fs_;
+        context.warehouse = warehouse_;
+        context.catalog_options = api_->GetMergedOptions();
```
zjw1111, line 370: Could you filter or redact sensitive options before passing them to `CatalogOptionsSystemTable`? `GetMergedOptions()` still contains the bearer `token` (and may contain other storage/authentication credentials), while `CatalogOptionsSystemTable::BuildRows()` emits every key/value unchanged. Although the table is disabled by default, a server override can enable `catalog-options-table.enabled`, which can expose an engine service credential to users who can query `sys.catalog_options`. Please centralize sensitive-key masking and apply it when building these rows.

src/paimon/rest/rest_catalog.cpp
```diff
@@ -0,0 +1,431 @@
...
+        out.AddMember(rapidjson::StringRef("id"), response.GetSchemaId(), allocator);
+        PAIMON_ASSIGN_OR_RAISE(int32_t highest_field_id,
+                               TableSchema::ComputeHighestFieldId(src["fields"]));
+        rapidjson::Value fields(rapidjson::kArrayType);
+        fields.CopyFrom(src["fields"], allocator);
+        out.AddMember(rapidjson::StringRef("highestFieldId"), highest_field_id, allocator);
+        out.AddMember(rapidjson::StringRef("fields"), fields.Move(), allocator);
+        for (const char* key : {"partitionKeys", "primaryKeys"}) {
+            rapidjson::Value keys(rapidjson::kArrayType);
+            if (src.HasMember(key) && src[key].IsArray()) {
```
zjw1111: Could you reject a missing `partitionKeys`, `primaryKeys` or `options` member instead of defaulting it to an empty value? The problem is not just strictness: an omitted or misspelled `partitionKeys` makes a partitioned table load as unpartitioned, and a dropped `options` map silently changes the table's configuration — both are silent semantic changes rather than a visible failure, and the resulting `TableSchema` is then used for reads.
For reference, Java does not default absent members either. Jackson passes `null` into `Schema`'s `@JsonCreator` constructor, and construction then fails for all three: `options` fails immediately in `new HashMap<>(options)`; `partitionKeys` survives `normalizePartitionKeys` as `null` and then fails in `normalizeFields` via `duplicateFields(partitionKeys)`; `primaryKeys` fails on `primaryKeys.isEmpty()` in the same method. 
Please distinguish among absent, wrong-typed, and valid, return `Status::Invalid` for the first two on all three members, and add a non-empty partition-key round-trip test plus negative tests for the missing and wrong-typed cases.

src/paimon/rest/rest_http_client.cpp
SteNicholas: Done in the latest revision — the client now sets `CURLOPT_POSTREDIR, CURL_REDIR_POST_ALL` so 301/302/303 keep the method and body instead of replaying a body-carrying request as a bodyless GET (this also keeps the custom DELETE verb, covering the future body-carrying delete), and redirect targets are restricted to http(s) via `CURLOPT_REDIR_PROTOCOLS`. `RestHttpClientTest.PostRedirectKeepsMethodAndBody` asserts a redirected POST arrives at the target with method and body intact; `RedirectIsFollowed` and `RedirectLoopIsNotRetried` cover the plain redirect and the loop case.

src/paimon/rest/rest_http_client.cpp
SteNicholas: Done in the latest revision — `RestHttpClient::Config` now carries `retry_max_delay_ms` (64s default) bounding every single sleep and `retry_timeout_ms` (5min default) as an overall budget spanning all attempts and backoff sleeps. A positive `Retry-After` beyond either bound stops retrying entirely rather than sleeping less than the server requested, exactly as you suggested. Covered by `ComputeRetryDelayMs`, `RetryDelayBounds` and `ExcessiveRetryAfterStopsRetrying`.

src/paimon/rest/rest_http_client.cpp
SteNicholas: Done in the latest revision — `IsRetriableTransportError` is now an explicit allowlist (`CURLE_SEND_ERROR`, `CURLE_RECV_ERROR`, `CURLE_PARTIAL_FILE`, `CURLE_HTTP2`, `CURLE_HTTP2_STREAM`): failures where the request reached a live server and broke mid-flight. Everything else — malformed urls, resolution/connect failures, timeouts, TLS failures, redirect loops — costs one attempt instead of the retry schedule, and the comment now documents the allowlist rationale. Negative tests: `ConnectionRefusedIsNotRetried`, `EmptyReplyIsNotRetried`, `RedirectLoopIsNotRetried`; positive: `TruncatedResponseIsRetriedForGet`/`ForDelete` (and `NotRetriedForPost`).

src/paimon/core/catalog/file_system_catalog.cpp
SteNicholas: Good catch, thanks — done in the latest revision. `FileSystemCatalog::CreateTable`/`DropTable`/`RenameTable` (both rename identifiers) now call `CatalogUtils::CheckNotBranch` before the system-table check, matching Java `AbstractCatalog`, with a comment on the drop path spelling out the hazard you describe (a branch identifier resolving to the main table directory). `FileSystemCatalogTest.TestBranchIdentifierRejectedForTableOperations` covers all three operations and asserts the table itself survives the rejected calls.

src/paimon/rest/rest_http_client.cpp
SteNicholas: Done in the latest revision — `RestHttpClient::Execute` validates every header before anything is sent: names must be RFC 7230 tokens and values must not contain CR, LF or NUL; a violating header fails the request up front. The error message deliberately does not echo the offending name or value (the name may contain the very control bytes being rejected, the value may carry credentials). Since validation sits on the single request path, it covers both client `header.` options and server-provided config values. Covered by `RestHttpClientTest.InvalidHeadersAreRejected`.

include/paimon/utils/special_field_ids.h
SteNicholas: Done in the latest revision — added the out-of-line definition in `src/paimon/core/utils/special_field_ids.cpp`, next to the existing definitions of the other `SpecialFieldIds` constants (kept as `static const` + definition rather than switching this one member to `constexpr` inline, so the class stays uniform).

docs/source/building.rst
SteNicholas: Done in the latest revision — `docs/source/user_guide/catalog.rst` now documents both metastores: the filesystem section is reworded and a new "REST Catalog" section replaces the stale future-work note, covering the `PAIMON_ENABLE_REST` build requirement, the `root_path`-is-the-warehouse/instance-name semantics, the `CatalogOptions` keys (`metastore`, `uri`, `token.provider`, `token`, `table-default.<key>`) and a configuration example. The `Catalog::Create` API doc in `include/paimon/catalog/catalog.h` also explains the `root_path` meaning for `metastore=rest`.

src/paimon/rest/rest_catalog.cpp
SteNicholas: Good catch, thanks — done in the latest revision. Sensitive-key masking is centralized in `CatalogUtils::MaskSensitiveOptions` and applied in `CatalogOptionsSystemTable::BuildRows`, so the merged options (including server-enabled `catalog-options-table.enabled`) never expose credentials. The marker list matches the Java `SensitiveConfigUtils` key patterns (password/secret/token/credential/accesskey/accountkey/encryptionkey/authorization/privatekey/apikey/sas), matched on the key lower-cased with separators removed so `dlf.access-key-secret` and `fs.s3a.access.key` are both caught. The `sys.catalog_options` integration test asserts the masked values and that no raw credential appears in any emitted row.

src/paimon/rest/rest_catalog.cpp
SteNicholas: Done in the latest revision — `RestCatalog::ToTableSchema` now distinguishes absent, wrong-typed and valid for all three members and returns `Status::Invalid` for the first two: missing or non-array `partitionKeys`/`primaryKeys`, missing or non-object `options`, with a comment noting that defaulting would silently change table semantics and that Java rejects absent members too. `RestCatalogTest.BrokenSchemaRejected` covers the missing and wrong-typed cases per member, and `RestCatalogTest.PartitionKeysRoundTrip` adds the non-empty partition-key round trip you asked for.

src/paimon/rest/rest_http_client.cpp
SteNicholas: Correction to my earlier reply: the shipped code sets `CURLOPT_POSTREDIR, CURL_REDIR_POST_301 | CURL_REDIR_POST_302`, not `CURL_REDIR_POST_ALL` — I described the intent rather than what landed, sorry.

301/302 keep the method and body, so the `CreateDatabase`/`CreateTable` replay you described cannot happen for those codes, and the custom DELETE verb survives them too (covering the future body-carrying delete). 303 is deliberately left to become a GET: RFC 9110 §15.4.4 defines it as "retrieve a representation of the target with GET", and the Java client behaves the same way — Apache HttpClient 5 preserves the method for 301/302/307/308 and converts only 303. Redirect targets are still restricted to http(s) via `CURLOPT_REDIR_PROTOCOLS`.

That does leave the narrow residual case you identified, for 303 only: a server answering `POST /v1/{prefix}/databases` with a 303 pointing at the list endpoint would have the request replayed as GET, return 200, and `CreateDatabase` would report `Status::OK()` although nothing was created. It requires a server that violates the open API, and closing it by keeping POST on a 303 would diverge from both the RFC and the Java client, so I left the behavior as is rather than silently taking `CURL_REDIR_POST_ALL`.

Happy to switch to `CURL_REDIR_POST_ALL`, or to surface any 3xx to the caller for non-idempotent methods instead of following it, if you would rather have the client refuse to be redirected at all on a create path — just say which you prefer and I will follow up.

src/paimon/rest/rest_catalog.cpp
SteNicholas: Follow-up on where this ended up: the masking no longer lives in `CatalogUtils::MaskSensitiveOptions` as I wrote above. It moved to a dedicated `SensitiveConfigUtils` under `common/utils`, because the same markers were needed for two surfaces — the option map behind `sys.catalog_options` and the free-form server error message that `RestApi::ErrorToStatus` reports — and keeping one copy in the catalog layer meant the REST layer had a second, drifting one.

The class now mirrors the Java `SensitiveConfigUtils` API and masking rule rather than only its key markers:

- `IsSensitiveKey` / `RedactValue` / `RedactMap` / `RedactText`, with the same 11 key markers, the same 12 text markers (the key set plus `signature`, which in an option key names a signing algorithm rather than a credential) and the same literal Azure SAS markers (`sig=`, `"sig"`, `'sig'`).
- `RedactValue` masks a key naming a true secret as a whole, while an identifier-like key keeps the last four characters of a value at least twelve characters long — so `dlf.access-key-secret` becomes `******` while `dlf.access-key-id` becomes `****id-1`, enough to tell two credentials apart in a support case without disclosing either. This matches Java's `redactValue`/`maskValue`.

`CatalogOptionsSystemTable::BuildRows` calls `RedactMap`, and the `sys.catalog_options` integration test asserts both the fully masked values and the tail-only form. Worth noting this is stricter than Java, where `CatalogOptionsTable.toRow` emits option values verbatim.

src/paimon/rest/rest_catalog.cpp
SteNicholas: One correction to my reply above, for the record: when I wrote that "the default branch `main` maps case-insensitively to the bare table", the code compared the branch name case-sensitively, so `t1$branch_MAIN` was sent to the server as a branch and came back `NotExist`. That was intent, not behavior.

It is implemented now: `NormalizeBranch` lower-cases before comparing, matching how the Java client builds the identifier it sends (`Identifier`'s constructor uses `!"main".equalsIgnoreCase(branch)`), so `t1$branch_MAIN` and `ListSnapshots(identifier, "MAIN")` both address the bare table. `BranchTableLoadsBranchSchemaFromServer` now asserts `t1$branch_MAIN` loads the main table's schema with no `branch` option set, and `ListSnapshots` asserts the `"MAIN"` argument returns the main table's snapshots.

Note this inherits a quirk of that model: a branch literally named `MAIN` cannot be addressed, since the client resolves it to the bare table before the request is sent. `BranchManager::IsMainBranch`, which names the on-disk branch directory, stays case-sensitive — the normalization only decides how a table is addressed on the server.

zjw1111 (approved):
+1

mrdrivingduck (approved):

zjw1111:
Close because already merged at https://github.com/apache/paimon-cpp/pull/165

PR #443, opened 2026-07-23
feat(parquet): support page-level (page index) filtering for nested column types (patch arrow)

zhf999:
### Purpose

Page-level filtering (row skipping driven by the Parquet page index / OffsetIndex)
was previously **disabled whenever the read schema contained a nested column**
(struct / list / map). `ParquetFileBatchReader::SetReadSchema` detected any nested
field and fell back to the non-page-filtered path, so nested reads could not skip
non-matching pages at the I/O and decode level.

This change enables page-level filtering for nested column types and reworks the
reader so that each leaf column is filtered and driven independently.

What changed:

- **Enable page filtering for nested schemas.** Removed the `has_nested_field`
  workaround in `ParquetFileBatchReader` that skipped page-index pushdown for both
  the predicate and the selection-bitmap paths when any nested field was present.

- **Nested-aware page-filtered reader (`PageFilteredRowGroupReader`).** Each
  top-level field is read through an Arrow `ColumnReader` tree
  (`FileReader::GetColumn` + leaf projection). A per-leaf `data_page_filter` is
  installed to skip non-matching pages during decode, and each leaf is then driven
  independently with reset / skip / read primitives. The result schema is built
  from the actually-produced column types (needed because column pruning can change
  nested types).

- **Arrow patch (`cmake_modules/arrow.diff`).** Instead of a single monolithic
  entry point, `parquet::arrow::ColumnReader` gains small, per-leaf primitives:
  `LeafColumnIndices()`, `ResetLeaf()`, `SkipRecords()` and `ReadRecords()`
  (returning the actual record count), implemented across the leaf / extension /
  list / struct readers. `parquet::arrow::FileColumnIterator::NextChunk()` is made
  `virtual` so the page-filter wiring lives in a Paimon-side
  `PageFilteringColumnIterator` subclass, keeping the patch surface minimal.

### Tests

Unit tests in `src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp`:

- Nested row-group and page filtering: `NestedStructColumnRowGroupFilter`,
  `NestedListColumnRowGroupFilter`, `NestedMapColumnRowGroupFilter`,
  `MultipleAdjacentNestedColumns`, `MultipleNestedColumns`.
- Leaf / sub-column projection: `NestedStructColumnOnlyReadIdField`,
  `NestedStructSubColumnProjection`.
- Misaligned pages across nested leaves: `NestedColumnsMisalignedPages` — writes a
  `struct<key:int64, s:struct<x:int32,y:int64>, tags:list<int32>,
  props:map<utf8,int32>>` with `write_batch_size=1` and a byte-based data page size
  so leaves paginate at different (and, for the utf8 map key, irregular) row counts;
  selects two disjoint row ranges via a bitmap and deep-compares every returned row
  against the source data to catch any per-leaf skip/read desync.
- Existing flat page-filtering, `ComputePageRanges`, streaming and bitmap cases
  continue to pass.

### API and Format

- Paimon public API under `include/paimon/` is unchanged.
- The bundled Arrow build is patched via `cmake_modules/arrow.diff`: new virtual
  methods on `parquet::arrow::ColumnReader` and a virtual
  `FileColumnIterator::NextChunk()`. This affects only the vendored Arrow used by
  the read path, not any symbol shipped from `include/paimon/`.
- No change to the Paimon storage format, on-disk layout, or protocol; this is a
  read-path optimization.

### Documentation

No user-facing documentation change; this is an internal read-path optimization.

### Generative AI tooling

Generated-by:  GLM 5.2


src/paimon/format/parquet/page_filtered_row_group_reader.h
```diff
@@ -57,12 +56,10 @@ class PageFilteredRowGroupReader {
...
     static Result<std::unique_ptr<arrow::RecordBatchReader>> ReadFilteredRowGroup(
-        ::parquet::ParquetFileReader* parquet_reader, const TargetRowGroup& target_row_group,
-        const std::vector<int32_t>& column_indices,
-        const std::shared_ptr<arrow::Schema>& arrow_schema,
-        const ::arrow::io::CacheOptions& cache_options, bool pre_buffered,
-        const std::vector<::arrow::io::ReadRange>& page_ranges, int64_t max_chunksize,
-        std::shared_ptr<::arrow::MemoryPool> pool);
+        ::parquet::arrow::FileReader* arrow_file_reader, const TargetRowGroup& target_row_group,
+        const std::vector<int32_t>& column_indices, const ::arrow::io::CacheOptions& cache_options,
+        bool pre_buffered, const std::vector<::arrow::io::ReadRange>& page_ranges,
```
lxy-9602: Please put the output to the last param.

src/paimon/format/parquet/page_filtered_row_group_reader.cpp
```diff
@@ -277,31 +301,40 @@ Result<std::unique_ptr<arrow::RecordBatchReader>> PageFilteredRowGroupReader::Re
...
 
         columns.push_back(std::move(chunked_array));
     }
 
-    auto table = arrow::Table::Make(arrow_schema, std::move(columns), expected_rows);
-    return std::make_unique<TableRecordBatchReader>(std::move(table), max_chunksize,
-                                                    std::move(pool));
+    std::vector<std::shared_ptr<arrow::Field>> result_fields;
+    for (size_t i = 0; i < columns.size(); ++i) {
+        const auto& field = manifest.schema_fields[field_indices[i]].field;
```
lxy-9602, line 330: The name `columns` may be a bit misleading here, since it seems to refer only to leaf columns now. Would it make sense to rename it?

src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp
```diff
@@ -1065,26 +1051,27 @@ static std::shared_ptr<arrow::StructArray> MakeMapColumnData(int32_t num_rows) {
...
 }
 
 /// Test: rowgroup-level filtering on a file with a list column.
 ///
 /// Schema: { id: int32, tags: list<item: int32> }
 /// 100 rows, 10 per page, 2 row groups.
-/// Predicate: id >= 70 → row groups 0 skipped, row groups 1 read → 50 rows expected.
+/// Predicate: id >= 70 → page 0-7 skipped, page 8-9 read → 30 rows expected.
 TEST_F(PageFilteredRowGroupReaderTest, NestedListColumnRowGroupFilter) {
     std::string file_name = dir_->Str() + "/nested_list_filter.parquet";
```
lxy-9602, line 1057: `NestedListColumnRowGroupFilter`? May change to PageFilter?

lxy-9602 (dismissed):
+1

src/paimon/format/parquet/page_filtered_row_group_reader_test.cpp
```diff
@@ -44,6 +47,7 @@
 #include "paimon/status.h"
 #include "paimon/testing/utils/read_result_collector.h"
 #include "paimon/testing/utils/testharness.h"
+#include "paimon/testing/utils/variant_test_data.h"
```
zjw1111: remove unused headers

lxy-9602 (approved):
+1

PR #446, opened 2026-07-23
fix: nested list schema evolution

XiaoHongbo-Hope:
### Purpose

Reading a data-evolution table fails when a field was added inside a nested `list<struct>` or `map<…, struct>` column. Old files written before the field was added error with `Invalid: PruneDataType does not support partial projection inside list`.

The read schema is a superset of the file's, but the reader rejected the nested shape difference instead of null-filling the added field. This change makes such files readable immediately (the added nested field is null-filled), without a rewrite. `PruneDataType` now accepts added fields inside `list`/`map` structs while still rejecting field drops, renames, and leaf type changes; a new `AlignArrayToReadType` reshapes each file's nested array to exactly the read type — null-filling added fields, matching struct children by name (Parquet drops nested field-ids), decoding ORC dictionary leaves, and applying read-side nullability.

### Tests

- Unit (`nested_projection_utils_test.cpp`, `field_mapping_reader_test.cpp`): added field inside `list<struct>` / `map`; drop-and-add still fails; ORC dictionary-leaf decode; read-side nullability; nested `large_binary` blob stays `large_binary`.
- Integration (`write_and_read_inte_test.cpp`): `TestSchemaEvolutionAddFieldInsideListAndMap` — evolve `list<struct>` / `map` schema, read back, assert added fields are null.
- Full `paimon-core-test` and `paimon-write-and-read-inte-test` pass locally.

### API and Format

Read-path only. No public API or storage format change.

### Generative AI tooling

Co-authored with Claude 4.8.


src/paimon/core/utils/nested_projection_utils_test.cpp
```diff
@@ -222,6 +222,92 @@ TEST(NestedProjectionUtilsTest, PruneDataTypeListDroppingSiblingOfVariantStillFa
...
+                         NestedProjectionUtils::AlignArrayToReadType(list_arr, read_type, pool));
+    ASSERT_TRUE(aligned->type()->Equals(*read_type)) << aligned->type()->ToString();
+    auto out_struct = std::static_pointer_cast<arrow::StructArray>(
+        std::static_pointer_cast<arrow::ListArray>(aligned)->values());
+    auto c_col = out_struct->GetFieldByName("c");
+    ASSERT_NE(c_col, nullptr);
+    ASSERT_EQ(c_col->null_count(), c_col->length());  // added field is all null
+    auto a_col = std::static_pointer_cast<arrow::Int32Array>(out_struct->GetFieldByName("a"));
+    ASSERT_EQ(a_col->Value(0), 1);
+    ASSERT_EQ(a_col->Value(2), 3);
```
lxy-9602, line 308: I wonder if it would be clearer to write the data in JSON format here, similar to `field_mapping_reader_test.cpp`.

src/paimon/core/utils/nested_projection_utils_test.cpp
```diff
@@ -222,6 +222,92 @@ TEST(NestedProjectionUtilsTest, PruneDataTypeListDroppingSiblingOfVariantStillFa
...
+}
+
 TEST(NestedProjectionUtilsTest, HasNestedSubfieldProjectionNoProjection) {
     auto file_schema = arrow::schema({
```
lxy-9602, line 399: Could you please add integration tests for map and list in `write_and_read_inte_test.cpp` to ensure the end-to-end path is covered?

src/paimon/common/reader/data_evolution_file_reader.cpp
```diff
@@ -82,7 +86,13 @@ Result<BatchReader::ReadBatchWithBitmap> DataEvolutionFileReader::NextBatchWithB
...
         assert(sub_array->num_fields() > field_offsets_[i]);
-        target_sub_array_vec.push_back(sub_array->field(field_offsets_[i]));
+        // Null-fill nested fields added by schema evolution (no-op otherwise).
+        PAIMON_ASSIGN_OR_RAISE(
+            std::shared_ptr<arrow::Array> aligned,
+            NestedProjectionUtils::AlignArrayToReadType(sub_array->field(field_offsets_[i]),
+                                                        read_schema_->field(i)->type(),
+                                                        arrow_pool_.get()));
+        target_sub_array_vec.push_back(aligned);
     }
```
lxy-9602, line 90: One question I still have: if each file is already aligned via `AlignArrayToReadType` in `field_mapping_reader.cpp`, why do we need to do the same thing again in `data_evolution_file_reader.cpp`?

XiaoHongbo-Hope: > One question I still have: if each file is already aligned via `AlignArrayToReadType` in `field_mapping_reader.cpp`, why do we need to do the same thing again in `data_evolution_file_reader.cpp`?

Good catch, updated.

src/paimon/core/utils/nested_projection_utils_test.cpp
XiaoHongbo-Hope: > Could you please add integration tests for map and list in `write_and_read_inte_test.cpp` to ensure the end-to-end path is covered?

Sure, added.

src/paimon/core/utils/nested_projection_utils_test.cpp
XiaoHongbo-Hope: > I wonder if it would be clearer to write the data in JSON format here, similar to `field_mapping_reader_test.cpp`.

Sure, updated

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -601,8 +679,9 @@ Result<std::shared_ptr<arrow::Array>> NestedProjectionUtils::AlignArrayToReadTyp
             return arrow::MakeArray(new_data);
         }
         default:
-            return Status::Invalid(fmt::format("AlignArrayToReadType cannot reconcile {} to {}",
-                                               array->type()->ToString(), read_type->ToString()));
+            // Leaf: cast to the read type (decodes an ORC dictionary, widens
+            // large_string to string, etc.).
+            return CastingUtils::Cast(array, read_type, arrow::compute::CastOptions::Safe(), pool);
```
lxy-9602, line 712: I wonder why does `AlignArrayToReadType` use generic Arrow safe cast for all leaf mismatches? If this helper is only for reshaping nested arrays and decoding dictionaries, please restrict it to dictionary/format-representation cases and fail other leaf mismatches. If nested type evolution is intended, please use the same schema-evolution cast rules as `FieldMappingReader` instead of raw Arrow `Cast`.

XiaoHongbo-Hope: Good point — restricted it to representation-only changes. The leaf branch now normalizes away the two physical differences that legitimately reach it under ORC lazy decoding (a dictionary wrapper and the string/binary 32/64-bit offset width) and casts only when the normalized types match; any other leaf mismatch now fails fast with `AlignArrayToReadType unsupported leaf type change`.

Genuine nested type evolution is intentionally not handled here: it is already rejected upstream in `PruneDataType`/`PruneRepeatedItemType` (leaf `type_id` must match), and top-level type promotion continues to go through `FieldMappingReader` cast executors. Added `AlignArrayToReadTypeRejectsLeafTypeChange` to cover the new guard.

XiaoHongbo-Hope: > I wonder why does `AlignArrayToReadType` use generic Arrow safe cast for all leaf mismatches? If this helper is only for reshaping nested arrays and decoding dictionaries, please restrict it to dictionary/format-representation cases and fail other leaf mismatches. If nested type evolution is intended, please use the same schema-evolution cast rules as `FieldMappingReader` instead of raw Arrow `Cast`.

 Restricted the leaf cast to representation-only cases (dictionary decode + string/binary width); other type mismatches now fail fast. Added a test.

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -588,6 +588,27 @@ Result<std::shared_ptr<arrow::Array>> NestedProjectionUtils::FilterMapArrayBySel
...
+        t = std::static_pointer_cast<arrow::DictionaryType>(t)->value_type();
+    }
+    switch (t->id()) {
+        case arrow::Type::LARGE_STRING:
+            return arrow::utf8();
+        case arrow::Type::LARGE_BINARY:
+            return arrow::binary();
+        default:
+            return t;
+    }
```
lxy-9602, line 605: I’m not sure we should change large here. As far as I remember, only string has dictionary encoding while binary does not, and blob is the only type using large binary, so changing it to binary may not be right.

XiaoHongbo-Hope: > I’m not sure we should change large here. As far as I remember, only string has dictionary encoding while binary does not, and blob is the only type using large binary, so changing it to binary may not be right.

Thanks, removed the large_binary normalization.

duanyyyyyyy: SR will use large_binary converter to convert blob col to SR chunk.

lxy-9602 (commented):
LGTM

src/paimon/core/utils/nested_projection_utils.cpp
XiaoHongbo-Hope: Blob columns don't go through this path — their type already matches the read type, so `AlignArrayToReadType` returns them untouched, and it emits the read type anyway. This change is orthogonal to blob and the large_binary converter.

XiaoHongbo-Hope:
> LGTM

Thanks, could you please help trigger CI again.

lucasfang:
Could you please add more details to the PR description?

lucasfang:
+1

lxy-9602 (approved):
+1

XiaoHongbo-Hope:
> Could you please add more details to the PR description?

Added

PR #450, opened 2026-07-24
fix(abi): dynamic_cast error on Predicate across shared libraries

### LINKED ISSUE #449: dynamic_cast/dynamic_pointer_cast on Predicate fails across shared-library boundaries (missing RTTI key function)
paimon::Predicate objects created in one shared library (e.g. libpaimon.so) can fail to dynamic_cast / dynamic_pointer_cast to their derived types (LeafPredicate, CompoundPredicate, PredicateFilter) when the cast is performed in a different DSO — for example inside a format plugin such as libpaimon_parquet_file_format.so / libpaimon_orc_file_format.so / third party plugins.

The cast returns nullptr even though the runtime type is correct, which silently breaks predicate handling (e.g. predicate pushdown in PredicateConverter)

zhangweilst:
### Purpose

Linked issue: close #449

### Tests

Added predicate_abi_inte_test

### API and Format

No

### Documentation

No

### DIFF
--- a/src/paimon/common/predicate/predicate.cpp
+++ b/src/paimon/common/predicate/predicate.cpp
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2026-present Alibaba Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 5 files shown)


src/paimon/common/predicate/predicate.cpp
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2024-present Alibaba Inc.
```
zjw1111: Thanks for adding this file. Could you update the license header year here? New files added to the repo this year use `Copyright 2026-present` (e.g. the recent `common/data/variant/*` and `common/data/shredding/*` files), whereas this one still carries `Copyright 2024-present`.

The same applies to the other new file in this PR, `test/inte/predicate_abi_inte_test.cpp` — it would be great to align both to `2026-present`.

zjw1111 (commented):
Thanks for the fix and for the thorough regression test that guards the key-function anchoring at the symbol level.

src/paimon/common/predicate/predicate.cpp
zhangweilst: Sure, updated and fixed the test failure

zjw1111:
+1

lxy-9602 (approved):
+1

PR #451, opened 2026-07-27
test: add ut for release-0.3

lszskye:
### Purpose

This change improves reader correctness and expands coverage for several previously untested paths.

- Make asynchronous union global-index tasks own their actions and predicate arguments, avoiding dangling references when an earlier task throws while other tasks remain queued.
- Align streaming `$binlog` behavior by packing adjacent `UPDATE_BEFORE` and `UPDATE_AFTER` records into one list-valued update event.
- Avoid adding `_VALUE_KIND` twice when constructing no-merge read schemas.
- Fix `$files` statistics projection for partitioned partial writes by mapping statistics only to columns physically written to the data file.
- Add coverage for bucket selection, Hive smallint hashing, list aggregation, compaction rewrites, eager deletion-file generation, index partition filtering, and manifest cleanup after failures.

### Tests
- deferred asynchronous union-index tasks after an earlier task throws
- streaming binlog update packing within and across batches
- unmatched `UPDATE_BEFORE` handling
- binlog sequence-number schema
- partitioned partial-write statistics in `$files`
- default, Hive, and MOD bucket selection
- Hive smallint bucket compatibility
- listagg distinct mode with an empty delimiter
- full-compaction file rewrites
- eager deletion-vector generation
- partition-filtered index manifest reads
- cleanup of newly generated manifest files after merge failure


src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -141,23 +213,53 @@ class ChangelogBatchReader : public BatchReader {
...
-        PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.Reserve(value_kind_array->length()));
-        for (int64_t i = 0; i < value_kind_array->length(); ++i) {
-            if (value_kind_array->IsNull(i)) {
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.Reserve(row_group_lengths.size()));
+        int64_t offset = 0;
+        for (int32_t row_group_length : row_group_lengths) {
+            int64_t rowkind_index = offset + row_group_length - 1;
+            if (value_kind_array->IsNull(rowkind_index)) {
                 PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.AppendNull());
-                continue;
```
lxy-9602, line 156: In what scenario can `value_kind_array` be null?

src/paimon/core/table/system/audit_log_system_table.cpp
```diff
@@ -186,17 +290,38 @@ class ChangelogTableRead : public TableRead {
...
     }
 
     Result<std::unique_ptr<BatchReader>> CreateReader(
         const std::shared_ptr<Split>& split) override {
-        std::vector<std::shared_ptr<Split>> splits = {split};
-        return CreateReader(splits);
+        PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<BatchReader> reader,
+                               data_read_->CreateReader(split));
+        bool pack_update_before_after = false;
+        if (converter_->PackUpdateBeforeAfter()) {
```
lxy-9602, line 322: Could you clarify how the `CreateReader` logic is related to the number of splits?

src/paimon/core/table/system/metadata_system_tables.cpp
```diff
@@ -348,6 +347,42 @@ Result<std::vector<std::string>> RowValueStrings(const std::vector<DataField>& f
...
+    partition_fields.reserve(table_schema->PartitionKeys().size());
+    for (const auto& partition_key : table_schema->PartitionKeys()) {
+        PAIMON_ASSIGN_OR_RAISE(DataField field, table_schema->GetField(partition_key));
+        partition_fields.push_back(std::move(field));
+    }
+    PAIMON_ASSIGN_OR_RAISE(std::vector<std::string> partition_values,
+                           RowValueStrings(partition_fields, partition));
+
+    size_t length = std::min(partition_fields.size(), partition_values.size());
+    for (size_t i = 0; i < length; ++i) {
```
lxy-9602: In what cases can the numbers of `partition_fields` and `partition_values` be different?

src/paimon/core/table/system/system_table_test.cpp
```diff
@@ -51,6 +59,32 @@ Result<std::shared_ptr<TableSchema>> CreateTableSchemaForTest(
     return std::shared_ptr<TableSchema>(std::move(table_schema));
 }
 
+class ArrayBatchReader : public BatchReader {
+ public:
```
lxy-9602: Why not use `MockFileBatchReader`？

src/paimon/core/table/system/system_table_test.cpp
```diff
@@ -68,6 +102,97 @@ TEST(SystemTableTest, TestChangelogArrowSchemaReturnsInvalidOptions) {
...
+        ASSERT_OK_AND_ASSIGN(BatchReader::ReadBatch batch, reader->NextBatch());
+        if (BatchReader::IsEofBatch(batch)) {
+            break;
+        }
+        auto& [c_array, c_schema] = batch;
+        arrow::Result<std::shared_ptr<arrow::Array>> output_result =
+            arrow::ImportArray(c_array.get(), c_schema.get());
+        ASSERT_TRUE(output_result.ok()) << output_result.status().ToString();
+        output_batches.push_back(std::move(output_result).ValueOrDie());
+    }
```
lxy-9602: Why not use `ReadResultCollector`?

src/paimon/core/table/system/audit_log_system_table.cpp
lszskye: Keeping the original call direction would result in infinite recursion:
CreateReader(single split)
  -> CreateReader(vector of splits)
    -> CreateReader(single split)
      -> ...

src/paimon/core/table/system/audit_log_system_table.cpp
lszskye: In the current scenario, value_kind_array cannot be null. This check is only a defensive safeguard.

lxy-9602 (approved):
+1

PR #453, opened 2026-07-28
feat(blob): support placeholder fallback for partial updates

### LINKED ISSUE #452: [Feature] Support row-level blob placeholder fallback for data-evolution partial updates
### Search before asking

### Motivation

Java Paimon supports partial updates of blob columns on data-evolution tables: a partial update rewrites only the touched rows, and the new blob file records every untouched row as a placeholder entry (`BlobFormatWriter.PLACE_HOLDER_LENGTH = -2`, no data bytes). At read time, `BlobFallbackRecordReader` groups the blob files of a bunch by `maxSequenceNumber` and resolves each row to the newest layer holding a real (non-placeholder) entry.

paimon-cpp currently supports neither side of this protocol:

- The blob format writer has no way to produce a `-2` placeholder entry, and the blob format reader does not understand one.
- `DataEvolutionSplitRead::BlobBunch` keeps only the newest file per row range and rejects overlapping layers, so tables produced by Java engines via data-evolution partial updates (e.g. Spark `MERGE INTO` touching a blob column) either fail to read or lose the fallback semantics.

### Solution

Align the whole chain with the Java implementation:

1. **Format layer**: `bin_length == -2` marks a placeholder entry (no payload bytes, offset not advanced), byte-compatible with the Java writer. The write side detects a placeholder value in-band and records `-2`. By default the blob reader fails when it encounters a placeholder entry; the internal format option `blob.internal.emit-placeholder-sentinel` switches it to emit an in-band 9-byte sentinel (`0x01` + `"BLOBPLHD"`) instead. The sentinel is in-band (rather than Java's singleton reference comparison) because `format/blob` is a plugin library and placeholder semantics must survive the main library's schema-mapping readers — the same precedent as sniffing serialized `BlobDescriptor`s by their magic header.
2. **`BlobBunch`**: keep all blob files (aligned with Java's `BlobFileBunch`), where files sharing a `max_sequence_number` form one non-overlapping layer. When every file shares a single `max_sequence_number`, `SequentialReadOptimize()` keeps the existing concat fast path.
3. **`BlobFallbackBatchReader`** (new): merges the layers of one blob bunch, ordered by `max_sequence_number` descending; row-id ranges a layer does not cover are padded with placeholder gap segments so all layers step in lockstep; each output row takes the first non-placeholder layer, and a row that is a placeholder in every layer degrades to null. Row-range pushdown is honored per layer, including inside gaps.
4. **`DataEvolutionSplitRead`**: when a blob bunch spans multiple sequence layers, build the fallback reader instead of the concat reader, and pass the internal format option through `AbstractSplitRead::CreateRawFileReaders`.

Covered by format-level golden-bytes tests (byte-aligned with `BlobFormatWriterTest` in Java), unit tests for the fallback merge (batch-boundary sweeps, gaps, all-placeholder, null-wins semantics), `BlobBunch` layering tests, and end-to-end tests including a compacted multi-layer scenario mirroring Java's `BlobUpdateTest.testReadCompactedBlobSequenceGroups`, per-row row-range reads, and a `blob_as_descriptor` read-mode variant.

### Anything else?

Known divergences from Java left for follow-up discussion:

- `BlobBunch` requires all layers to share one `schema_id`, while Java's `BlobFileBunch` only requires equal `writeCols`; a partial update committed after a schema change would currently fail to read in C++ and may need the check relaxed.
- For a row that is a placeholder in every layer, Java preserves `_ROW_ID` and sets `_SEQUENCE_NUMBER` to `-1`, while the C++ fallback nulls all columns of the bunch schema; this is only observable when a blob bunch serves system columns (blob-only splits).
- The in-band sentinel carries a theoretical collision risk for a user blob whose content equals the 9 sentinel bytes, the same class of risk as descriptor magic sniffing; Java's reference comparison is immune.
- Java applies deletion vectors to placeholder gap readers; the C++ data-evolution blob read path does not wire deletion vectors yet (pre-existing).
- Java also supports `ARRAY<BLOB>`/`MAP<_, BLOB>` placeholders; out of scope while paimon-cpp rejects nested blob fields.

SteNicholas:
### Purpose

Linked issue: close #452

A data-evolution partial update rewrites only the touched rows of a blob column and records every untouched row as a placeholder entry (`bin_length -2`, no data bytes). This PR makes the whole chain understand such entries.

One design note up front, since it is the main thing to review: Java identifies a placeholder by JVM reference identity on a singleton (`BlobPlaceholder.INSTANCE`), so a user blob can never be mistaken for one. That is not available here — values cross the C++ read/write chain as Arrow `large_binary` bytes, with no room for object identity — so placeholders are identified by exact byte equality with an internal reserved sentinel instead. The on-disk protocol (`bin_length -2`) stays identical to Java; the accepted collisions below are the price of that substitution.

- **Blob format, write side**: the placeholder protocol is opt-in via the internal format option `blob.internal.write-placeholder` (`BlobDefs::kWritePlaceholderKey`, `false` by default), enabled only for data-evolution partial updates, i.e. blob-only column writes of a table with data evolution enabled. Under the protocol, a value exactly equal to the internal reserved sentinel `_PAIMON_BLOB_PLACEHOLDER` (`BlobDefs::kPlaceholderSentinel`, exposed as `BlobDefs::PlaceholderSentinelView()`) is persisted as a `bin_length -2` entry without payload bytes; any other value is stored verbatim. Outside the protocol the writer never interprets values at all. The gate cannot distinguish a first write from an update, so a blob-only first write also runs under the sentinel channel.
- **Blob format, read side**: the reader fails on placeholder entries by default; the internal option `blob.internal.emit-placeholder-sentinel` (`BlobDefs::kEmitPlaceholderSentinelKey`, `false` by default, set only by the fallback read path) switches it to emit the sentinel bytes for `-2` entries, so after the batch has passed through schema-mapping readers, the fallback merge can identify placeholders by exact byte equality. Sentinel bytes are never stored in blob files. Unlike Java, whose blob reader always emits the placeholder singleton and only fails when a consumer touches it, the failure here happens at read time with an explicit message.
- **Accepted collisions**: there is no escaping anywhere in the chain, so a user blob whose bytes exactly equal the sentinel collides with the marker in two ways. Written through the partial-update channel, it is persisted as a placeholder entry that a single-layer read then rejects loudly, since no older layer can resolve it. Left untouched in an older layer under a later partial update, it reads as a placeholder in every layer and silently degrades to a null blob. The sentinel is distinctive enough that both are accepted as negligibly improbable; they are documented on `BlobDefs::kPlaceholderSentinel` and pinned by tests. Outside the two internal channels values are never interpreted, so a sentinel-equal blob round-trips verbatim through a normal write and read.
- **Internal channels are not user-reachable**: both `blob.internal.*` keys are stripped from user-supplied table options on the write path (`BlobDataFileWriterFactory::CreateWriter`) and on the read path (`AbstractSplitRead::PrepareReaderBuilder`) before each applies its own decision, so setting them in table options has no effect and only the internal data-evolution paths can enable the protocol.
- **`BlobBunch`**: keeps all blob files of a bunch, where files sharing a `max_sequence_number` form one non-overlapping layer; overlaps across layers are the expected shape of partial updates. When every file shares a single `max_sequence_number`, `SequentialReadOptimize()` keeps the existing concat fast path. This mirrors Java's `BlobFileBunch`, which likewise retains every file rather than filtering by max sequence number.
- **`BlobFallbackBatchReader`** (new, aligned with Java's `BlobFallbackRecordReader` / `AllPlaceholdersRecordReader`): merges the layers of one blob bunch ordered by `max_sequence_number` descending; row-id ranges a layer does not cover are padded with placeholder gap segments so all layers step in lockstep; each output row takes the newest non-placeholder layer, and an explicitly written null wins over older layers. A row that is a placeholder in every layer degrades to a null blob while keeping its `_ROW_ID` and reporting `-1` as its `_SEQUENCE_NUMBER`; resolved rows report their layer's sequence number. Row-range pushdown is honored per layer, including inside gaps. Unlike Java's `ForceSingleBatchReader`, the layers are stepped in a `read_batch_size` window rather than materialized as a single batch.
- **`BlobFileBatchReader`**: `GetPreviousBatchFileRowId` now stays usable when a selection bitmap removed rows, mapping batch positions back to original file row indexes through `target_blob_row_indexes_` — required by `CompleteRowTrackingFieldsBatchReader` to synthesize `_ROW_ID` under row-range pushdown.
- **`CompleteRowTrackingFieldsBatchReader`**: takes an explicit optional `file_field_names` (from `DataFileMeta::write_cols`) declaring the physical fields of a format without a self-describing file schema (blob); self-describing formats keep querying `GetFileSchema()`. Unrelated `NotImplemented` errors stay visible instead of being treated as an empty schema.
- **`DataEvolutionSplitRead`**: builds the fallback reader when a blob bunch spans multiple sequence layers, and passes the internal format options through `AbstractSplitRead::CreateRawFileReaders` (explicit `extra_format_options` parameter, no default arguments).
- **`DataEvolutionFileReader`**: `Create` accepts a single inner reader — a blob-only write collapses every layer of a merge split into one blob bunch, and the degenerate union still maps the bunch's fields into read schema order and null fills unmatched read fields. Reader offsets are now validated against the reader count instead of requiring at least two readers. Java relaxed the same check from `> 1` to `>= 1` in the commit that introduced the placeholder protocol.
- **`DataEvolutionBatchScan`**: `WrapToIndexedSplits` derived a split's row id range from its first and last file only. That bounding range relies on an invariant that does not hold — the row-id ranges of the files in a split may be unordered, discontiguous or overlapping, since `DataEvolutionSplitGenerator` can bin-pack disjoint groups into one split — so it both admitted row ids no file of the split covers and, for a partial-update blob layer covering a strict subrange, truncated the range and silently dropped requested row ranges at scan time. The index is now intersected with each file's row-id range separately and the results are sorted and merged (`Range::SortAndMergeOverlap(..., adjacent=true)`), aligning with Java's `DataEvolutionBatchScan.wrap()`. `WrapToIndexedSplits` becomes `static` and visible for testing.

Not covered, unchanged by this PR: deletion vectors are rejected outright by `DataEvolutionSplitRead`, so unlike Java's fallback reader — which applies a deletion vector to both file segments and placeholder gaps — the fallback path here has no deletion-vector handling to align yet.

### Tests

- `blob_format_writer_test.cpp`: placeholder golden bytes aligned with the Java writer, strict vs placeholder-aware read modes (including descriptor mode and selection bitmaps), and sentinel-equal / sentinel-prefixed user bytes stored verbatim (`TestWritePlaceholderGoldenBytes`, `TestReadPlaceholderStrictAndAwareModes`, `TestReadPlaceholderWithSelectionBitmap`, `TestSentinelBytesVerbatimWithoutPlaceholderMode`, `TestSentinelPrefixedValueVerbatimInPlaceholderMode`).
- `blob_file_batch_reader_test.cpp`: batch positions map back to original file row indexes under a selection bitmap (`TestRowNumbersWithSelectionBitmap`).
- `complete_row_tracking_fields_reader_test.cpp`: declared physical fields take precedence over the inner reader's self-described schema (`TestSetReadSchemaWithDeclaredFileFields`); existing cases keep exercising the self-describing branch.
- `blob_fallback_batch_reader_test.cpp` (new): fallback merge across layers under batch-size sweeps — gaps (leading/middle/trailing, disjoint ranges), all-placeholder rows degrading to null, null-wins semantics, sentinel-prefixed pass-through, three layers, row-tracking projections including the all-placeholder `-1` semantics, and misaligned-group / creation failures.
- `data_evolution_batch_scan_test.cpp` (new): `WrapToIndexedSplits` over a split whose files have unordered and discontiguous row-id ranges (mirroring Java's `testWrapToIndexSplitsWithUnorderedAndDiscontiguousDataFiles`), row ids falling into the gaps between file ranges being excluded, and the no-intersection failure (`TestWrapToIndexedSplitsWithUnorderedAndDiscontiguousDataFiles`, `TestWrapToIndexedSplitsExcludesRowIdsInFileRangeGaps`, `TestWrapToIndexedSplitsWithoutIntersection`).
- `data_evolution_split_read_test.cpp`: `BlobBunch` keeps all sequence layers, rejects overlaps within one layer, and reports `RowCount` / `SequentialReadOptimize` accordingly.
- `data_evolution_file_reader_test.cpp`: creation rejects empty readers and reader offsets referencing a missing reader, replacing the old at-least-two-readers requirement (`TestInvalid`).
- `blob_table_inte_test.cpp` (IT): end-to-end partial-update reads — placeholder fallback with null overwrite and a `blob_as_descriptor` variant, multi-layer and compacted-layer layouts (mirroring Java's `BlobUpdateTest.testReadCompactedBlobSequenceGroups`) with per-row and gap-crossing row-range reads, row-range pushdown over updated and untouched rows, a subrange sequence layer under the row-tracking projection with row-range pushdown, and an all-placeholder row keeping its `_ROW_ID` with `_SEQUENCE_NUMBER -1` (`TestDataEvolutionBlobPartialUpdate*`). The row-tracking tests write only the blob column, so the split contains only blob files and the blob bunch itself carries the row-tracking fields.
- `blob_table_inte_test.cpp` (IT), sentinel handling: a sentinel-equal user blob written through a normal full-row write round-trips verbatim, including through a later partial-update fallback (`TestBlobValueEqualToPlaceholderSentinelBytes`); the accepted collision where such a value sits in an older layer under a later partial update and degrades to a null blob is pinned so the behavior cannot change silently (`TestBlobSentinelValueInBaseLayerDegradesToNull`); and `blob.internal.*` keys set in user table options are ignored on both paths, which would otherwise make the table unreadable (`TestUserSuppliedInternalPlaceholderOptionsIgnored`).

### API and Format

- No changes under `include/paimon`.
- Blob file format: `bin_length -2` marks a placeholder entry occupying no file space, aligned with Java's `BlobFormatWriter.PLACE_HOLDER_LENGTH`. Readers without placeholder support already reject such entries with an explicit error, and files without placeholders are unaffected.
- The in-repo blob format description (`blob_file_batch_reader.h`) now documents both special index values: `-1` (null) and `-2` (placeholder).
- `_PAIMON_BLOB_PLACEHOLDER` is an internal reserved value only inside the two internal placeholder channels; normal writes and reads never interpret it.
- Two new internal (non user-facing) format options, both `false` by default: `blob.internal.write-placeholder` (set only for data-evolution blob-only column writes) and `blob.internal.emit-placeholder-sentinel` (set only by the data-evolution fallback read path). Neither is settable by users: both keys are stripped from user-supplied table options on the write and read paths. Java needs no equivalent option, since its placeholder travels as an object rather than as a format-level flag.

### Documentation

Internal read/write behavior of data-evolution partial updates; no user-facing documentation change.

### Generative AI tooling

### DIFF
--- a/src/paimon/common/data/blob_defs.h
+++ b/src/paimon/common/data/blob_defs.h
@@ -17,6 +17,10 @@
 #pragma once
 
 #include <cstdint>
+#include <cstring>
+#include <map>
+#include <string>
+#include <string_view>
 
 namespace paimon {
 
@@ -45,6 +49,64 @@ class BlobDefs {
 
     /// A bin_length value of -1 in the index indicates a null blob entry.
     static constexpr int64_t kNullBinLength = -1;
+    /// A bin_length value of -2 in the index indicates a placeholder blob entry, written by
+    /// data-evolution partial updates for rows whose blob value is not updated. A placeholder
+    /// entry occupies no file space; readers must fall back to an older blob file covering the
+    /// same row to resolve the value. Aligned with Java's BlobFormatWriter.PLACE_HOLDER_LENGTH.
+    static constexpr int64_t kPlaceholderBinLength = -2;
+    /// Sentinel bytes standing for a placeholder blob value in two internal channels:
+    ///
+    /// - Write channel: a data-evolution partial update (a blob-only column write, see
+    ///   kWritePlaceholderKey) marks a not-updated row with these bytes, and the blob format
+    ///   writer persists it as a bin_length -2 entry. Use PlaceholderSentinelView() to build
+    ///   such write arrays. Outside that mode the writer never interprets values, so arbitrary
+    ///   user bytes can never be turned into a placeholder entry.
+    /// - Read channel: a placeholder-aware reader (see kEmitPlaceholderSentinelKey) emits these
+    ///   bytes for -2 entries so the fallback merge can identify placeholders after the batch
+    ///   has passed through schema-mapping readers.
+    ///
+    /// Both channels identify a placeholder by exact byte equality with this internal reserved
+    /// value (IsPlaceholderSentinel), and the fallback merge byte-compares every layer of a
+    /// bunch — including files written outside the write channel. A user blob whose bytes
+    /// exactly equal the marker therefore collides with it in two ways: written through the
+    /// partial-update channel it is persisted as a placeholder entry, which a single-layer read
+    /// rejects loudly (no older layer can resolve it); left untouched in an older layer under a
+    /// later partial update it reads as a placeholder in every layer and silently degrades to a
+    /// null blob. The marker is distinctive enough that these collisions are accepted as
+    /// negligibly improbable. Sentinel bytes are never stored in blob files.
+    static constexpr char kPlaceholderSentinel[] = "_PAIMON_BLOB_PLACEHOLDER";
+    /// Byte length of kPlaceholderSentinel, excluding the literal's terminating NUL.
+    static constexpr int32_t kPlaceholderSentinelLength = sizeof(kPlaceholderSentinel) - 1;
+    /// Internal (non user-facing) format option, "false" by default: when "true", the blob
+    /// reader emits kPlaceholderSentinel for placeholder entries instead of failing on them.
+    /// Only the data-evolution blob fallback read path sets this.
+    static constexpr char kEmitPlaceholderSentinelKey[] = "blob.internal.emit-placeholder-sentinel";
+    /// Internal (non user-facing) format option, "false" by default: when "true", the blob
+    /// format writer persists a value exactly equal to kPlaceholderSentinel as a bin_length -2
+    /// entry. Only set for data-evolution partial updates, i.e. blob-only column writes of a
+    /// table with data evolution enabled; all other writes store bytes verbatim.
+    static constexpr char kWritePlaceholderKey[] = "blob.internal.write-placeholder";
+
+    /// The sentinel bytes for building a data-evolution partial-update write array: a row equal
+    /// to this view is persisted as a placeholder entry (see kWritePlaceholderKey).
+    static std::string_view PlaceholderSentinelView() {
+        return {kPlaceholderSentinel, static_cast<size_t>(kPlaceholderSentinelLength)};
+    }
+
+    /// True when the bytes are exactly the placeholder sentinel.
+    static bool IsPlaceholderSentinel(const char* data, size_t size) {
+        return size == static_cast<size_t>(kPlaceholderSentinelLength) &&
+               memcmp(data, kPlaceholderSentinel, kPlaceholderSentinelLength) == 0;
+    }
+
+    /// Removes the internal placeholder option keys from a format options map. The placeholder
+    /// channels must only ever be enabled by the internal data-evolution write and read paths,
+    /// so every consumer building format options from user-supplied table options strips these
+    /// keys before applying its own decision.
+    static void EraseInternalPlaceholderOptions(std::map<std::string, std::string>* options) {
+        options->erase(kEmitPlaceholderSentinelKey);
+        options->erase(kWritePlaceholderKey);
+    }
     /// Blob file format version.
     static constexpr int8_t kFileVersion = 1;
     /// Magic number identifying the start of each blob bin.
--- a/src/paimon/common/reader/blob_fallback_batch_reader.cpp
+++ b/src/paimon/common/reader/blob_fallback_batch_reader.cpp
@@ -1,0 +1,403 @@
+/*
+ * Copyright 2026-present Alibaba Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "paimon/common/reader/blob_fallback_batch_reader.h"
+
+#include <algorithm>
+#include <string_view>
+#include <utility>
+
+#include "arrow/api.h"
+#include "arrow/c/bridge.h"
+#include "fmt/format.h"
+#include "paimon/common/data/blob_defs.h"
+#include "paimon/common/data/blob_utils.h"
+#include "paimon/common/metrics/metrics_impl.h"
+#include "paimon/common/reader/reader_utils.h"
+#include "paimon/common/table/special_fields.h"
+#include "paimon/common/utils/arrow/mem_utils.h"
+#include "paimon/common/utils/arrow/status_utils.h"
+
+namespace paimon {
+
+Result<std::unique_ptr<BlobFallbackBatchReader>> BlobFallbackBatchReader::Create(
+    std::vector<std::vector<Segment>>&& sequence_groups,
+    const std::shared_ptr<arrow::Schema>& read_schema, int32_t read_batch_size,
+    const std::shared_ptr<MemoryPool>& pool) {
+    if (sequence_groups.size() < 2) {
+        return Status::Invalid(
+            "Blob fallback needs at least two sequence groups; a single group should be read "
+            "sequentially.");
+    }
+    if (read_schema == nullptr) {
+        return Status::Invalid("Blob fallback read schema cannot be nullptr.");
+    }
+    if (read_batch_size <= 0) {
+        return Status::Invalid(fmt::format(
+            "Blob fallback read batch size '{}' should be larger than zero", read_batch_size));
+    }
+    int32_t blob_field_idx = -1;
+    for (int32_t i = 0; i < read_schema->num_fields(); i++) {
+        if (BlobUtils::IsBlobField(read_schema->field(i))) {
+            if (blob_field_idx != -1) {
+                return Status::Invalid(
+                    "Blob fallback read schema should contain exactly one blob field.");
+            }
+            blob_field_idx = i;
+        }
+    }
+    if (blob_field_idx == -1) {
+        return Status::Invalid("Blob fallback read schema should contain a blob field.");
+    }
+    int32_t row_id_field_idx = read_schema->GetFieldIndex(SpecialFields::RowId().Name());
+    int32_t seq_num_field_idx = read_schema->GetFieldIndex(SpecialFields::SequenceNumber().Name());
+    std::vector<GroupCursor> groups;
+    groups.reserve(sequence_groups.size());
+    for (auto& segments : sequence_groups) {
+        if (segments.empty()) {
+            return Status::Invalid("Blob fallback sequence group should not be empty.");
+        }
+        for (const auto& segment : segments) {
+            if (segment.reader == nullptr && segment.gap_selected_ranges.empty()) {
+                return Status::Invalid(
+                    "Blob fallback gap segment should cover at least one selected row id.");
+            }
+        }
+        GroupCursor cursor;
+        cursor.segments = std::move(segments);
+        groups.push_back(std::move(cursor));
+    }
+    return std::unique_ptr<BlobFallbackBatchReader>(
+        new BlobFallbackBatchReader(std::move(groups), read_schema, blob_field_idx,
+                                    row_id_field_idx, seq_num_field_idx, read_batch_size, pool));
+}
+
+BlobFallbackBatchReader::BlobFallbackBatchReader(std::vector<GroupCursor>&& groups,
+                                                 const std::shared_ptr<arrow::Schema>& read_schema,
+                                                 int32_t blob_field_idx, int32_t row_id_field_idx,
+                                                 int32_t seq_num_field_idx, int32_t read_batch_size,
+                                                 const std::shared_ptr<MemoryPool>& pool)
+    : groups_(std::move(groups)),
+      read_schema_(read_schema),
+      blob_field_idx_(blob_field_idx),
+      row_id_field_idx_(row_id_field_idx),
+      seq_num_field_idx_(seq_num_field_idx),
+      read_batch_size_(read_batch_size),
+      arrow_pool_(GetArrowPool(pool)) {}
+
+Result<int64_t> BlobFallbackBatchReader::FillWindow(size_t group_idx, int64_t want,
+                                                    std::vector<Chunk>* chunks) {
+    GroupCursor& cursor = groups_[group_idx];
+    int64_t collected = 0;
+    while (collected < want) {
+        if (!cursor.pending.empty()) {
+            const std::shared_ptr<arrow::StructArray>& front = cursor.pending.front();
+            int64_t available = front->length() - cursor.pending_pos;
+            int64_t take = std::min(available, want - collected);
+            chunks->push_back(Chunk{front, cursor.pending_pos, take, {}});
+            cursor.pending_pos += take;
+            collected += take;
+            if (cursor.pending_pos == front->length()) {
+                cursor.pending.pop_front();
+                cursor.pending_pos = 0;
+            }
+            continue;
+        }
+        if (cursor.segment_idx >= cursor.segments.size()) {
+            // group exhausted; only the first group may define a shorter window
+            break;
+        }
+        Segment& segment = cursor.segments[cursor.segment_idx];
+        if (segment.reader == nullptr) {
+            // gap segment: all rows are placeholders, stepped range by range so the row ids
+            // stay available for all-placeholder rows
+            if (cursor.gap_range_idx >= segment.gap_selected_ranges.size()) {
+                cursor.segment_idx++;
+                cursor.gap_range_idx = 0;
+                cursor.gap_range_pos = 0;
+                continue;
+            }
+            const Range& range = segment.gap_selected_ranges[cursor.gap_range_idx];
+            int64_t remaining = range.Count() - cursor.gap_range_pos;
+            int64_t take = std::min(remaining, want - collected);
+            Chunk chunk{nullptr, 0, take, {}};
+            if (row_id_field_idx_ >= 0) {
+                chunk.gap_row_ids.reserve(take);
+                for (int64_t k = 0; k < take; k++) {
+                    chunk.gap_row_ids.push_back(range.from + cursor.gap_range_pos + k);
+                }
+            }
+            chunks->push_back(std::move(chunk));
+            cursor.gap_range_pos += take;
+            collected += take;
+            if (cursor.gap_range_pos == range.Count()) {
+                cursor.gap_range_idx++;
+                cursor.gap_range_pos = 0;
+            }
+            continue;
+        }
+        PAIMON_ASSIGN_OR_RAISE(ReadBatchWithBitmap batch_with_bitmap,
+                               segment.reader->NextBatchWithBitmap());
+        if (BatchReader::IsEofBatch(batch_with_bitmap)) {
+            cursor.segment_idx++;
+            continue;
+        }
+        auto& [read_batch, bitmap] = batch_with_bitmap;
+        auto& [c_array, c_schema] = read_batch;
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Array> src_array,
+                                          arrow::ImportArray(c_array.get(), c_schema.get()));
+        PAIMON_ASSIGN_OR_RAISE(arrow::ArrayVector selected_array_vec,
+                               ReaderUtils::GenerateFilteredArrayVector(src_array, bitmap));
+        for (const auto& selected_array : selected_array_vec) {
+            if (selected_array->length() == 0) {
+                continue;
+            }
+            auto struct_array = std::dynamic_pointer_cast<arrow::StructArray>(selected_array);
+            if (struct_array == nullptr) {
+                return Status::Invalid("Blob fallback expects file readers to emit struct arrays.");
+            }
+            cursor.pending.push_back(std::move(struct_array));
+        }
+    }
+    return collected;
+}
+
+Result<std::vector<bool>> BlobFallbackBatchReader::ComputePlaceholderFlags(
+    const std::vector<Chunk>& chunks, int64_t row_count) const {
+    std::vector<bool> flags(row_count, false);
+    int64_t pos = 0;
+    for (const auto& chunk : chunks) {
+        if (chunk.array == nullptr) {
+            // gap rows stand for placeholders
+            std::fill(flags.begin() + pos, flags.begin() + pos + chunk.length, true);
+        } else {
+            std::shared_ptr<arrow::Array> blob_col = chunk.array->field(blob_field_idx_);
+            auto binary_col = std::dynamic_pointer_cast<arrow::LargeBinaryArray>(blob_col);
+            if (binary_col == nullptr) {
+                return Status::Invalid(fmt::format(
+                    "Blob fallback expects the blob column to be large binary, but got {}",
+                    blob_col->type()->ToString()));
+            }
+            for (int64_t k = 0; k < chunk.length; k++) {
+                int64_t idx = chunk.offset + k;
+                if (binary_col->IsNull(idx)) {
+                    continue;
+                }
+                std::string_view value = binary_col->GetView(idx);
+                if (BlobDefs::IsPlaceholderSentinel(value.data(), value.size())) {
+                    flags[pos + k] = true;
+                }
+            }
+        }
+        pos += chunk.length;
+    }
+    return flags;
+}
+
+Result<std::shared_ptr<arrow::Array>> BlobFallbackBatchReader::AssembleRowIdRun(
+    const std::vector<Chunk>& chunks, int64_t run_start, int64_t run_end) const {
+    arrow::ArrayVector pieces;
+    int64_t pos = 0;
+    for (const auto& chunk : chunks) {
+        int64_t overlap_start = std::max(run_start, pos);
+        int64_t overlap_end = std::min(run_end, pos + chunk.length);
+        if (overlap_start < overlap_end) {
+            if (chunk.array != nullptr) {
+                std::shared_ptr<arrow::Array> column = chunk.array->field(row_id_field_idx_);
+                pieces.push_back(column->Slice(chunk.offset + (overlap_start - pos),
+                                               overlap_end - overlap_start));
+            } else {
+                arrow::Int64Builder builder(arrow_pool_.get());
+                for (int64_t r = overlap_start; r < overlap_end; r++) {
+                    PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.Append(chunk.gap_row_ids[r - pos]));
+                }
+                std::shared_ptr<arrow::Array> piece;
+                PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.Finish(&piece));
+                pieces.push_back(std::move(piece));
+            }
+        }
+        pos += chunk.length;
+        if (pos >= run_end) {
+            break;
+        }
+    }
+    if (pieces.size() == 1 && pieces[0]->offset() == 0) {
+        return pieces[0];
+    }
+    // Concatenate flattens non-zero offsets left by Slice, so the exported batch honors the
+    // zero-offset BatchReader contract.
+    PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Array> concat_array,
+                                      arrow::Concatenate(pieces, arrow_pool_.get()));
+    return concat_array;
+}
+
+Result<std::shared_ptr<arrow::Array>> BlobFallbackBatchReader::AssembleColumn(
+    int32_t field_idx, const std::vector<int32_t>& group_choice,
+    const std::vector<std::vector<Chunk>>& group_chunks) const {
+    const auto row_count = static_cast<int64_t>(group_choice.size());
+    arrow::ArrayVector pieces;
+    int64_t run_start = 0;
+    while (run_start < row_count) {
+        const int32_t group = group_choice[run_start];
+        int64_t run_end = run_start + 1;
+        while (run_end < row_count && group_choice[run_end] == group) {
+            run_end++;
+        }
+        if (group < 0) {
+            // placeholder in every layer: the blob degrades to null; the row keeps its row id
+            // (taken from the newest group, which steps in lockstep), reports -1 as its
+            // sequence number, and returns null for every other field
+            if (field_idx == row_id_field_idx_) {
+                PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Array> row_id_piece,
+                                       AssembleRowIdRun(group_chunks[0], run_start, run_end));
+                pieces.push_back(std::move(row_id_piece));
+            } else if (field_idx == seq_num_field_idx_) {
+                arrow::Int64Scalar seq_scalar(-1);
+                PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+                    std::shared_ptr<arrow::Array> seq_piece,
+                    arrow::MakeArrayFromScalar(seq_scalar, run_end - run_start, arrow_pool_.get()));
+                pieces.push_back(std::move(seq_piece));
+            } else {
+                PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+                    std::shared_ptr<arrow::Array> null_piece,
+                    arrow::MakeArrayOfNull(read_schema_->field(field_idx)->type(),
+                                           run_end - run_start, arrow_pool_.get()));
+                pieces.push_back(std::move(null_piece));
+            }
+        } else {
+            int64_t pos = 0;
+            for (const auto& chunk : group_chunks[group]) {
+                int64_t overlap_start = std::max(run_start, pos);
+                int64_t overlap_end = std::min(run_end, pos + chunk.length);
+                if (overlap_start < overlap_end) {
+                    if (chunk.array == nullptr) {
+                        return Status::Invalid(
+                            "Unexpected: a gap row was chosen as a blob fallback result.");
+                    }
+                    std::shared_ptr<arrow::Array> column = chunk.array->field(field_idx);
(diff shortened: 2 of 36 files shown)


src/paimon/common/data/blob_defs.h
```diff
@@ -45,6 +46,28 @@ class BlobDefs {
...
+    /// BlobDescriptor is detected by its magic header; a placeholder-aware reader (see
+    /// kEmitPlaceholderSentinelKey) emits them for placeholder entries so that fallback merging
+    /// can still identify placeholders after the batch has passed through schema-mapping
+    /// readers. Layout: version(1) + magic "BLOBPLHD"(8).
+    static constexpr char kPlaceholderSentinel[] = {0x01, 'B', 'L', 'O', 'B', 'P', 'L', 'H', 'D'};
+    static constexpr int32_t kPlaceholderSentinelLength = 9;
+    /// Internal (non user-facing) format option: when "true", the blob reader emits
+    /// kPlaceholderSentinel for placeholder entries instead of failing on them. Only the
+    /// data-evolution blob fallback read path sets this.
+    static constexpr char kEmitPlaceholderSentinelKey[] = "blob.internal.emit-placeholder-sentinel";
```
lxy-9602, line 83: Could you please mention in the comment that `kEmitPlaceholderSentinelKey` is false by default?

src/paimon/common/reader/blob_fallback_batch_reader.h
```diff
@@ -0,0 +1,140 @@
+/*
+ * Copyright 2024-present Alibaba Inc.
+ *
```
lxy-9602, line 3: Copyright 2026

src/paimon/common/reader/blob_fallback_batch_reader.cpp
```diff
@@ -0,0 +1,325 @@
...
+            }
+        }
+    }
+
+    arrow::ArrayVector columns;
+    columns.reserve(read_schema_->num_fields());
+    for (int32_t field_idx = 0; field_idx < read_schema_->num_fields(); field_idx++) {
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Array> column,
+                               AssembleColumn(field_idx, group_choice, group_chunks));
+        columns.push_back(std::move(column));
```
lxy-9602, line 357: Please preserve row-tracking fields for all-placeholder rows.

`Create()` only requires the read schema to contain exactly one BLOB field, so the schema may also contain `_ROW_ID` or `_SEQUENCE_NUMBER`. When every sequence group contains a placeholder for a row, this branch creates a null array for every field, which would also null those row-tracking fields.

The per-file `CompleteRowTrackingFieldsBatchReader` runs before this fallback reader, and there is no completion reader after it to repair the values. Synthetic gap segments do not have a file reader either. Java's `BlobFallbackRecordReader.nullBlobRow` instead preserves `_ROW_ID` and sets `_SEQUENCE_NUMBER` to `-1`.

If the intended invariant is that this reader always receives a BLOB-only schema, please enforce that in `Create()` and ensure field routing guarantees it. Otherwise, this branch should synthesize the row ID and sequence number.

Please also add an integration test for the placeholder fallback path that projects exactly:

    {"b0", "_ROW_ID", "_SEQUENCE_NUMBER"}

The test should include a row for which every BLOB layer is a placeholder, so that the `group < 0` branch is exercised. It should verify that the BLOB result is null, `_ROW_ID` remains correct, and `_SEQUENCE_NUMBER` follows the intended semantics (Java uses `-1` for an all-placeholder fallback row). This would also validate whether the production field-routing path gives `BlobFallbackBatchReader` a BLOB-only schema or includes the row-tracking fields.

src/paimon/common/reader/blob_fallback_batch_reader_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+TEST_F(BlobFallbackBatchReaderTest, TestMisalignedGroupsFail) {
+    std::vector<std::vector<BlobFallbackBatchReader::Segment>> groups;
+    groups.push_back(MakeGroup({SegmentSpec::File({"PH", "u1", "PH"})}, 1024));
+    groups.push_back(MakeGroup({SegmentSpec::File({"b0", "b1"})}, 1024));
+    ASSERT_OK_AND_ASSIGN(
+        auto reader, BlobFallbackBatchReader::Create(std::move(groups), read_schema_, 1024, pool_));
+    Result<BatchReader::ReadBatch> batch = reader->NextBatch();
+    ASSERT_FALSE(batch.ok());
+    ASSERT_TRUE(batch.status().ToString().find("same number of rows") != std::string::npos);
```
lxy-9602: Please use `ASSERT_NOK_WITH_MSG`

src/paimon/core/operation/abstract_split_read.h
```diff
@@ -60,12 +61,16 @@ class AbstractSplitRead : public SplitRead {
...
+    /// reader, e.g. to switch the blob format reader into placeholder-aware mode for the
+    /// data-evolution blob fallback read path.
     Result<std::vector<std::unique_ptr<FileBatchReader>>> CreateRawFileReaders(
         const BinaryRow& partition, const std::vector<std::shared_ptr<DataFileMeta>>& data_files,
         const std::shared_ptr<arrow::Schema>& read_schema,
         const std::shared_ptr<Predicate>& predicate, DeletionVector::Factory dv_factory,
         const std::optional<std::vector<Range>>& row_ranges,
-        const std::shared_ptr<DataFilePathFactory>& data_file_path_factory) const;
+        const std::shared_ptr<DataFilePathFactory>& data_file_path_factory,
+        const std::map<std::string, std::string>& extra_format_options = {}) const;
```
lxy-9602, line 74: Please avoid default param in production src code.

src/paimon/common/data/blob_defs.h
```diff
@@ -45,6 +46,28 @@ class BlobDefs {
...
+
+    static bool IsPlaceholderSentinel(const char* data, size_t size) {
+        return size == static_cast<size_t>(kPlaceholderSentinelLength) &&
+               memcmp(data, kPlaceholderSentinel, kPlaceholderSentinelLength) == 0;
```
lxy-9602, line 99: Should include expose an interface for `kPlaceholderSentinel`? Otherwise, it’s not clear how the write path would construct the array containing it.

src/paimon/format/blob/blob_format_writer_test.cpp
```diff
@@ -931,4 +931,166 @@ TEST_P(BlobFormatWriterTest, TestAddBatchWithZeroLengthBlob) {
...
+        ASSERT_OK_AND_ASSIGN(std::unique_ptr<BlobFileBatchReader> reader,
+                             BlobFileBatchReader::Create(input_stream, /*batch_size=*/1024,
+                                                         /*blob_as_descriptor=*/false, pool_));
+        ::ArrowSchema c_schema;
+        ASSERT_TRUE(arrow::ExportSchema(*schema, &c_schema).ok());
+        ASSERT_OK(reader->SetReadSchema(&c_schema, /*predicate=*/nullptr,
+                                        /*selection_bitmap=*/std::nullopt));
+        Result<BatchReader::ReadBatch> batch = reader->NextBatch();
+        ASSERT_FALSE(batch.ok());
+        ASSERT_TRUE(batch.status().ToString().find("placeholder") != std::string::npos);
```
lxy-9602: ASSERT_NOK_WITH_MSG

src/paimon/format/blob/blob_file_batch_reader.h
```diff
@@ -87,9 +87,14 @@ namespace paimon::blob {
...
  public:
+    /// `emit_placeholder_sentinel` controls how placeholder entries (bin_length ==
+    /// BlobDefs::kPlaceholderBinLength) are read: when false (the default) they fail the read,
+    /// as resolving them requires the data-evolution blob fallback path; when true they are
+    /// returned as the non-null BlobDefs::kPlaceholderSentinel bytes for that path to merge away.
     static Result<std::unique_ptr<BlobFileBatchReader>> Create(
         const std::shared_ptr<InputStream>& input_stream, int32_t batch_size,
-        bool blob_as_descriptor, const std::shared_ptr<MemoryPool>& pool);
+        bool blob_as_descriptor, const std::shared_ptr<MemoryPool>& pool,
+        bool emit_placeholder_sentinel = false);
```
lxy-9602, line 107: Please avoid default param in src production code.

lxy-9602: Also please put `emit_placeholder_sentinel` before `pool` and make `pool` to be the last param.

test/inte/blob_table_inte_test.cpp
```diff
@@ -1045,6 +1046,311 @@ TEST_P(BlobTableInteTest, TestDataEvolutionBlobOnlyWriteWithFirstRowId) {
...
+        [](int32_t i) {
+            std::string blob = (i == 1 || i == 9) ? "\"update_" + std::to_string(i) + "\""
+                                                  : "\"blob_" + std::to_string(i) + "\"";
+            return std::to_string(i) + ", \"name_" + std::to_string(i) + "\", " + blob;
+        },
+        fields);
+    ASSERT_OK(ScanAndRead(table_path, schema->field_names(), expected_full));
+}
+
 TEST_P(BlobTableInteTest, TestDataEvolutionBlobOnlyFirstCommitFailsWithoutFirstRowId) {
```
lxy-9602, line 1694: Please add an integration test for a BLOB sequence layer whose physical file covers only a strict subrange of the full row-id range.

Suggested layout:

    old layer: max_seq=1, first_row_id=0, row_count=10
               [b0, b1, b2, b3, b4, b5, b6, b7, b8, b9]

    new layer: max_seq=2, first_row_id=2, row_count=2
               [u2, u3]

The newest group should therefore be represented internally as:

    Gap(2), File([u2, u3]), Gap(6)

and the result should be:

    [b0, b1, u2, u3, b4, b5, b6, b7, b8, b9]

Please preferably use a BLOB-only table and project exactly:

    {"b0", "_ROW_ID", "_SEQUENCE_NUMBER"}

The test should verify:

- `_ROW_ID` is `[0, ..., 9]`;
- rows 2 and 3 use the newer layer's sequence number;
- all other rows fall back to the older layer and use its sequence number;
- the leading and trailing gaps never expose the internal placeholder sentinel.

Java explicitly supports this layout by padding absent ranges with
`AllPlaceholdersRecordReader`. Its
`BlobFallbackRecordReaderTest.testBlobFallbackRecordReaderDerivesRowIdBoundsFromFiles`
already covers the equivalent shape (`old=[8,13]`, `new=[10,11]`), but this
exact layout is not currently covered by a C++ integration test.

The existing `BlobFallbackBatchReaderTest.TestGapPadding` validates the
reader in isolation. An integration test is still needed to validate the
`first_row_id` / `row_count` metadata, sequence grouping, synthetic gap
construction, field routing, and row-tracking values together.

src/paimon/common/data/blob_defs.h
```diff
@@ -45,6 +46,28 @@ class BlobDefs {
...
+    /// same row to resolve the value. Aligned with Java's BlobFormatWriter.PLACE_HOLDER_LENGTH.
+    static constexpr int64_t kPlaceholderBinLength = -2;
+    /// In-band sentinel bytes representing a placeholder blob value. The write path persists a
+    /// blob equal to these bytes as a placeholder entry, the same way a serialized
+    /// BlobDescriptor is detected by its magic header; a placeholder-aware reader (see
+    /// kEmitPlaceholderSentinelKey) emits them for placeholder entries so that fallback merging
+    /// can still identify placeholders after the batch has passed through schema-mapping
+    /// readers. Layout: version(1) + magic "BLOBPLHD"(8).
+    static constexpr char kPlaceholderSentinel[] = {0x01, 'B', 'L', 'O', 'B', 'P', 'L', 'H', 'D'};
+    static constexpr int32_t kPlaceholderSentinelLength = 9;
```
lxy-9602: Please avoid using an unescaped in-band BLOB value as the placeholder marker.

The BLOB type accepts arbitrary bytes, so a legitimate user value equal to
`kPlaceholderSentinel` is silently converted to `bin_length = -2` here and
will later fall back to an older value. The same collision also exists on
read: a real BLOB containing these bytes is classified as a placeholder by
`BlobFallbackBatchReader::ComputePlaceholderFlags`.

This differs from Java, where `BlobPlaceholder.INSTANCE` is identified by
object identity and cannot collide with normal BLOB data.

Using a more distinctive string such as `_PAIMON_BLOB_PLHD` would reduce the
probability of an accidental collision, but it would not make the encoding
safe because that string is still a valid BLOB value. Please use an
out-of-band placeholder marker, such as a dedicated bitmap/tag field, or
define an escaping protocol for legitimate values that match the reserved
marker.

Please also add a compatibility test that writes a normal BLOB whose bytes
exactly equal the marker and verifies that it is returned unchanged rather
than being persisted as `-2` or resolved through fallback.

src/paimon/common/data/blob_defs.h
SteNicholas: Done — the comment now states `kEmitPlaceholderSentinelKey` is "false" by default (same wording for the new `kWritePlaceholderKey`).

src/paimon/common/reader/blob_fallback_batch_reader.h
SteNicholas: Fixed — the new files now carry the 2026 copyright header.

src/paimon/common/reader/blob_fallback_batch_reader_test.cpp
SteNicholas: Done — `TestMisalignedGroupsFail` and `TestCreateValidation` now use `ASSERT_NOK_WITH_MSG`.

src/paimon/core/operation/abstract_split_read.h
SteNicholas: Done — `extra_format_options` is now a required parameter of `CreateRawFileReaders` / `PrepareReaderBuilder`; all callers pass it explicitly.

src/paimon/common/data/blob_defs.h
SteNicholas: Done — added `BlobDefs::PlaceholderSentinelView()` for producers to build partial-update write arrays (used by the integration tests), plus `IsPlaceholderSentinel` / `HasPlaceholderSentinelPrefix` for the two consuming ends.

src/paimon/format/blob/blob_format_writer_test.cpp
SteNicholas: Done — switched to `ASSERT_NOK_WITH_MSG`.

src/paimon/format/blob/blob_file_batch_reader.h
SteNicholas: Done on both — `emit_placeholder_sentinel` is now a required parameter placed before `pool`, and `pool` is the last parameter.

src/paimon/common/reader/blob_fallback_batch_reader.cpp
SteNicholas: Addressed — the fallback reader now resolves the row-tracking fields instead of nulling them: the read schema may contain `_ROW_ID` / `_SEQUENCE_NUMBER` (completed per file by `CompleteRowTrackingFieldsBatchReader`), an all-placeholder row keeps its `_ROW_ID` (taken from the newest group; gap segments synthesize the row ids of the ranges they cover) and reports `-1` as its `_SEQUENCE_NUMBER`, matching Java's `BlobFallbackRecordReader.nullBlobRow`; resolved rows report their layer's sequence number.

Tests: `BlobFallbackBatchReaderTest.TestRowTrackingFieldsPreserved` covers the `{blob, _ROW_ID, _SEQUENCE_NUMBER}`, `{blob, _ROW_ID}` and `{blob, _SEQUENCE_NUMBER}` projections including the all-placeholder `-1` case, and the integration test `TestDataEvolutionBlobPartialUpdateAllPlaceholderRowTracking` projects exactly `{b0, _ROW_ID, _SEQUENCE_NUMBER}` with a row that is a placeholder in every layer, verifying null blob, correct `_ROW_ID` and `_SEQUENCE_NUMBER == -1` end to end.

test/inte/blob_table_inte_test.cpp
SteNicholas: Done — added `TestDataEvolutionBlobPartialUpdateRowTrackingWithSubrangeLayer` with exactly the suggested layout: a blob-only table, old layer `max_seq=1, first_row_id=0, row_count=10`, new layer `first_row_id=2, row_count=2` (internally `Gap(2), File([u2, u3]), Gap(6)`), projecting `{b0, _ROW_ID, _SEQUENCE_NUMBER}`. It verifies `_ROW_ID` is `[0..9]`, rows 2-3 report the newer layer's sequence number, all other rows fall back to the older layer's sequence number, and the leading/trailing gaps never expose the internal placeholder sentinel.

src/paimon/common/data/blob_defs.h
SteNicholas: Addressed with the two mechanisms you suggested, combined:

1. **Out-of-band gating on the write side**: the writer no longer interprets values by default. The placeholder protocol is enabled via the internal format option `BlobDefs::kWritePlaceholderKey`, set only for data-evolution blob-only column writes (`AppendOnlyWriter::CreateRollingBlobWriter`). Every other write stores arbitrary user bytes verbatim, so a legitimate value equal to the sentinel can never be persisted as `bin_length -2`.
2. **Escaping protocol for the two internal channels**: within the write protocol and within the placeholder-aware read mode, a real value starting with the sentinel travels with one extra sentinel copy prepended, and the consuming end strips it — the writer unescapes before storing, the placeholder-aware reader escapes stored values it reads back, and the fallback merge unescapes (`BlobDefs::HasPlaceholderSentinelPrefix`). `ComputePlaceholderFlags` therefore can no longer misclassify real data; sentinel bytes are never stored in blob files and never returned to users.

Compatibility tests: the integration test `TestBlobValueEqualToPlaceholderSentinelBytes` writes a normal BLOB whose bytes exactly equal the marker and verifies it is read back unchanged and still survives a later placeholder fallback; `TestSentinelBytesVerbatimWithoutPlaceholderMode` and `TestEscapedSentinelRoundTrip` cover the format layer (verbatim storage outside the protocol, escape round trip inside it).

src/paimon/core/operation/data_evolution_split_read.h
lxy-9602, line 69: Please add `BlobFallbackBatchReader` here.

src/paimon/common/data/blob_defs.h
```diff
@@ -45,6 +47,60 @@ class BlobDefs {
...
+    ///   such write arrays. Outside that mode the writer never interprets values, so arbitrary
+    ///   user bytes can never be turned into a placeholder entry.
+    /// - Read channel: a placeholder-aware reader (see kEmitPlaceholderSentinelKey) emits these
+    ///   bytes for -2 entries so the fallback merge can identify placeholders after the batch
+    ///   has passed through schema-mapping readers.
+    ///
+    /// Both channels escape legitimate values so user bytes can never collide with the
+    /// sentinel: a real value whose bytes start with the sentinel travels with one extra copy
+    /// of the sentinel prepended, and the consuming end strips it (an exact match is a
+    /// placeholder, a longer match is unescaped). Sentinel bytes are never stored in blob files
```
lxy-9602: The `S + S` escape/unescape protocol introduces quite a bit of complexity across the writer, `BlobFileBatchReader`, and `BlobFallbackBatchReader`, and the behavior is not easy to understand or maintain.

Could we simplify this by using a sufficiently distinctive fixed placeholder marker, for example `_PAIMON_BLOB_PLACEHOLDER`, and only checking for exact equality? This would remove the sentinel-prefix detection and the producer-side escape requirement. If we accept the negligible collision probability, the marker can be documented as an internal reserved value.

src/paimon/core/io/complete_row_tracking_fields_reader.cpp
lxy-9602, line 108: `BlobFileBatchReader` advertises `SupportPreciseBitmapSelection() == true`, so a row-range selection is applied directly inside the BLOB reader. However, `GetPreviousBatchFileRowId()` returns `NotImplemented` whenever the selection removes any rows. `CompleteRowTrackingFieldsBatchReader` relies on this method to synthesize `_ROW_ID`, so reading `{"b0", "_ROW_ID", "_SEQUENCE_NUMBER"}` with a proper subset of row ranges fails.

The reader already maintains `target_blob_row_indexes_`, which maps every selected position back to its original file row index. Please use that mapping for the previous batch instead of rejecting selected reads

src/paimon/core/io/complete_row_tracking_fields_reader.cpp
```diff
@@ -39,9 +39,18 @@ CompleteRowTrackingFieldsBatchReader::CompleteRowTrackingFieldsBatchReader(
...
+    Result<std::unique_ptr<::ArrowSchema>> file_schema_result = reader_->GetFileSchema();
+    if (file_schema_result.ok()) {
+        std::unique_ptr<::ArrowSchema> c_file_schema = std::move(file_schema_result).value();
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(file_schema, arrow::ImportSchema(c_file_schema.get()));
+    } else if (file_schema_result.status().IsNotImplemented()) {
+        file_schema = arrow::schema(arrow::FieldVector{});
+    } else {
+        return file_schema_result.status();
+    }
     PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Schema> arrow_schema,
```
lxy-9602, line 58: For this wrapper, we only need to determine whether `_ROW_ID` and `_SEQUENCE_NUMBER` are physically present. `DataFileMeta.write_cols` is sufficient for a BLOB file and pretending that `NotImplemented` means an empty schema.

Could we pass the BLOB file’s known physical field names from `file_meta.write_cols` into `CompleteRowTrackingFieldsBatchReader`? Self-describing formats can continue to query `GetFileSchema()`, while the BLOB path explicitly declares its physical fields. This keeps unrelated `NotImplemented` errors visible and avoids fabricating a full Arrow file schema from column names alone.

src/paimon/core/operation/data_evolution_split_read.h
SteNicholas: Done (2dc1ee0).

src/paimon/common/data/blob_defs.h
SteNicholas: Done — the sentinel is now `_PAIMON_BLOB_PLACEHOLDER`, checked by exact equality only; the producer-side escape requirement and the sentinel-prefix detection are removed from the writer, `BlobFileBatchReader` and `BlobFallbackBatchReader`. The accepted collision probability is documented on `BlobDefs::kPlaceholderSentinel` as an internal reserved value (2dc1ee0).

src/paimon/core/io/complete_row_tracking_fields_reader.cpp
SteNicholas: Done — `GetPreviousBatchFileRowId` now tracks the previous batch's start position in the selected index space and maps it back to the original file row index through `target_blob_row_indexes_`, so selected reads work. Covered by `TestRowNumbersWithSelectionBitmap` and by the row-range section of `TestDataEvolutionBlobPartialUpdateRowTrackingWithSubrangeLayer` (2dc1ee0).

src/paimon/core/io/complete_row_tracking_fields_reader.cpp
SteNicholas: Done — `CompleteRowTrackingFieldsBatchReader` now takes an optional `file_field_names`; the blob path passes `DataFileMeta::write_cols`, self-describing formats keep querying `GetFileSchema()`, and unrelated `NotImplemented` errors propagate instead of being treated as an empty schema. Covered by `TestSetReadSchemaWithDeclaredFileFields` (2dc1ee0).

src/paimon/core/table/source/data_evolution_batch_scan.cpp
```diff
@@ -87,8 +89,12 @@ Result<std::shared_ptr<Plan>> DataEvolutionBatchScan::WrapToIndexedSplits(
...
         }
         PAIMON_ASSIGN_OR_RAISE(int64_t min, files[0]->NonNullFirstRowId());
-        PAIMON_ASSIGN_OR_RAISE(int64_t max, files[files.size() - 1]->NonNullFirstRowId());
-        max += files[files.size() - 1]->row_count - 1;
+        int64_t max = min + files[0]->row_count - 1;
+        for (size_t i = 1; i < files.size(); i++) {
+            PAIMON_ASSIGN_OR_RAISE(int64_t first_row_id, files[i]->NonNullFirstRowId());
+            min = std::min(min, first_row_id);
+            max = std::max(max, first_row_id + files[i]->row_count - 1);
+        }
```
lxy-9602, line 106: This relies on an invariant that does not hold: the row-id ranges covered by the files in a DataSplit are not guaranteed to be contiguous (or ordered).

Row IDs are allocated globally. For example, after multiple commits, one partition may contain files covering [0, 9] and [20, 29], while [10, 19] belongs to another partition. DataEvolutionSplitGenerator first creates separate overlapping-range groups, but its subsequent size-based bin packing may place multiple disjoint groups into the same DataSplit.

Using the bounding range [min, max] would therefore include row IDs not covered by this split. For example, intersecting [0, 29] with index results {5, 15, 25} returns 15 even though this split only covers [0, 9] and [20, 29].

Could we align this with Java DataEvolutionBatchScan.wrap(), which intersects the RowRangeIndex with each file's row-id range separately and then sorts and merges the resulting ranges?

Please also add a regression test where one DataSplit contains unordered and discontiguous file ranges, similar to Java's testWrapToIndexSplitsWithUnorderedAndDiscontiguousDataFiles.

SteNicholas: Done — aligned with Java's `DataEvolutionBatchScan.wrap()`: the `RowRangeIndex` is now intersected with each file's row-id range separately, and the resulting ranges are sorted and merged via `Range::SortAndMergeOverlap(..., adjacent=true)`. The bounding `[min, max]` is still computed, but only for the "no intersected ranges" error message.

Regression tests in the new `data_evolution_batch_scan_test.cpp`:

- `TestWrapToIndexedSplitsWithUnorderedAndDiscontiguousDataFiles` — a split whose files are unordered and discontiguous, mirroring Java's `testWrapToIndexSplitsWithUnorderedAndDiscontiguousDataFiles`.
- `TestWrapToIndexedSplitsExcludesRowIdsInFileRangeGaps` — exactly the example from your comment: a split covering `[0, 9]` and `[20, 29]` against index hits `{5, 15, 25}` now yields `{[5, 5], [25, 25]}`, dropping row id 15 and its score instead of returning it.
- `TestWrapToIndexedSplitsWithoutIntersection` — the no-intersection failure path.

`WrapToIndexedSplits` is now `static` and visible for testing so the above can be exercised directly.

lxy-9602:
+1

lxy-9602 (approved):
+1

PR #458, opened 2026-07-29
fix(lucene): normalize prefix and wildcard queries

lszskye:
### Purpose
Fix case-sensitive query mismatches in the Lucene full-text index.

Lucene normalizes alphanumeric tokens to lowercase when building the index, but PREFIX and WILDCARD queries previously used the original query text directly. As a result, queries could not match indexed lowercase terms.

This change:

- extracts the existing case-normalization logic for reuse;
- normalizes PREFIX query terms before constructing `PREFIX Query`;
- normalizes each literal segment of `WILDCARD Query` while preserving `*` and `?`;
- adds unit and integration coverage for uppercase PREFIX and WILDCARD queries.

### Tests
- LuceneGlobalIndexTest
- GlobalIndexTest

### Generative AI tooling

Generated-by: Codex (GPT-5)


lxy-9602 (dismissed):
+1

lxy-9602 (approved):
+1

PR #460, opened 2026-07-29
fix(blob): allow blob files across schema IDs

lxy-9602:
### Purpose

Linked issue: close #xxx

Blob files for the same field may span different schema IDs after schema evolution and normal file compaction.

Remove the schema ID consistency check from `BlobBunch` while retaining the write-column consistency check.

### Tests

Added `BlobTableInteTest.TestBlobFilesAcrossSchemaIds` to simulate a post-compaction layout where one normal file bridges blob files written before and after schema evolution.

### Generative AI tooling


lszskye (approved):
+1

PR #462, opened 2026-07-29
fix(core): support fallback keys for ignore-delete option

### LINKED ISSUE #461: [Bug] CoreOptions does not resolve Java fallback keys for ignore-delete (e.g. first-row.ignore-delete)
### Bug description

Paimon C++ only recognizes the `ignore-delete` table option. Java `CoreOptions` declares `IGNORE_DELETE` with fallback keys, so tables created by Java connectors may carry the semantics under a fallback key instead:

```java
public static final ConfigOption<Boolean> IGNORE_DELETE =
        key("ignore-delete")
                .booleanType()
                .defaultValue(false)
                .withFallbackKeys(
                        "first-row.ignore-delete",
                        "deduplicate.ignore-delete",
                        "partial-update.ignore-delete")
                .withDescription("Whether to ignore delete records.");
```

When Java (e.g. release-1.4) creates a `merge-engine = first-row` primary-key table and writes `first-row.ignore-delete = true` into the table options, C++ `CoreOptions` does not read that key and keeps the default `ignore_delete = false`. A C++ writer/compaction on such a table then rejects retract records in `FirstRowMergeFunction`:

```
By default, First row merge engine does not accept DELETE/UPDATE_BEFORE records.
You can config 'first-row.ignore-delete' to ignore the DELETE/UPDATE_BEFORE records.
```

So the ignore-delete semantics configured on the Java side are silently lost across languages.

### Reproduce

1. Use Java Paimon to create a primary-key table with `merge-engine = first-row`; the table schema carries `first-row.ignore-delete = true`.
2. Write data containing a `DELETE` (or `UPDATE_BEFORE`) record with C++ and trigger compaction, e.g. `+I Alice ...` followed by `-D Alice ...`.
3. Expected: the delete record is ignored and the first row `Alice` is kept. Actual: C++ compaction fails because `ignore_delete_` is `false`.

### Root cause

- C++ parses only `Options::IGNORE_DELETE` (`ignore-delete`) in `CoreOptions` (`src/paimon/core/core_options.cpp`), with default `false`.
- Java resolves `IGNORE_DELETE` through fallback keys `first-row.ignore-delete`, `deduplicate.ignore-delete` and `partial-update.ignore-delete`, checked in that order when `ignore-delete` itself is absent.

### Proposed fix

Align C++ `CoreOptions` parsing with Java `CoreOptions` fallback-key semantics:

- Parse `ignore-delete` first; when absent, fall back to `first-row.ignore-delete`, `deduplicate.ignore-delete`, `partial-update.ignore-delete` in Java's declaration order (primary key takes precedence when both are present).
- Add the fallback key constants (`Options::FALLBACK_FIRST_ROW_IGNORE_DELETE`, `Options::FALLBACK_DEDUPLICATE_IGNORE_DELETE`, `Options::FALLBACK_PARTIAL_UPDATE_IGNORE_DELETE`) to `paimon/defs.h`, following the existing `FALLBACK_BLOB_DESCRIPTOR_FIELD` pattern.
- Add unit tests covering fallback resolution, precedence and invalid values in `core_options_test.cpp`.

With this change, first-row tables created by Java with `first-row.ignore-delete = true` are correctly honored by C++ read/compaction, and cross-language compaction with DELETE/UPDATE_BEFORE records keeps the first row instead of failing.

SteNicholas:
### Purpose

Linked issue: close #461

The `ignore-delete` semantics can also be configured through the merge-engine-specific keys
`first-row.ignore-delete`, `deduplicate.ignore-delete` and `partial-update.ignore-delete`, which are
declared as fallback keys of `ignore-delete`. `CoreOptions` only parsed `ignore-delete` and kept the
default `false` otherwise, so a table configured with `first-row.ignore-delete=true` was read as
`ignore-delete=false`. The first-row merge engine then rejected retract records with:

```
By default, First row merge engine can not accept DELETE/UPDATE_BEFORE records.
You can config 'first-row.ignore-delete' to ignore the DELETE/UPDATE_BEFORE records.
```

This change parses the fallback keys in fixed order (`first-row.ignore-delete`,
`deduplicate.ignore-delete`, `partial-update.ignore-delete`) when `ignore-delete` itself is absent,
keeping the primary key's precedence when both are present. The fallback key constants follow the
existing `Options::FALLBACK_BLOB_DESCRIPTOR_FIELD` pattern.

### Tests

- `CoreOptionsTest.TestIgnoreDeleteFallbackKeys` — each fallback key resolves, the primary key takes
  precedence over a fallback key, fallback keys are checked in fixed order, and an invalid value is
  rejected.
- `PrimaryKeyTableUtilsTest.TestCreateFirstRowMergeFunctionWithIgnoreDelete` — a first-row merge
  function built from `first-row.ignore-delete` ignores a DELETE record and keeps the first row,
  identically to one built from `ignore-delete`; without either option retract records are still
  rejected.

### API and Format

`include/paimon/defs.h` gains three new option-key constants
(`Options::FALLBACK_FIRST_ROW_IGNORE_DELETE`, `Options::FALLBACK_DEDUPLICATE_IGNORE_DELETE`,
`Options::FALLBACK_PARTIAL_UPDATE_IGNORE_DELETE`). This is additive; no existing symbol changes and
the storage format and protocol are unaffected.

### Documentation

No new feature, so no documentation change. The new constants carry doc comments in
`include/paimon/defs.h` following the existing fallback-key convention.

### Generative AI tooling


lxy-9602 (approved):
+1

PR #463, opened 2026-07-30
feat(aggregate): add Java-compatible field aggregators

### LINKED ISSUE #457: [Feature] Add practical field aggregators available in Java Paimon
### Search before asking

### Motivation

Complete the Paimon C++ SDK with practical field aggregators that are already supported by Java Paimon.

### Solution

Add support for the following field aggregators:

  - `collect`
  - `merge_map`
  - `nested_update`
  - `rbm64`
  - `hll_sketch`
  - `theta_sketch`

### Anything else?

_No response_

liangjie3138:
### Purpose

  Linked issue: close https://github.com/alibaba/paimon-cpp/issues/457

  Port the six practical field aggregators missing on the C++ side — `collect`,
  `merge_map`, `nested_update`, `rbm64`, `hll_sketch`, `theta_sketch` .

  Also adds `GenericArray` / `GenericMap`, the in-memory `InternalArray` /
  `InternalMap` implementations these aggregators need to build ARRAY and MAP
  results.

  ### Tests

  18 new unit test cases in `paimon-core-test`, covering each aggregator's
  aggregate / retract / null / type-validation paths.

  Three pin behaviour that is easy to "fix" incorrectly, and each was checked by
  reverting the corresponding production change and confirming it fails:

  - `BinaryAggMergeFunctionTest.OwnedAccumulatorSurvivesNullInput` — drives a real
    `AggregateMergeFunction` so the use-after-free is exercised through the
    production path.
  - `FieldNestedUpdateAggTest.CountLimitCountsNullElementsOfAccumulator` — count
    limit is measured against the raw element count, matching Java.
  - `FieldIgnoreRetractAggTest.ReversedAggBypassesWrappedOverride` — the wrapper
    does not preserve the wrapped aggregator's reversed-aggregation override, also
    matching Java.

  No integration tests added.

  ### API and Format

  Additive only, no signatures changed:

  - `include/paimon/defs.h`: four new `Options` constants — `NESTED_KEY`,
    `NESTED_KEY_NULL_STRATEGY`, `NESTED_SEQUENCE_FIELD`, `COUNT_LIMIT`.
  - `CoreOptions`: a `NestedKeyNullStrategy` enum plus accessors.
  - `FieldAggregator`: new virtuals default to the existing `Agg` / `AggReversed`.

  New dependency: Apache DataSketches 5.2.0 (Apache-2.0), resolved through
  `resolve_dependency()` like the other bundled dependencies.

  ### Documentation

  None.

  ### Generative AI tooling

  Generated-by: Claude Code 2.1.220 (Claude Opus 5)


src/paimon/core/mergetree/compact/aggregate/field_roaring_bitmap64_agg.cpp
```diff
@@ -0,0 +1,62 @@
...
+    std::string_view accumulator_bytes = DataDefine::GetStringView(accumulator);
+    std::string_view input_bytes = DataDefine::GetStringView(input_field);
+    RoaringBitmap64 accumulator_bitmap;
+    RoaringBitmap64 input_bitmap;
+    PAIMON_RETURN_NOT_OK(
+        accumulator_bitmap.Deserialize(accumulator_bytes.data(), accumulator_bytes.size()));
+    PAIMON_RETURN_NOT_OK(input_bitmap.Deserialize(input_bytes.data(), input_bytes.size()));
+    accumulator_bitmap |= input_bitmap;
+    return VariantType(std::shared_ptr<Bytes>(accumulator_bitmap.Serialize(/*pool=*/nullptr)));
+}
```
lxy-9602: There is an issue here: the bitmap serialization currently used in Java is not cross-language compatible. To ensure compatibility, it needs to use `Roaring64NavigableMap.serializePortable()` instead. In some scenarios, data written by Java cannot be deserialized by C++ or other languages, and vice versa. Should we temporarily remove bitmap64 aggregation support until Java serialization is adjusted for compatibility, to avoid cases where data written in one language cannot be read in another?

src/paimon/core/mergetree/compact/aggregate/aggregate_merge_function.cpp
```diff
@@ -85,7 +85,8 @@ Status AggregateMergeFunction::Add(KeyValue&& kv) {
             PAIMON_ASSIGN_OR_RAISE(merged_field,
                                    aggregators_[i]->Retract(accumulator, input_field));
         } else {
-            merged_field = aggregators_[i]->Agg(accumulator, input_field);
+            PAIMON_ASSIGN_OR_RAISE(merged_field,
+                                   aggregators_[i]->AggResult(accumulator, input_field));
         }
```
lxy-9602, line 90: We can still use `Agg` in the naming here; the only difference is that it now has a return value.

src/paimon/core/mergetree/compact/aggregate/field_aggregator.h
```diff
@@ -35,6 +35,17 @@ class FieldAggregator {
...
+    /// external representations override this method to report invalid input without throwing.
+    ///
+    /// @param accumulator Current aggregated value.
+    /// @param input_field Value to merge into the accumulator.
+    /// @return The merged value, or an error Status.
+    virtual Result<VariantType> AggResult(const VariantType& accumulator,
+                                          const VariantType& input_field) {
+        return Agg(accumulator, input_field);
+    }
+
```
lxy-9602, line 54: Could we remove the newly added `AggResult / AggReversedResult` functions and simply update the return values of `Agg / AggReversed` instead?

src/paimon/core/mergetree/compact/aggregate/field_aggregate_utils.cpp
```diff
@@ -0,0 +1,237 @@
...
+    return FieldAggregateUtils::Equals(lhs_value, rhs_value, type);
+}
+
+Result<bool> EqualRows(const std::shared_ptr<InternalRow>& lhs,
+                       const std::shared_ptr<InternalRow>& rhs,
+                       const std::shared_ptr<arrow::StructType>& type) {
+    if (!lhs || !rhs || lhs->GetFieldCount() != rhs->GetFieldCount() ||
+        lhs->GetFieldCount() != type->num_fields()) {
+        return lhs == rhs;
+    }
```
lxy-9602, line 56: Should we also add a comparison for `getRowKind()` here?

src/paimon/core/mergetree/compact/aggregate/field_collect_agg.cpp
```diff
@@ -0,0 +1,175 @@
...
+namespace {
+
+Result<bool> Contains(const std::vector<VariantType>& values, const VariantType& candidate,
+                      const std::shared_ptr<arrow::DataType>& element_type) {
+    for (const VariantType& value : values) {
+        PAIMON_ASSIGN_OR_RAISE(bool equal,
+                               FieldAggregateUtils::Equals(value, candidate, element_type));
+        if (equal) {
+            return true;
+        }
```
lxy-9602, line 43: It looks like there may be some performance regression here. Given the size of this PR, maybe we could add a `TODO` or note documenting the current O(n²) complexity and optimize it in a follow-up.

src/paimon/core/core_options.cpp
```diff
@@ -1378,6 +1307,61 @@ Result<bool> CoreOptions::FieldAggIgnoreRetract(const std::string& field_name) c
     return field_agg_ignore_retract;
 }
 
+Result<std::vector<std::string>> CoreOptions::FieldNestedUpdateAggNestedKey(
+    const std::string& field_name) const {
```
lxy-9602, line 1382: Please add tests for the newly added `CoreOptions` APIs to `CoreOptionsTest`.

src/paimon/core/core_options.cpp
```diff
@@ -1378,6 +1307,61 @@ Result<bool> CoreOptions::FieldAggIgnoreRetract(const std::string& field_name) c
...
+}
+
+Result<CoreOptions::NestedKeyNullStrategy> CoreOptions::FieldNestedUpdateAggNestedKeyNullStrategy(
+    const std::string& field_name) const {
+    ConfigParser parser(impl_->raw_options);
+    std::string strategy = "merge";
+    std::string key = std::string(Options::FIELDS_PREFIX) + "." + field_name + "." +
+                      std::string(Options::NESTED_KEY_NULL_STRATEGY);
+    PAIMON_RETURN_NOT_OK(parser.Parse(key, &strategy));
+    if (strategy == "merge") {
```
lxy-9602: I’d like to confirm whether `strategy` is case-sensitive in Java.

src/paimon/core/mergetree/compact/aggregate/field_sketch_agg.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+namespace paimon {
+namespace {
+
+template <typename T>
+std::shared_ptr<Bytes> CopyBytes(const std::vector<T>& serialized) {
+    pooled_unique_ptr<Bytes> result =
+        Bytes::AllocateBytes(serialized.size() * sizeof(T), GetDefaultPool().get());
+    if (!serialized.empty()) {
+        std::memcpy(result->data(), serialized.data(), result->size());
```
lxy-9602, line 41: Just to clarify: I noticed many places are using `GetDefaultPool()`. Will the data volume during merging be relatively large? In paimon-cpp, batches are merged together, and data that has not yet been converted into Arrow arrays may stay held in memory for a while. Is it appropriate for that memory to be allocated from a default pool that is separate from the user-provided memory pool?

src/paimon/core/mergetree/compact/aggregate/field_sketch_agg.cpp
```diff
@@ -0,0 +1,136 @@
...
+    return result.ok() ? std::move(result).value() : accumulator;
+}
+
+Result<VariantType> FieldThetaSketchAgg::AggResult(const VariantType& accumulator,
+                                                   const VariantType& input_field) {
+    bool accumulator_null = DataDefine::IsVariantNull(accumulator);
+    bool input_null = DataDefine::IsVariantNull(input_field);
+    if (accumulator_null || input_null) {
+        return accumulator_null ? input_field : FieldAggregateUtils::OwnedBinary(accumulator);
+    }
```
lxy-9602, line 115: ### Annotation 1

**[P1] Preserve binary ownership in the reversed null path**

When `PartialUpdateMergeFunction` processes an older sequence-group record, it calls `AggReversedResult(accumulator, field)`. The base implementation swaps the arguments and invokes `AggResult(field, accumulator)`.

For `hll_sketch`, and `theta_sketch`, if the older field is null, the null branch returns `input_field` directly. This value is a non-owning `std::string_view` produced by the `use_view=true` getter. If the current accumulator was created by a previous aggregation, its buffer is owned solely by the `shared_ptr<Bytes>` stored in `row_->fields_[i]`.

The subsequent `row_->SetField(i, result)` replaces that `shared_ptr<Bytes>` with the returned view, releases the underlying buffer, and leaves a dangling view in the row. Later access may cause corrupted data, deserialization failures, or a use-after-free.

A valid trigger sequence is:

```text
seq=10, sketch=A
seq=20, sketch=B       // union creates owned Bytes
seq=15, sketch=NULL    // enters the reversed null path
```

Please ensure that either null-argument order returns an owning copy of the non-null binary value, while preserving null when both arguments are null. Also add a partial-update regression test covering a sequence group, a previously aggregated binary value, and an older record with a null field. The existing `OwnedAccumulatorSurvivesNullInput` test only covers the forward `AggResult()` path.

src/paimon/core/mergetree/compact/aggregate/field_nested_update_agg.cpp
```diff
@@ -0,0 +1,370 @@
...
+    std::vector<int32_t> fields;
+    fields.reserve(names.size());
+    for (const std::string& name : names) {
+        int32_t index = row_type->GetFieldIndex(name);
+        if (index < 0) {
+            return Status::Invalid(fmt::format("Field '{}' configured by '{}' does not exist in {}",
+                                               name, option_name, row_type->ToString()));
+        }
+        if (std::find(fields.begin(), fields.end(), index) != fields.end()) {
+            return Status::Invalid(
```
lxy-9602: I recall that `GetFieldIndex()` also returns -1 when it encounters duplicate fields, so this find check seems unnecessary.

src/paimon/core/mergetree/compact/aggregate/field_nested_update_agg.cpp
```diff
@@ -0,0 +1,370 @@
...
+            FieldAggregateUtils::Equals(lhs_value, rhs_value, row_type_->field(field)->type()));
+        if (!equal) {
+            return false;
+        }
+    }
+    return true;
+}
+
+Result<bool> FieldNestedUpdateAgg::RowsEqual(const InternalRow& lhs, const InternalRow& rhs) const {
+    for (int32_t field = 0; field < row_type_->num_fields(); ++field) {
```
lxy-9602: Why not use `FieldAggregateUtils::Equals` directly here?

src/paimon/core/mergetree/compact/aggregate/field_nested_update_agg_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+    ASSERT_OK_AND_ASSIGN(CoreOptions invalid_strategy,
+                         CoreOptions::FromMap({{"fields.f.nested-key", "id"},
+                                               {"fields.f.nested-key-null-strategy", "invalid"}}));
+    ASSERT_NOK(FieldNestedUpdateAgg::Create(NestedType(), invalid_strategy, "f"));
+
+    ASSERT_OK_AND_ASSIGN(CoreOptions negative_limit,
+                         CoreOptions::FromMap({{"fields.f.count-limit", "-1"}}));
+    ASSERT_NOK(FieldNestedUpdateAgg::Create(NestedType(), negative_limit, "f"));
+}
+
```
lxy-9602, line 486: [P1] Please align the new aggregator tests with the Java test matrix

The current C++ tests mainly cover basic happy paths and are substantially narrower than the corresponding Java tests.

The Java behavior is primarily covered by:

- paimon-core/src/test/java/org/apache/paimon/mergetree/compact/aggregate/FieldAggregatorTest.java
- paimon-core/src/test/java/org/apache/paimon/mergetree/compact/aggregate/FieldAggregatorRetractNullTest.java

Important missing cases include:

- collect: distinct and non-distinct aggregation/retraction for primitive, ROW, ARRAY, and MAP elements.
- nested_update: composite nested keys, multiple sequence fields, count-limit boundary cases, all null-key strategies on both accumulator and retract input, and retraction without a nested key.
- rbm64, hll_sketch, and theta_sketch: the complete null-input matrix and the unsupported-retraction contract.

These gaps allow Java compatibility issues to go unnoticed. Please port the relevant Java cases or add equivalent parameterized tests.

Please also add end-to-end write/compaction/read tests for the new aggregators to verify the final data read from files, rather than testing only the in-memory aggregation result.

src/paimon/core/mergetree/compact/aggregate/aggregate_merge_function.cpp
liangjie3138: Agreed. Renamed back to Agg / AggReversed, with only the return type changed to Result<VariantType>.

src/paimon/core/mergetree/compact/aggregate/field_aggregator.h
liangjie3138: Removed both.

src/paimon/core/mergetree/compact/aggregate/field_aggregate_utils.cpp
liangjie3138: Good catch. EqualRows now compares the row kind before walking the fields, and I added a retract testthat relies on it.

src/paimon/core/mergetree/compact/aggregate/field_collect_agg.cpp
liangjie3138: Added a TODO here that names all three quadratic lookups (this scan, the merge_map key lookup, the nested_update keyed upsert) and the hash-by-type fix they share.

src/paimon/core/core_options.cpp
liangjie3138: Extended CoreOptionsTest to cover the new accessors — defaults, parsing from a map, and the invalid-value cases.

src/paimon/core/mergetree/compact/aggregate/field_sketch_agg.cpp
liangjie3138: You're right, and my first pass only fixed half of it — I missed that AggReversed swaps the arguments, so the same view comes back through input_field.

  Both sketch aggregators now hand the surviving value to a shared OwnedBinary helper that copies it into a new buffer, whichever side it came from, and return null only when both sides are null. Added the
  partial-update regression test for your seq 10/20/15 case, plus the four argument orderings at the aggregator level.

src/paimon/core/mergetree/compact/aggregate/field_nested_update_agg.cpp
liangjie3138: Dropped this check.

src/paimon/core/mergetree/compact/aggregate/field_nested_update_agg.cpp
liangjie3138: Replaced.

src/paimon/core/mergetree/compact/aggregate/field_nested_update_agg_test.cpp
liangjie3138: Ported the missing cases from the Java suite into the collect, nested_update and sketch unit tests.

src/paimon/core/mergetree/compact/aggregate/field_roaring_bitmap64_agg.cpp
liangjie3138: Thanks for catching this. I've removed the bitmap64 aggregator for now. One thing left open: existing tables already using rbm64 aggregation can't be read by C++ until we have a compatible format

lxy-9602 (approved):
+1

lxy-9602:
This has already been merged in https://github.com/apache/paimon-cpp/pull/183, so we can close this PR.

PR #464, opened 2026-07-30
Fix memory pool lifetime for Arrow stream buffers

mrdrivingduck:
I found this while debugging an S3 crash. I initially suspected the S3 filesystem, but the issue was in the Arrow input-stream adapter: a returned buffer can outlive its memory pool when a reader is closed while a caller still holds a buffer. This is more likely with network filesystems, where asynchronous reads may finish after the reader is closed.

This patch uses `shared_ptr`'s aliasing constructor to keep the pool alive with the returned buffer. It also adds deterministic coverage for `Read`, `ReadAt`, and `ReadAsync` using a small `InputStream` mock. Without the change, the tests crash; with it, they pass.

### DIFF
--- a/src/paimon/common/utils/arrow/arrow_input_stream_adapter.cpp
+++ b/src/paimon/common/utils/arrow/arrow_input_stream_adapter.cpp
@@ -40,6 +40,19 @@ arrow::Status ValidateArrowIoRange(int64_t value, const char* name) {
     return arrow::Status::OK();
 }
 
+struct BufferWithMemoryPool {
+    std::shared_ptr<arrow::MemoryPool> pool;
+    std::shared_ptr<arrow::Buffer> buffer;
+};
+
+std::shared_ptr<arrow::Buffer> KeepMemoryPoolAlive(std::shared_ptr<arrow::Buffer> buffer,
+                                                   const std::shared_ptr<arrow::MemoryPool>& pool) {
+    auto holder =
+        std::make_shared<BufferWithMemoryPool>(BufferWithMemoryPool{pool, std::move(buffer)});
+    auto* buffer_ptr = holder->buffer.get();
+    return std::shared_ptr<arrow::Buffer>(std::move(holder), buffer_ptr);
+}
+
 }  // namespace
 
 ArrowInputStreamAdapter::ArrowInputStreamAdapter(
@@ -79,7 +92,7 @@ arrow::Result<std::shared_ptr<arrow::Buffer>> ArrowInputStreamAdapter::Read(int6
     if (read_bytes < nbytes) {
         ARROW_RETURN_NOT_OK(buffer->Resize(read_bytes));
     }
-    return std::shared_ptr<arrow::Buffer>(std::move(buffer));
+    return KeepMemoryPoolAlive(std::shared_ptr<arrow::Buffer>(std::move(buffer)), pool_);
 }
 
 arrow::Result<int64_t> ArrowInputStreamAdapter::ReadAt(int64_t position, int64_t nbytes,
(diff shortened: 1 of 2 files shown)


src/paimon/common/utils/arrow/arrow_stream_adapter_test.cpp
```diff
@@ -88,4 +164,54 @@ TEST(ArrowStreamAdapterTest, TestInputAndOutputStream) {
...
 }
 
+TEST(ArrowStreamAdapterTest, TestReadKeepsMemoryPoolAliveUntilBufferReleased) {
+    auto stream = std::make_shared<DeferredInputStream>();
+    std::shared_ptr<arrow::MemoryPool> pool(GetArrowPool(GetDefaultPool()));
+    auto adapter = CreateAdapter(stream, pool);
+    ASSERT_NE(adapter, nullptr);
+
+    auto result = adapter->Read(kTestSize);
+    ASSERT_TRUE(result.ok());
```
zjw1111: could you use `ASSERT_OK_AND_ASSIGN` here

mrdrivingduck: No, seems to be an `arrow::Result` instead of `paimon::Result` here, so `ASSERT_OK_AND_ASSIGN` is not supported.

zjw1111: Oh, I misspoke earlier. You can just use `std::shared_ptr<arrow::Buffer> buffer = adapter->Read(kTestSize).ValueOrDie();` directly.

mrdrivingduck: done

lucasfang (approved):
+1

src/paimon/common/utils/arrow/arrow_input_stream_adapter.cpp
```diff
@@ -40,6 +40,19 @@ arrow::Status ValidateArrowIoRange(int64_t value, const char* name) {
     return arrow::Status::OK();
 }
 
+struct BufferWithMemoryPool {
```
zjw1111, line 43: We have a performance concern, as this would increment the shared_ptr reference count on every I/O operation. Have you evaluated the performance overhead of introducing shared_ptr here?

mrdrivingduck: I ran a **pure in-memory** microbenchmark. Results (`before -> after`, ns/op):

| Read size | Read | ReadAt | ReadAsync |
| --- | ---: | ---: | ---: |
| 4 KiB, 1 thread | +10.4% | +10.9% | +7.9% |
| 64 KiB, 1 thread | +3.7% | +8.4% | +4.6% |
| 1 MiB, 1 thread | +0.0% | -1.2% | +0.1% |

The overhead becomes smaller as the read size increases. The fix is still needed to keep the memory pool alive while returned buffers are referenced; otherwise, the ownership between the pool, reader, and buffers would need to change.

zjw1111: OK, get.

zjw1111:
Could you please move this PR to https://github.com/apache/paimon-cpp? All future development will take place under the Apache repo.

mrdrivingduck:
> Could you please move this PR to https://github.com/apache/paimon-cpp? All future development will take place under the Apache repo.

Yes, but should we move S3 related stuff first?

mrdrivingduck:
> Could you please move this PR to https://github.com/apache/paimon-cpp? All future development will take place under the Apache repo.

https://github.com/apache/paimon-cpp/pull/182

mrdrivingduck:
> Yes, but should we move S3 related stuff first?

Oh, seems that it has already been moved.

PR #468, opened 2026-07-30
feat(parquet): support configuring Arrow pre-buffer hole-size-limit for range coalescing

zhf999:
### Purpose

The Parquet reader already exposes `parquet.read.cache-option.lazy`, `parquet.read.cache-option.prefetch-limit`, and `parquet.read.cache-option.range-size-limit` to configure Arrow's pre-buffer `CacheOptions`, but the `hole_size_limit` was always left at Arrow's default (8 KiB). This byte-level gap threshold decides whether adjacent I/O ranges are coalesced into a single read request, which directly determines the I/O pattern on paths where Paimon's own `ReadAheadCache` is not in effect (e.g. non-prefetch reads or cache modes that skip caching).

This PR:

- Adds a new option `parquet.read.cache-option.hole-size-limit` (bytes) to control `arrow::io::CacheOptions::hole_size_limit`, defaulting to 8 KB (`DEFAULT_PARQUET_READ_CACHE_OPTION_HOLE_SIZE_LIMIT`) so that small gaps between column-chunk ranges are merged into fewer, larger reads, which benefits high-latency storage.
- Validates the configuration in `ParquetFileBatchReader::CreateArrowReaderProperties`:
  - `hole-size-limit` must be non-negative;
  - `range-size-limit` must be strictly greater than `hole-size-limit`.
- Documents the byte-level semantics of both coalescing limits in `parquet_format_defs.h`.

### Tests

- UT: `ParquetFileBatchReaderTest.TestCreateArrowReaderProperties` (extended)
  - Default cache options now carry `hole_size_limit = DEFAULT_PARQUET_READ_CACHE_OPTION_HOLE_SIZE_LIMIT`.
  - All four `parquet.read.cache-option.*` options set together are correctly propagated to `arrow::io::CacheOptions`.
  - Negative `hole-size-limit` is rejected with `Status::Invalid`.
  - `range-size-limit <= hole-size-limit` is rejected with `Status::Invalid`.

### API and Format

No public API change under `include/paimon/` and no storage format/protocol change. Adds one new read option key `parquet.read.cache-option.hole-size-limit` in `src/paimon/format/parquet/parquet_format_defs.h`; the effective default Arrow pre-buffer coalescing behavior changes as described above.

### Documentation

No new user-facing feature docs required; the new option and the semantics of the coalescing limits are documented via comments in `parquet_format_defs.h`.

### Generative AI tooling

No

### DIFF
--- a/src/paimon/format/parquet/parquet_file_batch_reader.cpp
+++ b/src/paimon/format/parquet/parquet_file_batch_reader.cpp
@@ -617,12 +617,28 @@ Result<::parquet::ArrowReaderProperties> ParquetFileBatchReader::CreateArrowRead
         OptionsUtils::GetValueFromMap<int64_t>(options, PARQUET_READ_CACHE_OPTION_PREFETCH_LIMIT,
                                                DEFAULT_PARQUET_READ_CACHE_OPTION_PREFETCH_LIMIT));
     PAIMON_ASSIGN_OR_RAISE(
+        int64_t cache_hole_size_limit,
+        OptionsUtils::GetValueFromMap<int64_t>(options, PARQUET_READ_CACHE_OPTION_HOLE_SIZE_LIMIT,
+                                               DEFAULT_PARQUET_READ_CACHE_OPTION_HOLE_SIZE_LIMIT));
+    PAIMON_ASSIGN_OR_RAISE(
         int64_t cache_range_size_limit,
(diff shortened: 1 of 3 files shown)


duanyyyyyyy:
LGTM

lxy-9602 (approved):
+1

